PR #3, opened 2018-07-13
Added a button to the relationship details popup to be able to flip t…

magewe1:
…he direction of a relationship.


mtnc:
There is already an action `FLIP_RELATIONSHIPS` in `src/gui/redux/relationships/actions.ts`. It is used in `src/gui/events/KeyboardEventListener.ts` to flip selected relationships. Please re-use the existing function for the flip icon.

PR #5, opened 2019-01-29
Prerelease of Apollon 2.0

jwilland-tum:
This pull request introduces Apollon 2.0, a completely refactored solution for modelling UML diagrams.

The current prerelease comes with
 - a new underlying domain model to support new features like nesting elements and a plugin system to easily extend the editor by new element types and
 - an optimised implementation of react components to improve performance and prevent endless props forwarding.

For the final release of version 2.0, some features still require refactoring, like the adjustment of relationships to the new domain model. So far, the interface did not change, but be aware that the new release will contain breaking changes.


dev/index.html
```diff
@@ -1,43 +0,0 @@
-<!doctype html>
-<html>
```
krusche: Why did you delete this file? It was quite good for testing

PR #10, opened 2019-03-10
Reconnectable Relationships

jwilland-tum:
### Motivation and Context

For a better experience, users should be able to reconnect existing relationships with drag and drop interaction.

### Description

- Existing relationships now have an interactive area on both ends to start dragging the relationship
- After dragging started, behaviour is the same as with creating new relationships
- If the users drops the relationship on a connectable port, the relationship is reconnected to the new port
- If the user drops the relationship somewhere else, or on the same Port it is already connected to, the previous state of the relationship is restored
- If multiple relationships lay over each other, only the newest relationship can be reconnected, but if one of them is selected, the selected one will be the one to reconnect

### Screenshots

reconnectable_relationships


jwilland-tum:
Draggable area: There seems to be a bug in safari that pointer-events are not used correctly if no stroke is set at all or set to `none`. The draggable area is now as wide as the blue highlight area when hovering or selecting a relationship in every browser. Please let me know, if I should even increase that size, or if should use a different shape like half a circle (like the connectors) with an additional path. Also, should we stick with the `moving`-cursor or should we use the same cursor we use when creating relationships (`crosshair`-cursor)?

Wrong hover state for connectors: This issue was reproducible so far when dropping relationships on the invisible part of the connector (the part on top of the element itself). Fixed this issue by forcing a leave event also when the user drops a relationship.

Unfortunately, I was not able to reproduce an example in which dragging relationships does not work at all. Might be the case that the draggable area was too small. Please let me know if you can reproduce this even after the most recent bugfixes.

PR #16, opened 2019-03-21
Improve Interface and Typings

jwilland-tum:
### Motivation and Context

In the most recent released version of Apollon, the interface to integrate and interact with Apollon was primarily adjusted to fit in https://github.com/ls1intum/ArTEMiS and suit the short term needs. With the new Apollon Version 2.0, we want to refactor the exposed functionality and typing definitions to allow a more generic way to communicate with the application. Especially when it comes to the newly introduced diagram types.

### Description

Functionality is defined in the file [ApollonEditor](https://github.com/ls1intum/Apollon/blob/feature/interface/src/ApollonEditor.ts). All functionality and typing definitions required to perform all kind of interaction between Apollon and the integrating system are defined there.


src/ApollonEditor.ts
```diff
@@ -0,0 +1,137 @@
...
+export interface UMLElement {
+  id: string;
+  name: string;
+  owner: string | null;
+  type: ElementKind;
+  bounds: { x: number; y: number; width: number; height: number };
+}
+
+export const enum Location {
+  North,
```
krusche: while this is a really nice relation to Compass in Artemis ;-) I would suggest that we use Up, Left, Down, Right here instead. Also we might want to call it Direction instead of Location. What do you think?

jwilland-tum: Sounds good! Applied rename to:
```
export const enum Direction {
  Up,
  Right,
  Down,
  Left,
}
```

PR #24, opened 2019-04-07
Communication Diagrams

jwilland-tum:
### Motivation and Context

Introduce Communication Diagrams to Apollon

### Description

Users should be able to model UML Communication Diagrams in Apollon. Therefore we implemented Communication Link that contain messages along their path. Messages have a certain direction and a name value.

### Screenshots

comm

### DIFF
--- a/src/components/sidebar/sidebar-component.tsx
+++ b/src/components/sidebar/sidebar-component.tsx
@@ -284,6 +284,16 @@ class SidebarComponent extends Component<Props, State> {
             })(),
           ],
         });
+      case DiagramType.CommunicationDiagram:
+        this.setState({
+          previews: [
+            (() => {
+              const c = new ObjectName();
+              c.name = this.props.translate('packages.objectDiagram.objectName');
+              return c;
+            })(),
+          ],
+        });
     }
   };
 }
--- a/src/components/sidebar/sidebar-component.tsx
+++ b/src/components/sidebar/sidebar-component.tsx
@@ -15,7 +15,7 @@ import { Enumeration } from '../../packages/class-diagram/classifier/enumeration
 import { Interface } from '../../packages/class-diagram/classifier/interface/interface';
 import { Package } from '../../packages/common/package/package';
 import { ElementType } from '../../packages/element-type';
-import { ObjectAttribute } from '../../packages/object-diagram/object-attribute/object-attribute';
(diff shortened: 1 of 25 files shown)


src/components/sidebar/sidebar-component.tsx
```diff
@@ -284,6 +284,16 @@ class SidebarComponent extends Component<Props, State> {
             })(),
           ],
         });
+      case DiagramType.CommunicationDiagram:
```
jwilland-tum: Should we allow different elements than `ObjectName`s? Some examples contained `Actor`s as well.

jwilland-tum:
Should `ObjectName`s have methods?

src/packages/communication-diagram/communication-link/communication-link-popup.tsx
```diff
@@ -0,0 +1,130 @@
...
+            <Header>{this.props.translate('packages.communicationDiagram.communicationLink')}</Header>
+          </Flex>
+          <Divider />
+        </Section>
+        <Section>
+          <Header>{this.props.translate('popup.messages')}</Header>
+          {element.messages.map((message, i) => (
+            <Flex key={i}>
+              <TextField value={message.name} onUpdate={this.rename(message)} />
+              <Flip onClick={this.flip(message)} />
```
jwilland-tum: Should the flip icon in the popup show the current direction of a message?

jwilland-tum: Should the flip icon in the popup show the current direction of a message?

src/components/sidebar/sidebar-component.tsx
krusche: we don't need actors for now. Objects are enough. Can we display attributes and methods in the objects? (this should work similar to Object Diagrams)

src/packages/communication-diagram/communication-link/communication-link-popup.tsx
krusche: I think the current direction of a message would only make sense if the connected objects are displayed somehow (e.g. below **Messages** with a connection, e.g. ObjectA --- ObjectB). Otherwise left / right would not always makes sense (e.g. when the arrows show up or down in some cases)

krusche:
> Should `ObjectName`s have methods?

I guess you mean `Object`. Yes, it should be possible to add methods into them

PR #76, opened 2019-07-21
[Bugfix] Investigate high compilation time

jwilland-tum:
Fixed some issues which led to a typescript compilation time of approx. one hour.

### DIFF
--- a/src/components/store/model-store.tsx
+++ b/src/components/store/model-store.tsx
@@ -1,7 +1,16 @@
-import React, { PropsWithChildren } from 'react';
+import React, { Component, PropsWithChildren } from 'react';
 import { Provider } from 'react-redux';
-import { applyMiddleware, combineReducers, compose, createStore, DeepPartial, Store } from 'redux';
-import createSagaMiddleware from 'redux-saga';
+import {
+  applyMiddleware,
+  combineReducers,
+  compose,
+  createStore,
+  DeepPartial,
+  Reducer,
+  Store,
+  StoreEnhancer,
+} from 'redux';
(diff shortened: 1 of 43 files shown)


src/components/store/model-store.tsx
```diff
@@ -22,36 +32,41 @@ type Props = OwnProps & CanvasContext;
...
-  const store = createStore(reducer, initialState, enhancer);
+  const store: Store<ModelState, Actions> = createStore(reducer, initialState, enhancer);
 
   if (layer) {
     sagaMiddleware.run(saga);
     store.dispatch(LayouterRepository.layout());
   }
 
-  return store;
+  return { store };
```
jwilland-tum: Should not override the internal component state. Always use an object!

src/services/actions.ts
```diff
@@ -17,21 +18,23 @@ import { ReconnectableActions } from './uml-relationship/reconnectable/reconnect
...
-  MovableActions &
-  MovingActions &
-  ResizableActions &
-  ResizingActions &
-  SelectableActions &
-  UpdatableActions &
-  AssessmentActions &
-  UndoActions &
-  CopyActions;
+export type Actions =
```
jwilland-tum: Instead of composing a type consisting of all actions merged together, we need and either-or type representing any valid action type.

jwilland-tum:
Changed from `enum` to `const` when declaring `ElementType` to enforce strongly typed lookup-dictionaries.

jwilland-tum:
Changed each `Reducer` to accept all valid action type.

jwilland-tum:
Compilation time is now less than 10 seconds again.

PR #86, opened 2020-02-18
Feature/create application header

TobiasPr:
- renamed example folder to public (standalone version folder)
- inserted application header similar to Artemis, therefore 3 new dependencies were added: 
  - url-loader (for image)  
  - file-loader (fallback for url-loader)
  - ejs-loader (index.html became ejs. file to automatically include verion of package.json in application header) 
- fix of lint errors of previous pull request (connectable.tsx)
- change of drop event location (now relative to viewport instead of relative to page)

apollon_application_header


krusche:
@TobiasPr If you create a PR, please also add some kind of description 
If the UI changes, please also add a screenshot

TobiasPr:
@krusche I updated the PR and also included templates for pull requests and issues on github similar to Artemis, so that it won't happen again in the future

Issue #88, opened 2020-02-22
[UI] Minor UI Issues


jannismain:
I hope you don't mind me reporting tiny nitpicks like this. :-)

When trying out Apollon, I notices the following minor UI issues:

<img width="1364" alt="Screenshot 2020-02-22 at 1 32 15 PM" src="https://user-images.githubusercontent.com/14290527/75092715-d1a76100-557a-11ea-8525-04b1dd66fcbb.png">

1. The content of the "Diagram Type" Dropdown goes past its boundary. 
I would recommend extending the sidebar or using `...` to indicate overflowing content.
2. Image for "Schnittstelle" overlays "Komponente"
3. "Readonly" checkbox seems very close to the Mode Dropdown Menu.
Also, would write [read-only](https://www.merriam-webster.com/dictionary/read-only)

Best Regards,
Jannis

TobiasPr:
@jannismain thank you for reporting the issue and sorry for the late reply. I will have a look into it :)

PR #89, opened 2020-03-19
Bugfix/88 minor ui issues

TobiasPr:
### Checklist

### Motivation and Context
Fix for #88 + We want more users to report their issues with Apollon so that we can improve

### Description
Changes as requested in issue
Does not work for Firefox, just for chrome so far. We will change that when we include a proper framework for the standalone version (right now vanilla javascript + html + css)

### Steps for Testing

1. select diagram type which is too long for dropdown and check if dropdown adapts
2. select diagram type "Component Diagram" and check if interface preview is correct

Component-Diagram Interface Preview
### Screenshots
Dropdown + read-only field
minor_bugs_interface

Dropdown + read-only field
ui_minor_bugs_dropdown_read_only


alextsakpinis:
Seems good, great job!

PR #97, opened 2020-05-04
Bugfix/assessment dialog wrong order of subelements

TobiasPr:
### Checklist

### Motivation and Context

Solves https://github.com/ls1intum/Artemis/issues/1342

### Description
Container now reorders children, when children are appended to container. Each container implements a reoroderChidlren to do that 
I also removed some dead code appendElements and remove Elements from UMLContainer

### Steps for Testing

1. create class or object diagram
2. add methods and attributes in arbitrary order
3. make sure to add a method before another attribute
4. go to assessment view and check assessment popup of element
5. expected result: assessment items are in the correct order (attributes before methods and order within as depicted in diagram)

### Screenshots

image


package.json
```diff
@@ -42,7 +42,7 @@
   "devDependencies": {
     "@types/react": "16.9.4",
     "@types/react-dom": "16.9.4",
-    "@types/react-redux": "7.1.7",
+    "@types/react-redux": "7.1.3",
```
filip-gregurevic, line 45: Is this the dependency you had to downgrade, as you mentioned in the meet?

TobiasPr: yes it is

madwau:
I also tested it locally and can confirm that the order is correct.

krusche:
> I also tested it locally and can confirm that the order is correct.

thanks!

PR #150, opened 2021-01-20
Feature/storybook

iozbeyli:
### Motivation and Context
- It enables us to develop elements without firing up the whole system

### Description

- Adds storybook and couple of stories for components. 
- Removes processor for styled components and instead use stylelint for css-in-js
- Fixes font family warnings of stylelint

### Steps for Testing
yarn 
yarn storybook

### Screenshots
Screen Shot 2021-01-20 at 13 35 03


src/main/packages/syntax-tree/syntax-tree-nonterminal/syntax-tree-nonterminal-component.tsx
```diff
@@ -17,6 +17,6 @@ export const SyntaxTreeNonterminalComponent: SFC<Props> = ({ element }) => (
   </g>
 );
 
-interface Props {
+export interface Props {
```
TobiasPr, line 20: why are you exporting all of these Props? I thought they are only required locally for the component

TobiasPr (commented):
Looks good so far, however I have one question regarding dependencies: Why do we have babel/core and babel-loader as new dependencies - is this required by storybook?

iozbeyli:
> Looks good so far, however I have one question regarding dependencies: Why do we have babel/core and babel-loader as new dependencies - is this required by storybook?

Yes, they are for storybook.

PR #153, opened 2021-02-22
Reduce redux operations for many elements

iozbeyli:
### Checklist

### Motivation and Context
Updating redux state with the movement of elements can cause problems when there are many elements on the page. With this change we are updating redux state at the end of the movement when there are many elements. Number of elements is determined according to cpu power of the user.

### Description
1. Added cpu performance measure at the beginning
2. Moved element position change to state when there are many elements
3. Removed unnecessary state duplication from reducers
4. Used animationframe instead of debounce for updating movement on mouse move

### Steps for Testing
1. Place elements on the screen and move them
2. Performance tab of Google dev tools can be used, it also allows you to add cpu throttle

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "*.{ts,tsx,js,css,md}": "prettier --write"
   },
   "dependencies": {
+    "cpu-benchmark": "1.0.1",
     "is-mobile": "2.2.2",
     "pepjs": "0.5.3",
     "react": "17.0.1",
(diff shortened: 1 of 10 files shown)


SimonEntholzer (approved):
Code changes seem to make sense, I'm not yet too familiar with apollon.

While testing locally (also with the google dev tools) I think the PR delivers a performance improvement, but I can't really say how big. Testing with a lot of components on develop and this branch yielded quite similar results, without and with cpu throttling.

PR #159, opened 2021-04-13
Enhancement/discrete changes

iozbeyli:
### Motivation
For real time collaboration between users, sending every change to the server is not efficient. Therefore, the update needs to be send only at the end of user actions. For example, when user is moving an element, every bit of movement of the element notifies the subscribers for a model change. Instead we want to notify subscriber only when user lets go the element and is done with updating the model.

### Description
Add a new state in to store that is called lastAction which stores the last action that is dispatched. If the lastAction dispatched ends with END or DELETE, Apollon notifies subscribers about a discrete model change. This way we are only getting the end of each action. Every user action typically calls 3 actions sequentially which are action_START, action and action_END. And there are two exceptions which are APPEND and DELETE.

### Test Coverage
The tests for the new functions are added. 
apollon-editor.ts has 88.56% coverage


MDK5 (approved):
Tested in testing session, works well with the explained limitations

tobiasjungmann (approved):
Tested in testing session. Worked as described

src/main/apollon-editor.ts
```diff
@@ -224,6 +225,23 @@ export class ApollonEditor {
...
+   * Register callback which is executed when the model changes
+   * @param callback function which is called when the model changes
+   * @return returns the subscription identifier which can be used to unsubscribe
+   */
+  subscribeToModelDiscreteChange(callback: (model: UMLModel) => void): number {
+    return this.discreteModelSubscribers.push(callback) - 1;
+  }
+
+  /**
+   * Remove assessment subscription, so that the corresponding callback is no longer executed when the assessment of elements are changed.
```
krusche, line 238: The comment is wrong

src/main/apollon-editor.ts
```diff
@@ -224,6 +225,23 @@ export class ApollonEditor {
     this.modelSubscribers.splice(subscriptionId);
   }
 
+  /**
+   * Register callback which is executed when the model changes
+   * @param callback function which is called when the model changes
+   * @return returns the subscription identifier which can be used to unsubscribe
+   */
+  subscribeToModelDiscreteChange(callback: (model: UMLModel) => void): number {
```
krusche, line 233: It's unclear what discrete changes are. Could you explain this in the JavaDoc?

src/main/apollon-editor.ts
```diff
@@ -294,6 +312,25 @@ export class ApollonEditor {
...
+    this.notifyDiscreteModelSubscribers();
+  };
+
+  private notifyDiscreteModelSubscribers = () => {
+    try {
+      // if state not available -> do not emit changes
+      if (!this.store) return;
+      const model = this.model;
+      if (
+        this.store.getState().lastAction.lastAction.endsWith('END') ||
```
krusche, line 324: Looks a little bit magic. Can you add some comments what kind of actions are filtered here?
Why do you need `lastAction` twice?

krusche (changes requested):
The PR is not well documented. It's unclear what discrete changes are. Please provide more details in the PR description and in the Code documentation.
Have you created automatic tests for the new functionality?

PR #169, opened 2021-05-16
Enhancement/color

iozbeyli:
Try on http://134.209.254.13:8888/

### Checklist

### Motivation and Context
- Adding colors improves the information that is transpired through uml diagrams
- Colors can be used to show related elements between two diagrams (useful for plagiarism in Artemis for example)

### Description
- Adds fill, stroke and text color to elements. 
- Users can choose it from a predetermined list of colors to keep the overall structure simple

### Steps for Testing
1. Create each diagram and each elements and connect them to each other
2. Color each element

### Screenshots
https://user-images.githubusercontent.com/16825065/118381922-43405480-b5f0-11eb-8eb2-e40a65a34dd2.mp4


SimonEntholzer (approved):
Works locally :+1:

ivohashamov (approved):
I like the changes, maybe it makes sense to fix the width of the diagram work field to a certain percentage, so it doesn't move when the options on the right change

JohannesStoehr (approved):
Worked during the testing session

MDK5 (approved):
Tested on ts1 in testing session. works fine, only tokens in petri nets cannot be colored

MDK5 (approved):
Tested during testing session, everything worked well

nicolasruscher (approved):
tested in testing session 👍

coolchock (approved):
tested in testing session, works as expected

lschlesinger (approved):
Coloring works fine for different types of diagrams.

tobiasjungmann (approved):
Works and looks good. Tested in the testing session on Ts2

krusche (changes requested):
I have some change requests / questions: 
1) can you display the color options below the model elements (in case something is selected) so that the elements are always visible?
2) The terms stroke sounds wrong. Can you use `line color`? And for the others please use `fill color` and `text color`.
3) Did you add German translations as well? 
4) And how would this work in Artemis? (would this be activated by default? what happens to the colors used in the assessments?) I don't think this functionality is needed in Artemis, so developers who use the library should be able to disable coloring options.
5) Did you write tests? If yes, please add the test coverage to the PR description

iozbeyli:
> I have some change requests / questions:
> 
>     1. can you display the color options below the model elements (in case something is selected) so that the elements are always visible?
> 
>     2. The terms stroke sounds wrong. Can you use `line color`? And for the others please use `fill color` and `text color`.
> 
>     3. Did you add German translations as well?
> 
>     4. And how would this work in Artemis? (would this be activated by default? what happens to the colors used in the assessments?) I don't think this functionality is needed in Artemis, so developers who use the library should be able to disable coloring options.
> 
>     5. Did you write tests? If yes, please add the test coverage to the PR description

1. Yes, I can do that but it would make it hard to use. Listing elements on the side when something is selected has no use that I can think of. Listing useful information about the selected elements is more beneficial but of course I can do it if you want.
2. and 3. I can change the names and add translations I missed that part.
4. Any of the colors that was used before overwrites the user set colors. I can add the disable option or even add an enable option if you prefer that one.
5. I will add the tests.

Additionally, should I add a legend for colors, I see that it is commonly used in presentations?

krusche:
1. I would still prefer this, because users might be confused why the model elements suddenly disappear. It's very common that a model element is selected and users might not find the model elements to drag in any more then. Also, you can still drag and drop elements even if something is selected. The selection will just disappear then.
3. Add an enable option. I think colors are problematic in this context.

PatZintel (commented):
image
the X schould change color depending on the chosen color, also colors dont get updated correctly directly after choosing a new one
image

lschlesinger (commented):
I find that coloring feature very valuable!!
I think the possibility to build legends/colormaps with the provided color building block should be highlighted or explained somewhere, the use-case for this building block is not clear otherwise IMO
Bildschirmfoto 2021-06-14 um 17 50 41

MDK5 (changes requested):
Tested in testing session, only one issue found:
grafik
The colored x is not very intuitive as a close button, currently is suggests its either the same as reset or only shows the selected color

MDK5 (approved):
Tested during testing session, everything could be colored :+1:

ndricimrr (approved):
Looks good. Colors working.

FuchsDominik (approved):
Couldn't find any issues with the colors :) 👍

JohannesStoehr (changes requested):
The color picker for attributes does not update the chosen color, but changes the diagram correctly.
After also changing the color for the class itself, the picker of all elements were updated. Seems like a problem with the change detection.

FlorianGlombik (approved):
Coloring worked fine, tested on http://134.209.254.13:8888/

I've got two suggestions that could be addressed in follow up PRs
1. An option to clear all colors at once might be a use case
2. It would be nice if the color of the descriptions could be adjusted on a finer granular level. E.g. just one side or even just one multiplicity.
AppollonColorChanges

PR #182, opened 2021-08-03
Fix package size issue

iozbeyli:
### Description
When there are two container type elements like package, activity side by side, the size of the right one was getting bigger unnecessarily, this PR fixes this issue. 

### Steps for Testing
Create a model with two container type element with normal elements inside then export it.

#### Before
image
#### After
image

### DIFF
--- a/src/main/packages/common/uml-package/uml-package.ts
+++ b/src/main/packages/common/uml-package/uml-package.ts
@@ -5,7 +5,7 @@ import { computeBoundingBoxForElements, IBoundary } from '../../../utils/geometr
 import { Text } from '../../../utils/svg/text';
 
 export abstract class UMLPackage extends UMLContainer {
-  render(layer: ILayer, children: ILayoutable[] = [], noChildren?: boolean): ILayoutable[] {
+  render(layer: ILayer, children: ILayoutable[] = [], calculateWithoutChildren?: boolean): ILayoutable[] {
     const radix = 10;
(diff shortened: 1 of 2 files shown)


ndricimrr (approved):
Tested in testing session. Working well for me. 👍

MaisaBenSalah:
Tested it and it is working fine for me :+1:

src/main/packages/common/uml-package/uml-package.ts
```diff
@@ -5,7 +5,7 @@ import { computeBoundingBoxForElements, IBoundary } from '../../../utils/geometr
 import { Text } from '../../../utils/svg/text';
 
 export abstract class UMLPackage extends UMLContainer {
-  render(layer: ILayer, children: ILayoutable[] = []): ILayoutable[] {
+  render(layer: ILayer, children: ILayoutable[] = [], noChildren?: boolean): ILayoutable[] {
```
MDK5, line 8:
```suggestion
  render(layer: ILayer, children: ILayoutable[] = [], withoutChildren?: boolean): ILayoutable[] {
```
or
```suggestion
  render(layer: ILayer, children: ILayoutable[] = [], calculateWithoutChildren?: boolean): ILayoutable[] {
```

MDK5 (approved):
Tested in testing session on ts1, bug is fixed, just one minor comment about the code

PR #193, opened 2021-11-17
`Development`: Add id's

rs-maker:
### Checklist

### Motivation and Context
 The https://github.com/ls1intum/Artemis cypress e2e tests are having a tough time finding good stable selectors for all of the Apollon components. To avoid having to use unstable selectors we want to add some now. 

### Description
 The selectors i added are 
id="modeling-assessment-container"
id="modeling-editor-canvas"
id="modeling-editor-sidebar"

These idswill not affect the apollon editor at all but will help test the components functionality with cypress.

### Steps for Testing

You can try out the apollon editor and see that nothing has changed. Alternatively you can check the code and see that the ids are not selected at all by apollon code.

### Screenshots

Screenshot (13)
Screenshot (10)
Screenshot (11)

### DIFF
--- a/src/main/components/assessment/assessment.tsx
+++ b/src/main/components/assessment/assessment.tsx
@@ -61,9 +61,6 @@ class AssessmentComponent extends Component<Props, State> {
   render() {
     const { elements } = this.state;
 
-    /*
-     * The id selector is added for the github.com/ls1intum/Artemis cypress e2e tests to find the object more easily
-     */
     return (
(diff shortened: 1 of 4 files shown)


jhainzlschmid (approved):
Changes make sense. However I personally prefer to use the HTML id attribute instead of a custom attribute for Cypress (against the Cypress documentation). We already have a couple of id selectors. In my opinion it would be nicer to migrate everything to use ids instead of using ids and cypress attributes, but thats just an opinion.

Issue #71, opened 2019-07-16
[Feature]: Enable resize on each corner of an element


jwilland-tum:
Right now, resize is only available in the bottom right corner of an element. This can be frustrating with resizing packages. Therefore enable resize handlers on each corner of an element.

PR #197, opened 2021-12-16
Feature/resizing enabled on each corner

minrows:
### Checklist

### Motivation and Context
This feature enables resizing of an elements from all 4 corner.
Issue Number: https://github.com/ls1intum/Apollon/issues/71

### Description
Previously resizing of an element could be done only from the bottom right corner.
This feature will now enable users to resize it from all 4 corners.

### Steps for Testing
1. Open up Apollon and add any Elements
2. Try resizing from all corners

### Screenshots
video


martindevtum:
https://user-images.githubusercontent.com/63931720/153768521-0140482a-ebd0-41ef-a2fa-8fd18634cd9c.mov


When resizing from top to bottom, when the canvas is in a scrollable state we encounter this bug, where the element gets moved down the canvas.

Steps to reproduce:
1. Open editor
2. Insert any element into canvas
3. Make the canvas scrollable by moving it down the canvas
4. resize element from top corner

Reproduced on: Macbook, Google Chrome v. 98

martindevtum (changes requested):

martindevtum:
@minrows  
The issue from before seems fixed :)

Now there is still another unwanted behaviour. 
If you resize elements from any corner except the bottom-right one, so for all the corners you adjusted, to make the element as small as possible, the element repositions itself. 

Is:

https://user-images.githubusercontent.com/63931720/156809221-aceb7e5a-c537-4ce4-bd96-1428d9d15e31.mov


Should:
- Element should not reposition itself.

martindevtum (changes requested):
@minrows 

For `resizable.tsx`: 
Please do not copy and paste every function 4 times if they are all almost the same.
The old file was very clear with no redundant code, but after your changes it grew almost to four times the original size with a lot of the changes seeming very redundant. 
Could you please refactor this file again?

PR #204, opened 2022-02-15
enhancement: add additional es5 output to yarn prepare

martindevtum:
### Motivation and Context

Since we updated the ts output to es2020 but some files are currently still needed in commonjs syntax within the server package of apollon_standalone, we decided to also provide a commonjs output for apollon.
This makes it possible to integrate the upgraded apollon in apollon_standalone.

### Description

I created an additional tsconfig file to configure the es5 output.
yarn prepare will now create a es6 output in lib/es6/ and a es5 output in lib/es5.
Consuming projects can then integrate whichever version is needed. (node will find the required version on its own)

### Steps for Testing

This can be tested by running `yarn prepare` and taking a look at the created lib directory.

These changes should be tested in conjunction with my Apollon_standalone PR (https://github.com/ls1intum/Apollon_standalone/pull/24).


minrows (approved):
Looks good to me. 👍

krusche:
Before we merge this, we need to check that the import in Artemis is working correctly

martindevtum:
@krusche @minrows  I tested it in Artemis and it is working for me. Artemis is also automatically using the correct es6 files!

minrows:
> @krusche @minrows I tested it in Artemis and it is working for me. Artemis is also automatically using the correct es6 files!

I too am currently testing and will update you in a few moments.

minrows:
@martindevtum 
Meanwhile, could you please quickly merge the latest develop branch to this branch?

minrows:
@martindevtum 
It seems like I am having 4 failed test cases. Could you make sure same is not the case for you?
Apart from that everything else is working fine. 

this

martindevtum:
> @martindevtum It seems like I am having 4 failed test cases. Could you make sure same is not the case for you? Apart from that everything else is working fine.
> 
> this

Thanks, good catch. This issue arises as the lib structure has now changed. As it is now containing subdirectories /es5 and /es6. Imports like that (that are used very sparingly) have to be adjusted to /lib/es6/.... . I will create an integration branch in Artemis.

bassner (approved):
I've tested this locally and verified that both ES5 and ES6 outputs are created. It works correctly in conjunction with the standalone version as well; the browser used ES6.

For Artemis, we need to create a PR that bumps the version *and* updates deep imports in some spec test files to import from the ES5 directory as Artemis is running Jest in CommonJS mode. However, only test files are affected. I've tested this PR with Artemis locally and Apollon works normally within Artemis.

krusche (changes requested):
Can we add to the README that the library has two output formats and describe shortly (with some links) what the difference would be in terms of usage?

martindevtum:
> Can we add to the README that the library has two output formats and describe shortly (with some links) what the difference would be in terms of usage?

I added it into the readme. But really nothing should change in terms of usage. the correct entrypoint will be automatically resolved depending on the nature of the consuming project. The only thing that is breaking, would be already existing deep imports, but there are very few of them in Apollon_standalone and Artemis and I adjusted them in each project accordingly in the respective PRs.

README.md
```diff
@@ -38,7 +38,12 @@ To unmount the editor instance, call its `destroy()` method:
 editor.destroy();
 ```
 
-For a complete overview of the API, please refer to the [lib/index.d.ts](https://github.com/ls1intum/Apollon/blob/master/lib/index.d.ts) file.
+For a complete overview of the API, please refer to the [lib/es6/index.d.ts](https://github.com/ls1intum/Apollon/blob/master/lib/index.d.ts) file.
```
krusche, line 41: please double check the link here, we do not use master as branch any more

README.md
```diff
@@ -38,7 +38,12 @@ To unmount the editor instance, call its `destroy()` method:
...
+
+### ESModules and CommonJs
+
+Apollon provides both an ESModules, as well as CommonJS version to be included. 
+The lay in `lib/es6` and `lib/es5` and the correct version should be resolved automatically.
```
krusche, line 46: "The lay" sounds like a typo, please formulate it correctly and mention that deep links need to distinguish the version.

We should check if we can avoid deep links in existing external projects

README.md
martindevtum: I think this information is outdated. My best guess is that they want to link to the lib/apolllon-editor.d.ts file. But since the lib directory is inside the .gitignore (rightfully so), we can't really link to the file on github as far as I can see. I would just remove the github link here.

README.md
martindevtum: I will have a look at the deep imports in Apollon_standalone and Artemis. I think they should be avoidable. In Artemis they were only used in test files anyways.

README.md
krusche: ok, please remove the link then

martindevtum: removed it

PR #207, opened 2022-03-14
Feature/resizing from all corners

minrows:
### Checklist

### Motivation and Context
This feature enables resizing of an elements from all 4 corner.
Issue Number: https://github.com/ls1intum/Apollon/issues/71

### Description
Previously resizing of an element could be done only from the bottom right corner.
This feature will now enable users to resize it from all 4 corners.

### Steps for Testing
1. Open up Apollon and add any Elements
2. Try resizing from all corners

### Screenshots
screen-capture (5)


src/main/components/uml-element/resizable/resizable.tsx
```diff
@@ -60,39 +91,92 @@ export const resizable =
...
       };
 
-      private onPointerDown = (event: React.PointerEvent<SVGElement>) => {
+      private onPointerDown = (event: React.PointerEvent<SVGElement>, resizeFrom: string) => {
         if (event.nativeEvent.which && event.nativeEvent.which !== 1) {
           return;
         }
 
-        this.setState({ resizing: true, offset: new Point(event.clientX, event.clientY) });
+        switch (resizeFrom) {
```
martindevtum, line 134: here i would create a `let offset` which we then assign in the switch statement.
Thus we can have one `this.setState({ resizing: true, offset })` in the end.

martindevtum: exactly as you did in the onPointerMove function.

martindevtum:
Code looks good! Way nicer.

src/main/components/uml-element/resizable/resizable.tsx
minrows: @martindevtum 
Done.

martindevtum:
Still one more bug:

**Currently:**
For Elements that can only be scaled in their Y-Axis, e.g. Petri Net Transitions, those elements get repositioned when resized from the Top Left and Bottom Left Corners. 

**Should Be:**
Equivalently to the Bottom Right and Top Right Corners, an element should not be repositioned when they are being resized.


Other than that it looks and feels great.

martindevtum (changes requested):
fix remaining bug.

src/main/services/uml-element/uml-element.ts
```diff
@@ -73,6 +73,7 @@ export abstract class UMLElement implements IUMLElement, ILayoutable {
   strokeColor?: string;
   textColor?: string;
   assessmentNote?: string;
+  resizeFrom = '';
```
martindevtum, line 76: I would propose to create an enum for this containing the four elements 
TOP_RIGHT,
TOP_LEFT,
BOTTOM_RIGHT,
BOTTOM_LEFT.

It is a bit more descriptive and overall better pratice.

src/main/services/uml-element/resizable/resizing-types.ts
```diff
@@ -12,6 +12,7 @@ export type ResizingActions = ResizeAction;
 export type ResizeAction = Action<ResizingActionTypes.RESIZE> & {
   payload: {
     ids: string[];
+    resizeFrom: string;
```
martindevtum, line 15: Should be an enum, see my other comment.

src/main/services/uml-element/resizable/resizing-reducer.ts
```diff
@@ -1,8 +1,37 @@
 import { Reducer } from 'redux';
+import { IBoundary } from '../../../utils/geometry/boundary';
 import { Actions } from '../../actions';
 import { ResizingActionTypes, ResizingState } from './resizing-types';
 
 export const ResizingReducer: Reducer<ResizingState, Actions> = (state = {}, action) => {
+  const obj = { x: 0, y: 0 };
+
+  const getUpdatedPosition = (elem: IBoundary, payload: any, resizeFrom: string) => {
```
martindevtum, line 9: we need to restrict the new positions to not be smaller or greater than the opposing corners via Math.max or Math.min.

Sent you the specifics.

src/main/components/uml-element/resizable/resizable.tsx
```diff
@@ -60,39 +91,94 @@ export const resizable =
...
+            />
           </WrappedComponent>
         );
       }
 
-      private resize = (width: number, height: number) => {
-        width = Math.round(width / 20) * 20;
-        height = Math.round(height / 20) * 20;
+      private resize = (width: number, height: number, resizeFrom: string) => {
+        width = Math.round(width / 5) * 5;
```
martindevtum, line 119: I'd think about if we should reset it to 20, now that we found another solution for the repositioning issue.

martindevtum (changes requested):

martindevtum:
feature works as expected.
I approve of the code changes.
👍

martindevtum (approved):

bassner (approved):
I've tested this manually and it works fine. Code looks good to me 👍

Issue #208, opened 2022-03-14
Boundary of some elements cease to exist while resizing.


minrows:
### Describe the bug
While resizing certain elements (like: Object of Activity Diagram), they cease to exist when reducing their dimension.

#### To Reproduce
1. Open Apollon
2. Select Activity Diagram from Diagram Type
3. Drag Activity Action node to the canvas and start reducing its width/height

#### Expected behavior
The minimum width of the elements should be set based on the name assigned to them. (For instance: the one of UML Package)

#### Screenshots
![screen-capture (6)](https://user-images.githubusercontent.com/14681902/158262278-1be2e2cf-6d49-495c-a5cd-5bf70dc17ee9.gif)

### Environment

<details><pre>
 - OS: Windows
 - Browser: Chrome
 - Version 99.0.4844.51
</pre></details>

PR #210, opened 2022-03-21
Bugfix: Boundary of some elements cease to exist while resizing

minrows:
…ments

### Checklist

### Motivation and Context
To specify the named bounds so that the issue with vanishing element boundary, while resizing is resolved.
Issue Number: https://github.com/ls1intum/Apollon/issues/208

### Steps for Testing
1. Open Apollon
2. Select Activity Diagram from Diagram Type
3. Drag Activity Action node to the canvas and start reducing its width/height
4. Observe that the element now no longer vanishes while its side is reduced as it used to (see: https://github.com/ls1intum/Apollon/issues/208 ).

### Screenshots
screen-capture-_7_

### DIFF
--- a/src/main/packages/flowchart/flowchart-decision/flowchart-decision.tsx
+++ b/src/main/packages/flowchart/flowchart-decision/flowchart-decision.tsx
@@ -9,8 +9,7 @@ export class FlowchartDecision extends UMLElement {
   type: UMLElementType = FlowchartElementType.FlowchartDecision;
 
   render(canvas: ILayer): ILayoutable[] {
-    const namedBounds = calculateNameBounds(this, canvas);
-    this.bounds = namedBounds;
+    this.bounds = calculateNameBounds(this, canvas);
     return [this];
   }
 }
--- a/src/main/packages/flowchart/flowchart-decision/flowchart-decision.tsx
+++ b/src/main/packages/flowchart/flowchart-decision/flowchart-decision.tsx
@@ -2,12 +2,15 @@ import { FlowchartElementType } from '..';
 import { ILayer } from '../../../services/layouter/layer';
 import { ILayoutable } from '../../../services/layouter/layoutable';
 import { UMLElement } from '../../../services/uml-element/uml-element';
+import { calculateNameBounds } from '../../../utils/name-bounds';
 import { UMLElementType } from '../../uml-element-type';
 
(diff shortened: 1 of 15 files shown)


src/main/packages/flowchart/flowchart-decision/flowchart-decision.tsx
```diff
@@ -2,12 +2,15 @@ import { FlowchartElementType } from '..';
...
 import { UMLElement } from '../../../services/uml-element/uml-element';
+import { calculateNameBounds } from '../../../utils/name-bounds';
 import { UMLElementType } from '../../uml-element-type';
 
 export class FlowchartDecision extends UMLElement {
   type: UMLElementType = FlowchartElementType.FlowchartDecision;
 
   render(canvas: ILayer): ILayoutable[] {
+    const namedBounds = calculateNameBounds(this, canvas);
+    this.bounds = namedBounds;
```
martindevtum, line 13: I would personally skip the creation of the constant and just assign to this.bounds directly, as we are not using namedBounds anywhere else

src/main/utils/name-bounds.ts
```diff
@@ -0,0 +1,16 @@
+import { ILayer } from '../services/layouter/layer';
+import { IBoundary, computeBoundingBoxForElements } from '../utils/geometry/boundary';
+import { Text } from '../utils/svg/text';
+
+export function calculateNameBounds(element: any, layer: ILayer) {
```
martindevtum, line 5: shouldn't we add the return type IBoundary here?

martindevtum (changes requested):
I would also like it if you would use the newly created name bounds function to already existing usages of nameBounds. 

It is for example used in uml-package.ts, please calculate the nameBounds here with our new function. 
This might be the only occurence, but please just. double check it again.


Other than that, it is working nicely.

bassner:
I tested this locally and it seems to work well, except for packages (class diagram), which can still cut off their text content vertically:
grafik
But I'm not 100% sure if that was even the point of this PR or if it was just about keep the correct width on certain objects?

minrows:
> 

Done. Minimum height is now calculated based on the name as well.

bassner (approved):
LGTM!

martindevtum:
looking good

martindevtum (approved):

PR #211, opened 2022-03-25
Enhancements/add cypress

martindevtum:
### Checklist

### Motivation and Context

We want to increase automatic test coverage for Apollon.
We want to create e2e tests in Apollon, to have greater confidence about the quality of Apollon and lower the risk of introducing bugs. 
This PR sets up cypress for Apollon.

### Description

I configured cypress and wrote a first small test with commands and a plugin to show that the setup works.
I also had to move all unit testing files into a new directory /tests/unit and adjust all paths.
 
The test folder structure for our project is now:
```
src
    -- main
    -- stories
    -- tests
          -- unit  
          -- e2e
```

This PR provides the basis for further test coverage improvements.
We should also going forward create e2e tests for features/bugfixes we implement.

### Steps for Testing

In order to test this, one would have to run unit tests to ensure they are still working and run the newly created cypress test.
We do that by:

```yarn test  &&  yarn test:e2e```


minrows (approved):
Created cypress test is working as expected and code looks good to me.

martindevtum:
> I think it looks good in general.
> 
> These changes are disturbing somehow. Can we investigate if we can just define absolute paths for imports to avoid this in the future? image

good point. It was relative before so I just kept it that way. But might be worth considering.

PR #219, opened 2022-04-24
Create a docker file and github action for Apollon.

dangcosta:
### Motivation and Context

Apollon had no docker image before. 

### Description

Added a Dockerfile and a GitHub action for building and pushing the image to ghcr.io. 

### Steps for Testing

```bash
docker run -it --rm -p 8888:8888 ghcr.io/ls1intum/apollon:dev
```

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - main
-      - feature/dockerize
+      - develop
       - release/*
     tags: '*'
   release:
@@ -33,7 +33,7 @@ jobs:
               if (context.ref === "refs/heads/main") {
                 return "latest";
               }
-              if (context.ref === "refs/heads/feature/dockerize") {
+              if (context.ref === "refs/heads/develop") {
                 return "dev";
               }
             }
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -33,6 +33,9 @@ jobs:
               if (context.ref === "refs/heads/main") {
                 return "latest";
               }
+              if (context.ref === "refs/heads/feature/dockerize") {
+                return "dev";
+              }
             }
             return "FALSE";
 
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,0 +1,71 @@
+name: Build and Push
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - feature/dockerize
+      - release/*
+    tags: '*'
+  release:
+    types:
+      - created
+
+jobs:
+  buildx:
+    name: Build and Push the Docker Images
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,74 @@
...
+          script: |
+            if (context.eventName === "pull_request") {
+              return "pr-" + context.issue.number;
+            }
+            if (context.eventName === "push") {
+              if (context.ref.startsWith("refs/tags/")) {
+                return context.ref.slice(10);
+              }
+              if (context.ref === "refs/heads/main") {
+                return "latest";
```
Mtze, line 34:
```suggestion
                return "stable";
```

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,74 @@
...
+              }
+              if (context.ref === "refs/heads/develop") {
+                return "dev";
```
Mtze, line 37:
```suggestion
                return "latest";
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:18-slim
+
+WORKDIR /app
+
+COPY . .
```
Mtze, line 5: Please only add the bundled application here

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN yarn install
+
+EXPOSE 8888/tcp
+
+CMD ["yarn", "start"]
```
Mtze, line 11: In the current production env we use 
```
/usr/bin/node /opt/apollon/build/server/bundle.js
```
To start the application - Maybe someone with more experience with Apollon could explain the difference? From my understanding `yarn start` is only meant for development. 



I guess we also need some env variables as well - At least we have 
```
NODE_ENV=production
``` 
on the prod server

Mtze (changes requested):
Please add a build step (to the `build.yml`) which bundles the application and publishes an artifact to GitHub:
```
yarn install
yarn build
tar cfzv apollon_standalone.tar build
```
and depends the new docker build pipeline on the old. Then you adapt the `Dockerfile` to only include the app and serve it with node.

Dockerfile
kliwniloc: I propose we build the bundled application in the container instead of the using the artifact from the build pipeline since otherwise the Dockerfile will be dependent on external builds (the build pipeline) and users will have a harder time trying to build the image themselves.

kliwniloc (changes requested):
~~Maybe add a `.dockerignore` file~~
No longer necessary since the `builder` is discarded anyways.

Dockerfile
kliwniloc: `yarn build` uses webpack to produce a `dist` folder with the static content, so we don't need to run any node application as far as I can tell.
We decided to use that static content in the Dockerfile and add an http server (nginx) resulting in an Image size of just 23MB.

PR #223, opened 2022-05-04
Development: Update dependencies of Apollon

minrows:
### Description
This PR updates the dependencies of the Apollon.


martindevtum (approved):
tested it. Everything is working as expected. No errors or warnings.
pinst is still doing what we use it for (only major version update).

bassner (approved):
Tested this and did not run into any problems 👍

PR #227, opened 2022-05-08
fix: ThemedLine in dark mode

mariusdkm:
This fixes the stroke Color of the ThemedLine, which didn't change color in the newly introduced dark mode.
In the context of the Component Actor in the Use Case Diagram the change looks like this:

Before:
Screenshot 2022-05-08 at 12 06 15

After:
Screenshot 2022-05-08 at 12 02 31

### Checklist

### Motivation and Context

Darkmode is great, but only if it looks right.

### Description

Changed the props `strokeColor` and `stroke` to match the ThemedCircle, which already had the right colors.

### Steps for Testing

1. Open the use case diagram and see the changes :)

### DIFF
--- a/src/main/components/theme/themedComponents.ts
+++ b/src/main/components/theme/themedComponents.ts
@@ -96,11 +96,9 @@ export const ThemedEllipse = styled.ellipse.attrs(
   stroke: ${(props) => props.strokeColor || props.theme.color.primaryContrast};
 `;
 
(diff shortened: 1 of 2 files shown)


minrows (commented):
Everything is working fine, except for the test cases. 
Could you please make sure that all test snapshots are updated?

minrows (approved):
LGTM. 👍

Issue #93, opened 2020-03-24
More flexible layout


TobiasPr:
### Is your feature request related to a problem?
When creating a diagram, the relationship between UML elements are sometimes not layouted in a good looking way

### Describe the solution you'd like
Make layout of relationships between UML elements more flexible (not drawn automatically). Use layout from current algorithm as default position for elements. Add possibility to add waypoints, similar to screenshot.

![more_flexible_layout](https://user-images.githubusercontent.com/22955066/77404760-7f907000-6db2-11ea-9b73-a9a23d97a381.png)

PR #253, opened 2022-07-27
`Development`: Adding manual waypoints feature

minrows:
### Checklist

### Motivation and Context
This PR introduces new feature in Apollon. With this PR, users can add waypoints in connectors while drawing diagrams.
Feature requested issue Number: https://github.com/ls1intum/Apollon/issues/93

#### In local machine:
1. Clone the repo
2. Checkout out `feature/manual-waypoints` branch
3. Start application by executing `yarn install` followed by `yarn start` command
4. Add and connect any elements in canvas 
5. Observe waypoints handles and drag to change them

#### In test server
- [x] Will be deployed in test server after releasing alpha version
- [x] Disable Manual layouting in "Interactive Mode"
- [x] Fix NaN Issue
- [x] Fix RTC issue

### Screenshots
screen-capture (21)

#### Additional Info
Changes released in [2.11.4-alpha.0](https://www.npmjs.com/package/@ls1intum/apollon/v/2.11.4-alpha.0)


akesfeden:
Tested on test server, in the screenshot below I can only see (and interact with) 2 connectors with waypoints, the others are not draggable. Is there a specific condition to create waypoints or all connectors should have them?

image

akesfeden:
I see, thanks for the explanation. In the case below, wouldn't it be nicer if we can also move the segments that are directly connecting to the components? Maybe that can be considered as an enhancement.

image

minrows:
> I see, thanks for the explanation. In the case below, wouldn't it be nicer if we can also move the segments that are directly connecting to the elements? Maybe that can be considered as an enhancement.
> 
> image

- User can still change the position of connectors (top, bottom, left and right) for the case of line segments that are directly connecting to the elements.
- And if allowed, the only position that user could drag line segment connecting to elements would be the width (for top and bottom) and height (for left and right) of an element, so I personally don't think that would give much flexibility.
- But yes, it could be considered as an enhancement in follow up PR. Thanks for the feedback (:

akesfeden (approved):
Tested on test server, works as intended, a possible enhancement idea is discussed in the comments.

Santia-go (approved):
Tested on TS1
Working as expected

https://user-images.githubusercontent.com/88991954/183946167-661a2993-ca57-42c7-9b40-a5300804da37.mov

Issue #279, opened 2022-12-17
Misalignment of elements of Object, Use case, Flowchart and Petri Net diagram in Artemis


minrows:
### Describe the bug
Elements such as the Object of Object/Communication diagram, Actor of Usecase diagram, Input/Output of Flowchat and Transition of Petri Net diagrams are not aligned perfectly to the grid.

#### To Reproduce

1. Insert object, actor, and transition elements in the canvas
2. Observe that the specified elements are not aligned perfectly to the grid. 

#### Expected behavior
All the specified elements should be aligned perfectly with the grid. 

#### Screenshots
![image](https://user-images.githubusercontent.com/14681902/208219246-8404e4d4-0f3e-4908-9f2a-c89f4eddc9ef.png)
![image](https://user-images.githubusercontent.com/14681902/208486895-c1d6125f-35e5-4f79-bd97-455dba95b4d4.png)

PR #280, opened 2022-12-21
`Development`: Fix alignment of elements to grid issue 2

mertyldrr:
### Checklist

### Motivation and Context
This PR fixes the issue of elements not aligning perfectly on the grid.
Issue Number: https://github.com/ls1intum/Apollon/issues/279

**Changes include:**
- Adjust the coefficients to produce a multiple of ten (each square on the grid has 10px width & height)
- Fix alignment issue with the following elements:
1. Object in object diagram and communication diagram
2. Actor in a use case
3. Transition and place in a Petri net
4. Input/Output in a flowchart
5. All elements of a class diagram (not mentioned in the ticket but these were also not perfectly aligned)

### Steps for Testing
1. Clone the repo
2. Checkout to branch `enhancement/uml-elements-scaling-fix`
3. Execute commands: `yarn install`, followed by `yarn start`
4. Observe elements mentioned above are aligned with the grid.

#### Before
class-develop
others-dev

#### After
class-fixed
others-fixed


src/main/utils/geometry/boundary.ts
```diff
@@ -47,3 +47,11 @@ export function computeBoundingBoxForElements(elements: { bounds: IBoundary }[])
   const height = Math.max(...boundaries.map((bounds) => bounds.y + bounds.height)) - y;
   return { x, y, width, height };
 }
+
+export function computeDimension(scale: number, value: number, isCircle?: boolean): number {
```
minrows, line 51: Could you maybe use the scale factor from `getScaleFactor` function (defined [here](https://github.com/ls1intum/Apollon/blob/05ab7674add6700bff26939a7299d9308fa3b9e1/src/main/apollon-editor.ts#L282)), instead of using it as parameter?

minrows (commented):
- Tested locally and in scale factor of 0.8 and 1. 
- Works as expected. 👍 
- Overall code changes LGTM as well. 
- Left a minor suggestion.

src/main/utils/geometry/boundary.ts
mertyldrr: Thanks for the suggestion @minrows, it makes sense, but we don't have access to editor.getScaleFactor, and using editor from the state would add more complexity since this is not a react function/component.

minrows (approved):

akesfeden (approved):
Tested locally, all elements mentioned in Changes include section are aligned perfectly with the grid (tested with various browser zoom levels including 100%).

PR #286, opened 2023-03-13
`Development`: Adapt tests to React 18

mertyldrr:
### Motivation and Context
This PR updates testing-library/react and modify test. Tests will run closer to what real users would experience when using Apollon.

### Steps for Testing
1. Clone the repo
2. Execute commands: `yarn install`, `yarn build` followed by `yarn start`
3. Observe application works as expected.


package.json
```diff
@@ -29,8 +29,8 @@
     "prepare": "rm -rf lib && tsc -p . && tsc -p tsconfig.es5.json",
     "prettier:check": "prettier --check 'src/{main,tests}/**/*.{json,ts,tsx,js,jsx,css,scss,html}'",
     "prettier:write": "prettier --write 'src/{main,tests}/**/*.{json,ts,tsx,js,jsx,css,scss,html}'",
-    "test": "jest --maxWorkers=50%",
-    "test:coverage": "jest --collectCoverage --maxWorkers=50%",
+    "test": "jest -w=4",
+    "test:coverage": "jest --collectCoverage -w=4",
```
pal03377, line 33: What does this do? Is there some longer flag name so that it's obvious more quickly?

src/main/utils/svg/multiline.tsx
```diff
@@ -31,7 +31,7 @@ export class Multiline extends Component<Props, State> {
   spaceWidth = 0;
   wordsWithComputedWidth = [] as { word: string; width: number }[];
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
```
pal03377, line 34: What is this change for? Also, both of these functions are apparently deprecated.

src/main/apollon-editor.ts
```diff
@@ -90,24 +91,26 @@ export class ApollonEditor {
...
+  ): Promise<Apollon.SVG> {
+    const container = document.createElement('div');
+    const root = createRoot(container);
     const element = createElement(Svg, { model, options, styles: theme });
-    const svg = render(element, div);
-    const { innerHTML } = div;
-    unmountComponentAtNode(div);
+    const svg = new Svg({ model, options, styles: theme });
+    root.render(element);
+    await delay(500);
```
pal03377, line 104: I think this should be some kind of waiting function for rendering, a hardcoded delay is a bit unclean.

pal03377: I suggest replacing this with `await delay(0)` and adding a commend on why you need to wait for one render cycle.

pal03377 (changes requested):
The program & tests are both still working, I would only suggest 2 code clean-ups (see below)

src/tests/unit/components/update-pane/update-pane-test.tsx
```diff
@@ -121,10 +125,17 @@ describe('test update pane', () => {
...
 
-    jest.runAllTimers();
-    fireEvent.pointerDown(container.querySelector('svg')!);
+    act(() => {
+      jest.runAllTimers();
+    });
+
+    act(() => {
+      fireEvent.pointerDown(container.querySelector('svg')!);
+    });
```
FelixTJDietrich, line 138: You are breaking up `act(...)` statements into multiple ones, why not use one? As far as I can tell from the stuff I saw, it should be fine to wrap all the code in one `act` statement.

FelixTJDietrich (commented):
Other than that and the comments from Paul, I can't find other issues :)

package.json
mertyldrr: --maxWorkers=50% = uses 50% of available cores
-w = uses 4 cores

src/tests/unit/components/update-pane/update-pane-test.tsx
mertyldrr: Thanks for the suggestion Felix. Initially tried the approach you suggested but didn't work at all. Usually order is important in Apollon. We have to create the application and then use getter and setter functions to get/modify application state. Using single act probably breaks this logic.

FelixTJDietrich: Yeah, you are right. The way you did it is probably safer 👍

FelixTJDietrich (approved):
I tested it again after the changes and everything passes and works as expected for me. Good job! :)

package.json
pal03377: I think you can do `--maxWorkers=4`, which would be more readable and [is just an alias](https://jestjs.io/docs/cli#--maxworkersnumstring) 🤔

pal03377 (approved):
Changes look good & tests are running! 🎉

PR #295, opened 2023-07-14
Fix assessment subscription in `apollon-editor.ts`

pal03377:
### Checklist
- [ ] I documented the TypeScript code using JSDoc style.
- [ ] ~I added multiple screenshots/screencasts of my UI changes~
- [ ] ~I translated all the newly inserted strings into German and English~

### Motivation and Context

We have a bug in Artemis where all tutor assessments get lost on save. Artemis sometimes does not get events for new assessments. Therefore, assessments can often get lost, especially in Safari.

### Description

Change `.splice(i)` (which removes all elements in the array starting from `i`) with an object-key-based approach. This additionally prevents problems from occurring when indexes shift because of `splice`.

Also, a subscription to `this.store` in `apollon-editor.ts` did sometimes not work because `this.store` was still undefined when it was called. I added code to automatically retry the subscription when if fails because of that.

### Steps for Testing

1. Deploy https://github.com/ls1intum/Artemis/pull/6901 (has version of the npm package with this fix in it)
2. Assess modeling exercise participation
3. Assess components
4. Press "Save" or "Override Assessment"
5. Assessments are no longer lost. They were gone previously.

For better reproducibility, test in Safari. In Safari, the bug occurs more often (closer to 100% of the time). In Chrome, the issue sometimes does not appear at all.

### Test Coverage

| File | Line |
|------|-------:|
| apollon-editor.ts | 79.59 % |

### DIFF
--- a/src/main/apollon-editor.ts
+++ b/src/main/apollon-editor.ts
@@ -310,6 +310,8 @@ export class ApollonEditor {
     setTimeout(() => {
       if (this.store) {
         this.store.subscribe(this.onDispatch);
+      } else {
+        setTimeout(this.componentDidMount, 100);
       }
     });
   };
--- a/src/main/apollon-editor.ts
+++ b/src/main/apollon-editor.ts
@@ -114,11 +114,11 @@ export class ApollonEditor {
   private currentModelState?: ModelState;
   private assessments: Apollon.Assessment[] = [];
   private application: RefObject<Application> = createRef();
-  private selectionSubscribers: {[key: number]: (selection: Apollon.Selection) => void} = {};
-  private assessmentSubscribers: {[key: number]: ((assessments: Apollon.Assessment[]) => void)} = {};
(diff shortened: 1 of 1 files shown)


RY997 (approved):
Worked well after deploying to artemis on ts2

rriyaldhi (approved):
Tested on TS4 using Safari, the score and feedback were saved using both Save and Override buttons.

rriyaldhi (commented):
Just realized that adding a point to a class doesn't enable the save button. But deleting it enables the save button. I don't know if this is related to this PR.

https://github.com/ls1intum/Apollon/assets/10885503/0a93a822-f522-487a-9e3b-ee589e4c0ee8

laadvo (commented):
Tested on ts4, experienced the same thing as @rriyaldhi. The save button is disabled for the first feedback. Otherwise the feedbacks get saved. Here is the screencast I made

https://github.com/ls1intum/Apollon/assets/73339358/793bbea3-2470-418d-ac4c-24b12100e2bb

PR #296, opened 2023-07-22
Enable Artemis to hook into the React render timeline for more reliable tests

pal03377:
### Checklist
- [X] I documented the TypeScript code using JSDoc style.

### Motivation and Context

To fix flaky client tests related to Apollon in #6955, I need to be able to wait for the next render cycle in React. To my knowledge, this is not possible without some "help" from the React code itself (or with a React testing library, which we don't use). Therefore, we need this PR.

### Description

Added a new promise to the `ApollonEditor` called `nextRender`, which resolves when the next scheduled React render of the editor component happens. It also ensures that `ApollonEditor.store` will be properly initialized before it resolves.

### Steps for Testing

- Test https://github.com/ls1intum/Artemis/pull/6955 using `@palsch/apollon:2.14.12`, where this PR is available as an NPM package.
- Run `npm start` locally to check that the standalone editor still works.

### Test Coverage

| File | Line |
|------|-------:|
| apollon-editor.ts | 82% |
| application.tsx | 100% |

### DIFF
--- a/src/main/apollon-editor.ts
+++ b/src/main/apollon-editor.ts
@@ -119,11 +119,11 @@ export class ApollonEditor {
   private modelSubscribers: { [key: number]: (model: Apollon.UMLModel) => void } = {};
   private discreteModelSubscribers: { [key: number]: (model: Apollon.UMLModel) => void } = {};
   private errorSubscribers: { [key: number]: (error: Error) => void } = {};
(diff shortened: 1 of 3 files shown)


pal03377:
We tested this through https://github.com/ls1intum/Artemis/pull/6955 in a testing session.

PR #298, opened 2023-08-03
UI Changes to Control Panel

AlexanderG2207:
### Checklist

### Motivation and Context
1. When scrolling between the elements of the switches in the control panel, there is no border on the left side of the middle elements. 
2. When switching between Light and Dark Mode there is the obvious indicator of the change, which is the editor changing its color. However, there is no indication on the Light/Dark switch buttons itself.

### Description
1. By removing the "border-left" property from the switch styles, the left border of the element is normal and the change between switch elements looks cleaner.
2. By clicking one of the Light/Dark switches, the `setThemings()` function is called, which takes a string of the button that was clicked. The function then adds the "selected" style property to the selected button and removes the "selected" style property from the unselected one.

### Steps for Testing
1. Run Apollon locally.
2. Open Apollon in Chrome.
3. Check if the borders transition well on the switches when moving between them.
4. Check if the button is selected when changing between light and dark mode.

### Before:
Before

### After:
After


loreanvictor:
down the line, perhaps the control panel should be switched from static HTML / JS to a proper React component, but I feel that falls beyond the scope of this PR.

loreanvictor (approved):

florian-glombik (commented):
The change itself looks fine and works for me on the local setup, however, it just works after adjusting `_getNewSubscriptionId` in `apollon-editor.ts` to use a radix in `parseInt`

Can you reproduce the same issue? - If yes It would be great to just fix it right away!
_Otherwise I am pretty confused as to why it only appears to be an issue for me when checking out bef2b891d33e4b5e4b778829cbaa9e4b82931ae7_

I would also welcome the change to a tsx component, but agree that it can be done within another PR :)

florian-glombik (changes requested):
It seems like a test needs to be fixed

```
Summary of all failing tests
FAIL src/tests/unit/apollon-editor-test.tsx (96.662 s)
  ● test apollon editor  › exportModelAsSvg

    TypeError: Cannot read properties of null (reading 'outerHTML')

      105 |
      106 |     return {
    > 107 |       svg: replaceColorVariables(container.querySelector('svg')!.outerHTML),
          |                                                                 ^
      108 |       clip: svg.state.bounds,
      109 |     };
      110 |   }
```

PR #292, opened 2023-05-25
Fix SVG and PDF Export

tobias-lippert:
### Checklist
- [x] I documented the TypeScript code using JSDoc style.
- [x] I added multiple screenshots/screencasts of my UI changes
- [x] I translated all the newly inserted strings into German and English

### Motivation and Context

Currently, the svg and pdf exports do not render the diagrams 100% correctly.

### Steps for Testing

1. Create a diagram or multiple if you want to and export it as svg and pdf. Make sure all lines that should be visible are not cut off, e.g the borders of class diagrams.

### Test Coverage

No tests yet.
| File | Branch | Line |
|------|-------:|-----:|

### Screenshots

Before diagram created with apollon standalone pdf:
image
Now:
image

Before: diagram created on artemis using apollon:
image

After:
image

More complex example:
image

PR #299, opened 2023-08-03
Fix Element borders when exporting to SVG

AlexanderG2207:
### Checklist

### Motivation and Context
When creating a model and exporting it to an SVG, the borders of class elements are not rendered correctly (Before Image). This leads to class elements looking different when exported compared to what the user had modeled in the editor. 

#292 had an approach that reduced the element height if it was a class member. I found that the issue was regarding the border rect and path being rendered before the class members were rendered, making the class members overlap the rect and path and thus making the borders invisible. That is why a class diagram doesn't have any rendering issues with its border in the editor because the members are rendered before the rect and path (Apollon Editor Image).

### Description
To fix this issue, the render order was changed for class members. Before, each Component for the SVG was rendered by creating each element and adding it to a map (Current SVG Export Image). With UMLClassifierComponents, the UMLMemberComponents were generated after the UMLClassifierComponent, making them block the rect and paths of the UMLClassifierComponent since they are rendered in the UMLClassifierComponent.

The UMLClassifierComponent allows ReactNodes (children) to be placed within the Component. This then firstly renders the children and then secondly the rect and paths (Updated SVG Export Image). So, by first checking if the given Element is of type UMLClassifierComponent and then creating an array of all children and placing them within the UMLClassifierComponent, they are rendered before the borders. This results in the borders being visible and the SVG being exported correctly (After Image).

### Steps for Testing
1. Run Apollon locally.
2. Open Apollon in Google Chrome.
3. Create a test diagram with as many different diagram types as possible.
4. Export to SVG.
5. Take note of any border issues once the SVG is rendered.

### Test Coverage
The "class-diagram-as-svg.json" file was adjusted in the unit test resources to get all the tests to pass. This is because the whole order of SVG element rendering was changed, and therefore, the test diagram needed to be adjusted accordingly.

| File | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
|------|-------:|-----:|------:|-------:|-----:|
| svg.tsx | 81.36 | 70 | 66.66 | 81.36 | ...36-141,175-177 

### Screenshots
Apollon Editor | Current SVG Export | Updated SVG Export
:-------------------------:|:-------------------------:|:-------------------------:
Apollon Editor | Current SVG Export | Updated SVG Export

Before | After
:-------------------------:|:-------------------------:
Before | After


tobias-lippert (approved):
Tested locally with your fork. The issue seems to be fixed.

src/main/scenes/svg.tsx
```diff
@@ -212,19 +214,56 @@ export class Svg extends Component<Props, State> {
...
+                    width={element.bounds.width}
+                    height={element.bounds.height}
+                    key={element.id}
+                    className={element.name ? element.name.replace(/[<>]/g, '') : ''}
+                    fill={element.fillColor || theme.color.background}
+                  >
+                    <ElementComponent key={index} element={element} scale={this.props.options?.scale || 1.0} />
+                  </svg>
+                );
+              }
```
pal03377, line 266: What happens when neither of the `if` conditions are true?

src/main/scenes/svg.tsx
```diff
@@ -212,19 +214,56 @@ export class Svg extends Component<Props, State> {
...
+                          >
+                            <MemberElementComponent key={memberIndex} element={memberElement} scale={this.props.options?.scale || 1.0} />
+                          </svg>
+                        );
+                      })}
+                    </ElementComponent>
+                  </svg>
+                );
+                // Render all other UMLElements and UMLRelationships normally, as they dont have issues when rendering to SVG
+              } else if (ElementComponent !== UMLClassifierComponent && ElementComponent !== UMLClassifierMemberComponent) {
```
pal03377, line 252: I would move the comment into the `else`
```suggestion
              } else if (ElementComponent !== UMLClassifierComponent && ElementComponent !== UMLClassifierMemberComponent) {
                  // Render all other UMLElements and UMLRelationships normally, as they dont have issues when rendering to SVG
```

pal03377 (commented):
Well done! I tested locally and didn't encounter any border issues. I only have some minor comments about the code:

src/main/scenes/svg.tsx
```diff
@@ -212,19 +214,56 @@ export class Svg extends Component<Props, State> {
...
-                  width={element.bounds.width}
-                  height={element.bounds.height}
-                  key={element.id}
-                  className={element.name ? element.name.replace(/[<>]/g, '') : ''}
-                  fill={element.fillColor || theme.color.background}
-                >
-                  <ElementComponent key={index} element={element} scale={this.props.options?.scale || 1.0} />
-                </svg>
-              );
+              // If the ElementComponent is of type UMLClassifierComponent, create an array of all members (attributes and methods)
```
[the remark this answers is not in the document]
AlexanderG2207: I added more detailed comments for each switch case :)

src/main/scenes/svg.tsx
```diff
@@ -212,19 +214,61 @@ export class Svg extends Component<Props, State> {
...
+                  // Unlike other components, the UMLClassifierComponent needs its members to be children within the component to avoid border rendering issues.
+                  const members = elements.filter((member) => member.owner === element.id);
+                  return (
+                    <svg
+                      x={element.bounds.x}
+                      y={element.bounds.y}
+                      width={element.bounds.width}
+                      height={element.bounds.height}
+                      key={element.id}
+                      className={element.name ? element.name.replace(/[<>]/g, '') : ''}
```
florian-glombik, line 229: This happens 3 times within this component, might be worth it to use a method `convertElementNameToClass` or something similar to improve readability and reduce code duplication?

florian-glombik (approved):
Tested it locally, and the issue is resolved for me.

The code looks good!

I did have the same issue within `apollon-editor.ts` as described in https://github.com/ls1intum/Apollon/pull/298#pullrequestreview-1605732048, not sure why this only seems to appear for me :sweat_smile:

src/main/scenes/svg.tsx
AlexanderG2207: Good idea! I created a method for the duplicate code.

PR #302, opened 2023-09-23
Add initial support for canvas zooming

matthiaslehnertum:
This PR intends to extend the functionality of Apollon by the ability to zoom into the drawing canvas. For now, zooming is only possible via buttons on the canvas. Zooming via the trackpad has been excluded for now to reduce the increment size.

### Checklist

### Motivation and Context
This changes intends to solve the issues of element bounds being serialised including the scale factor at the moment which leads to issues when opening a diagram file in another Apollon instance with a different scale factor and leads to diagrams being exported wrongly from the conversion service if they have been serialised with any other scale factor than 1.0.

### Description
The change adds zoom controls to the drawing canvas. The sequel PR https://github.com/ls1intum/Apollon/pull/307 then also removes all instances where the previous scale factor implementation was used.

### Steps for Testing
To test the change, you can simply open up Apollon, drop a few elements on the drawing canvas and then use the zoom-in and zoom-out buttons the observe their effect.

### Test Coverage

| File | Branch | Line |
|------|-------:|-----:|

### Screenshots

Editor with the current default zoom of 0.8:
Screenshot 2023-10-04 at 11 06 25

Editor with the current minimum zoom factor of 0.5:
Screenshot 2023-10-04 at 11 06 30

Editor with a zoom factor of 1.6:
Screenshot 2023-10-04 at 11 06 43

### DIFF
--- a/src/main/components/canvas/editor.tsx
+++ b/src/main/components/canvas/editor.tsx
@@ -8,9 +8,7 @@ import { AsyncDispatch } from '../../utils/actions/actions';
 import { EditorRepository } from '../../services/editor/editor-repository';
 import { clamp } from '../../utils/clamp';
 import { ZoomPane } from './zoom-pane';
-import {IUMLElement} from '../../services/uml-element/uml-element';
 import {UMLElementState} from '../../services/uml-element/uml-element-types';
-import {computeBoundingBoxForElements} from '../../utils/geometry/boundary';
 
 const minScale: number = 0.5;
 const maxScale: number = 5.0;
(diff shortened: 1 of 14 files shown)


src/main/components/canvas/editor.tsx
```diff
@@ -71,6 +87,56 @@ const SCROLL_DISTANCE = 5;
...
+        behavior: 'smooth'
+      });
+
+      this.props.changeZoomFactor(newZoomFactor);
+    });
+
+    window.addEventListener('gestureend', function (event) {
+      event.preventDefault();
+    });
+    */
```
AlexanderG2207, line 138: If this comment is not needed, I would recommend removing it.

AlexanderG2207 (commented):
Nice feature, well done! The code looks good, and all unit tests pass locally when updating the snapshots.

I have two remarks:

1. I get these compilation problems when testing in Google Chrome.
Screenshot 2023-10-06 at 14 58 48

2. I would make the "zoom out" button the same size as the "zoom in" button.
Screenshot 2023-10-06 at 15 00 28

matthiaslehnertum:
Hey @AlexanderG2207 , thank you for your review. I addressed the issues accordingly. Could you check if the PR is fine for you like this?

AlexanderG2207 (approved):
Thanks for implementing the changes👍

PR #309, opened 2023-10-09
Move repository from yarn to npm

matthiaslehnertum:
As per Stephan's request, this PR updates all of Apollon's dependencies to a recent version, replaces yarn by npm as a package manager and replaces tslint with eslint.

At the moment, a lot of the eslint default rules are disabled to have eslint behave the same way as tslint behaved before. In the future, we should sequentially reenable these rules and fix the issues detected. To not make this PR unnecessarily large, I decided to not do this in the current PR.

### Motivation and Context
A lot of Apollon's dependencies, especially tslint, were outdated.

### Steps for Testing
The most sensible test for this change would be to test if running and building the application behaves as expected.


README.md
AlexanderG2207, line 1: I would still add that for local usage, npm can be used to install the dependencies and to build and test under the `Development Setup`.

AlexanderG2207 (commented):
Built and tested with npm locally. All tests pass and everything seems to be working as usual in the editor. I would just adjust the README to accommodate the npm implementation.

PR #313, opened 2023-10-20
Fix issues with the zoomfactor being passed to svg components

matthiaslehnertum:
This tiny fix corrects and issue with the custom 'zoomFactor' property being passed to an svg element, causing console errors and two unittests to fail for Artemis.


loreanvictor (approved):
The code works as intended, all functionality apparently working as before with tests passing. I removed a no-console suppression from one file so that we don't have accidental `console.log()`s sneak into prod.

PR #314, opened 2023-10-24
Add initial pool and swimlane implementation

matthiaslehnertum:
This PR intends to add support for pools and swimlanes to the BPMN diagram type.

### Checklist

### Motivation and Context
Pools are swimlanes are central components of the BPMN standard, allowing to model activities of different operators and their exchanges.

### Description
Pools are added to the BPMN diagram type as a new entity type with the ability to add swimlane within it.

### Steps for Testing
1. Select the BPMN diagram type
2. Drop a pool element onto the canvas
3. Double click on the pool element
4. Add a few swimlanes
5. Try resizing the swimlanes
6. Add elements within the swimlanes and connect them
7. Delete one of the swimlanes

### Test Coverage

| File | % Stmts | % Branch | % Funcs | % Lines |
|------|-------:|-----:|-------:|-----:|
 main/packages/bpmn/bpmn-pool                                           |   49.58 |      100 |      40 |   49.58                                                        
  bpmn-pool-component.tsx                                               |     100 |      100 |     100 |     100                                                 
  bpmn-pool-update.tsx                                                  |    38.4 |      100 |       0 |    38.4                               
  bpmn-pool.ts                                                          |   40.32 |      100 |      50 |   40.32
  main/packages/bpmn/bpmn-swimlane                                       |   89.65 |      100 |   66.66 |   89.65
  bpmn-swimlane-component.tsx                                           |     100 |      100 |     100 |     100       
  bpmn-swimlane.ts                                                      |   81.81 |      100 |      50 |   81.81
### Screenshots
Example of a BPMN diagram with a pool with multiple swimlanes
Screenshot 2023-10-24 at 16 42 42

Example of the update dialog of the pool that allows inserting additional swimlanes
Screenshot 2023-10-24 at 17 20 36


src/main/packages/bpmn/bpmn-pool/bpmn-pool.ts
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  type: UMLElementType = BPMNElementType.BPMNPool;
+
+  render(layer: ILayer, children: ILayoutable[] = []): ILayoutable[] {
+    if (this.bounds.width < BPMNPool.MIN_WIDTH) {
+      this.bounds.width = BPMNPool.MIN_WIDTH;
+    }
+
+    console.log(layer);
```
loreanvictor, line 29: @matthiaslehnertum I assume this should be removed, right? also the linter does not check for console.logs after the migration towards eslint, I'd recommend adding the following linter rule so that subsequent logs do not escape into production code:

```js
'no-console': ['error', { allow: ['warn', 'error'] }]
```

With this, CI checks on GitHub will automatically fail in case of escaped console.logs (or by running `npm run lint`).

src/tests/unit/packages/bpmn/bpmn-pool/bpmn-transaction-component-test.tsx
```diff
@@ -0,0 +1,22 @@
+import { wrappedRender } from '../../../test-utils/render';
+import * as React from 'react';
+import { CSSProperties } from 'react';
+import { Multiline } from '../../../../../main/utils/svg/multiline';
+import { BPMNPool } from '../../../../../main/packages/bpmn/bpmn-pool/bpmn-pool';
+import { BPMNPoolComponent } from '../../../../../main/packages/bpmn/bpmn-pool/bpmn-pool-component';
+
+// override getStringWidth, because it uses by jsdom unsupported SVGElement methods
```
loreanvictor, line 8: @matthiaslehnertum I don't need this locally to get the tests passing, perhaps we need to pin a later version of JSDOM to resolve this issue? Additionally, this method of patching could potentially break subsequent tests as it is not reversed, I suspect in these cases polyfilling DOM APIs will yield a more robust and explicit work-around.

src/tests/unit/packages/bpmn/bpmn-swimlane/bpmn-swimlane-component-test.tsx
```diff
@@ -0,0 +1,22 @@
+import { wrappedRender } from '../../../test-utils/render';
+import * as React from 'react';
+import { CSSProperties } from 'react';
+import { Multiline } from '../../../../../main/utils/svg/multiline';
+import { BPMNSwimlane } from '../../../../../main/packages/bpmn/bpmn-swimlane/bpmn-swimlane';
+import { BPMNSwimlaneComponent } from '../../../../../main/packages/bpmn/bpmn-swimlane/bpmn-swimlane-component';
+
+// override getStringWidth, because it uses by jsdom unsupported SVGElement methods
+Multiline.prototype.getStringWidth = (str: string, style?: CSSProperties) => {
```
loreanvictor, line 9: same as the other test, I don't need this patch locally to get the tests passing.

loreanvictor (changes requested):
One final note: how can I disassociate a BPMN element from a pool? I can drag an element into a pool and they are associated (e.g. they move together), but dragging them outside of the pool won't disassociate the elements. This is in contrast to behaviour of similar elements in other diagram types (e.g. packages in class diagrams) where I can drag a child element out of a parent element to disassociate them.

matthiaslehnertum:
@loreanvictor I would merge the Groups PR first - that one fixes the disassociation issue for the pools and swimlanes as well

loreanvictor (changes requested):
a few issues I've found on:

- the `zoomFactor` property error appears again, perhaps it was fixed on some other elements? it happens when adding a pool.

  Screenshot 2023-11-02 at 12 46 47
  
- management of lanes in a pool differs from management of other "child elements": to add a lane you need to double click on the pool and add them via the context menu, but to manage lanes you need to select lanes individually.

  Screenshot 2023-11-02 at 12 47 40

  this is in contrast to other elements with children, for example attributes on a class, where the context menu allows adding AND managing children (which seems more intuitive).
  
  Screenshot 2023-11-02 at 12 48 07
  
  of course this second point is a UX issue not an actual issue, so it could also be further discussed and if it really needs to be changed / updated, that can happen via a subsequent PR.
  
p.s. in addition to these issues, one of the tests is still failing.

loreanvictor (approved):
Ok all the previous issues are nicely resolved and the tests are passing, manual testing also seems to be fine with the pools and lanes, etc.

PR #322, opened 2023-11-19
`Feature`: Inverted editor layout

matthiaslehnertum:
As per Stephan's request, this PR flips the layout of the Apollon Editor to have the component sidebar on the left side of the editor canvas.

### Checklist

### Motivation and Context
To make Apollon more similar to other editor applications, the component sidebar is moved to the left side.

### Steps for Testing
Open up Apollon, see if the sidebar is correctly rendered on the left side of the editor canvas and make sure paddings within the application seem correctly sized.

### Screenshots

Screenshot 2023-11-19 at 22 25 53


loreanvictor (approved):
works fine. I personally preferred the menu on the other side 😅 but I checked with various similar diagram editing software and the positioning of the menus is not consistent, so I suppose its a matter of preference. perhaps it would be nice if we could break the menu into two parts: keeping the general menu (diagram type, language, etc) on the left while having the elements menu on the right, which would result in a more balanced layout of the page instead of being skewed to the side (but I suspect thats up for discussion and then a follow-up PR).

PR #332, opened 2023-12-07
Add support for box selection

matthiaslehnertum:
### Checklist

### Motivation and Context
This PR provides the implementation of a selection box using the cursor in order to select multiple elements without the use of the keyboard.

### Description
Every time the user clicks OUTSIDE an element, the selection box is created. In order for it to become visible, the user needs to drag the cursor. An element needs to be contained FULLY within the selection box for it to be selected when the selection box is released (i.e. the click/drag action is over). This feature also works when pressing the SHIFT key without deselecting the already selected items.

If the user's cursor goes outside the scope of the canvas while the selection box is active, the selection box will "freeze", but it will continue when the cursor comes back to the canvas. If the user releases the click while the cursor is outside the canvas and they return to the canvas, the selection box will remain active; clicking again will "resume" the selection box and the user can continue to drag the cursor in order to change the already active selection box before releasing it.

Adjustments:
- hovering is disabled when the selection box is active
- the cursor shape normally changes to sizeall (pointer: 'move') when hovering over the inside of an element - disabled while selection box is active
- the cursor shape normally changes to sizenesw/sizenwse when hovering over the edges of an element - disabled while selection box is active

### Steps for Testing

1. Open Apollon in Chrome
2. Add a few UML Elements to the canvas
3. Click somewhere on the canvas where there are no elements and drag the mouse to see the selection box
4. Drag the mouse to include the elements that you want to select
5. Release the click and all the elements in the selection box should be selected while the others should remain deselected

### Screenshots

grafik

grafik


src/main/components/canvas/mouse-eventlistener.tsx
```diff
@@ -0,0 +1,273 @@
...
+        endX: undefined,
+        endY: undefined,
+      },
+    };
+  }
+
+  componentDidMount() {
+    const { layer } = this.props.canvas;
+    if (!this.props.readonly && this.props.mode !== ApollonMode.Assessment) {
+      layer.addEventListener('mousedown', this.mouseDown);
```
loreanvictor, line 83: I suspect these event listeners need to be cleaned up in  [`componentWillUnmount()`](https://legacy.reactjs.org/docs/react-component.html#componentwillunmount) hook.

src/main/components/canvas/mouse-eventlistener.tsx
```diff
@@ -0,0 +1,273 @@
...
+  ),
+);
+
+class MouseEventListenerComponent extends Component<Props, LocalState> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      selectionStarted: false,
+      selectionRectangle: {
+        startX: undefined,
```
loreanvictor, line 72: I would recommend using [`IBoundary`](https://github.com/ls1intum/Apollon/blob/1c2aa378ca8b129b6258b584684ec3b2a6422511/src/main/utils/geometry/boundary.ts#L3-L8) interface, as it would keep the code consistent and allow usage of geometry utilities.

src/main/components/canvas/mouse-eventlistener.tsx
```diff
@@ -0,0 +1,273 @@
...
+    });
+  };
+
+  /**
+   * Check whether a given IUMLElement is contained in the currently active selection rectangle.
+   * Elements are only considered selected if they are fully contained within the selection rectangle.
+   *
+   * @param element The element for which containment in the selection box is determined
+   */
+  private isElementInSelectionBox = (element: IUMLElement): boolean => {
```
loreanvictor, line 221: I suspect this method should be a utility function in geometry, i.e. perhaps [here](https://github.com/ls1intum/Apollon/blob/develop/src/main/utils/geometry/boundary.ts).

loreanvictor: Additionally, this right now only selects an element if it is fully inside the selection box, while I suspect the more user-friendly way would be to select an element if its boundary overlaps with the selection box enough.

src/main/components/canvas/mouse-eventlistener.tsx
```diff
@@ -0,0 +1,273 @@
...
+   */
+  private mouseUp = (): void => {
+    // if no selection has been started, we can skip determining which
+    // elements are contained in the selection box.
+    if (!this.state.selectionStarted) {
+      return;
+    }
+
+    const selection = this.getElementIDsInSelectionBox();
+    this.props.select(selection);
```
loreanvictor, line 176: why is the selection happening here? shouldn't it happen during mouse move? right now the user doesn't have a feedback of what they are successfully selecting before releasing the cursor, while that way they could immediately see if they are selecting an intended element or not, or whether they are mistakenly selecting unwanted elements. it would of course require calculating which elements should also be deselected.

loreanvictor (changes requested):
The change works as intended, though I feel it lacks test coverage? Besides that at least one of the changes (about unmounting) can potentially lead to memory leaks in future (listeners not being cleaned up). The rest are UX improvements that I think are important but also could be done iteratively in subsequent PRs.

PR #353, opened 2024-05-06
Set a Time Limit for Stuttering Suppression on Each Path in Realtime Collaboration

loreanvictor:
### Checklist
- ~~[x] I added multiple screenshots/screencasts of my UI changes~~ (NA)
- ~~[x] I translated all the newly inserted strings into German and English~~ (NA)

### Motivation and Context

During realtime collaboration, Apollon optimistically drops some incoming changes it predicts that will be overwritten by some future change, based on changes the client itself has issued on specific paths on the diagram. This suppression continues until the client receives a "confirmation" patch from the server (representing the change that the client predicted would overwrite suppressed changes).

In some cases, due to network loss, this confirmation change might not be received by a client. In these situations the client will indefinitely suppress all incoming changes on particular paths erroneously.

### Description

This change sets up a timed control for suppression of incoming changes on each path. After a certain time has passed from the last local change on some specific path, the client will stop suppressing incoming changes on said path.

This prevents lock-in due to network loss, but still allows realtime collaboration to operate smoothly.

### Steps for Testing

1. Clone [Apollon Standalone](https://github.com/ls1intum/Apollon_standalone)
2. [Link Apollon Standalone](https://github.com/ls1intum/Apollon_standalone?tab=readme-ov-file#link-local-project-of-apollon) with local Apollon build
3. Run Apollon Standalone
4. Create a diagram and share it in "Collaborate" mode. Open two browsers to that link and collaborate on the diagram.
5. You should not see a client stuck and not applying the changes from the other client properly. The two clients should remain convergent.

### Test Coverage
| File | Branch | Line |
|------|-------:|-----:|
| services/patcher/patcher.ts | 96.77% | 100% |
| services/patcher/patch-verifier.ts | 100% | 100% |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "react-redux": "8.1.3",
         "react-tooltip": "4.5.1",
         "redux": "4.2.1",
-        "redux-saga": "1.2.3",
+        "redux-saga": "1.3.0",
         "redux-thunk": "2.4.2",
         "rxjs": "7.8.1",
         "styled-components": "5.3.11",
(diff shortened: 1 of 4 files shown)


matthiaslehnertum (approved):
Looks good to me!

PR #355, opened 2024-05-09
Upgrade Redux to version 5.0.1

matthiaslehnertum:
### Motivation and Context

This PR upgrades the version of Redux used in Apollon to the latest version 5.0.1.

### Description
Updating Redux to the latest version unfortunately required quite a few changes to be made. In the current setup, types are converted to `any` in quite a few places which will require more changes in follow-up PRs to be resolved properly.


krusche (changes requested):
Casting types to `any` is not an ideal solution. Is there no other way?

src/main/packages/uml-communication-diagram/uml-communication-link/uml-communication-link-update.tsx
```diff
@@ -154,15 +154,15 @@ class CommunicationLinkUpdate extends Component<Props, State> {
 
   private rename = (value: ICommunicationLinkMessage) => (name: string) => {
     const { element, update } = this.props;
-    const messages: ICommunicationLinkMessage[] = [...element.messages];
+    const messages: ICommunicationLinkMessage[] = element.messages.map((message) => ({ ...message }));
```
loreanvictor, line 157: this is a change in behaviour. the previous syntax would have used the same elements, while this code also clones the items. @matthiaslehnertum is that intended?

docs/source/user/api/apollon-editor.d.ts
```diff
@@ -1,10 +1,9 @@
 import 'pepjs';
-import { DeepPartial } from 'redux';
 import { Styles } from './components/theme/styles';
 import { Patch } from './services/patcher';
 import { Locale } from './services/editor/editor-types';
 import * as Apollon from './typings';
-import { UMLDiagramType, UMLModel } from './typings';
+import { DeepPartial, UMLDiagramType, UMLModel } from './typings';
```
loreanvictor, line 6: this is a recurring change all over, may I ask why do we need to customise the `DeepPartial` type helper offered by redux? how was it changed?

matthiaslehnertum: The DeepPartial Type is no longer exported by Redux version 5 but all diagram element constructors are initialized with "DeepPartials" of the diagram element models - therefore, I introduced a self defined version of DeepPartial.

src/main/packages/uml-communication-diagram/uml-communication-link/uml-communication-link-update.tsx
matthiaslehnertum: With the instance being inserted into the store directly before, we had the issue that the editor itself would alter properties on the reference to the "serialized" element which triggers assertions within Redux as this is non-intended behavior. I therefore introduced the copy-logic to prevent such issues. 🤔

egenerse:
can you update these files?

>      1  src/main/packages/uml-component-diagram/uml-component-subsystem/uml-component-subsystem.ts:1
>      1  src/main/packages/uml-component-diagram/uml-component/uml-component-component.ts:3
>      1  src/main/packages/uml-deployment-diagram/uml-deployment-component/uml-component.ts:1

loreanvictor:
@matthiaslehnertum is this still ongoing? do you have any idea why the build is failing?

PR #625, opened 2026-04-19
feat: add opt-in quiz mode and interactive export support

FadyGergesRezk:
### Checklist

### Motivation and Context

This PR addresses the quiz-element export workflow needed by Artemis drag-and-drop quiz generation and makes quiz interactions explicitly opt-in.

### Description

This PR introduces an opt-in Quiz Mode flow in Apollon while keeping default behavior unchanged for existing consumers.

Main changes:
- Added an enableQuizMode option (default false) so quiz-related UI and interactions are only active when explicitly enabled.
- Added view handling for Modelling and Highlight states, used to switch between normal modelling and quiz-element selection.
- Updated sidebar behavior so quiz controls are hidden unless quiz mode is enabled.
- Added interactive element/relationship tracking in diagram state and model serialization (interactive field in model).
- Added utility functions to toggle and prune interactive selections.
- Integrated export filtering for include/exclude use cases and switched export bounds calculation to rendered DOM-based bounds merging (nodes, edges, overflow, text).
- Added Playground control (Enable Quiz Mode checkbox) to test the feature without exposing quiz controls by default in normal usage.

Compatibility:
- Existing integrations remain unaffected unless enableQuizMode is set to true.
- Default editor behavior remains normal modelling mode with no quiz-specific controls.

### Steps for Testing

1. Build and test:
1. Run npm run build:lib
2. Run npm run build:webapp
3. Run npm run test --workspace=@tumaet/apollon

2. Playground behavior:
1. Run npm run dev and open /playground
2. Verify Enable Quiz Mode is unchecked by default
3. Verify sidebar renders normal modelling tools only
4. Enable Quiz Mode and verify Model / Quiz Elements controls appear
5. Switch to Quiz Elements and click nodes/edges to mark them
6. Switch back to Model and verify normal modelling tools are shown again

3. Serialization and export behavior:
1. Select quiz elements in Quiz Elements view
2. Export JSON and verify interactive data is present for selected elements/relationships
3. Verify export still works when no interactive selections are made
4. Verify include/exclude-based export uses filtered rendered bounds

### DIFF
--- a/library/lib/components/Sidebar.tsx
+++ b/library/lib/components/Sidebar.tsx
@@ -10,6 +10,7 @@ import { DividerLine } from "./ui/DividerLine"
 import { useMetadataStore } from "@/store/context"
 import { useShallow } from "zustand/shallow"
 import { DraggableGhost } from "./DraggableGhost"
+import { ApollonView } from "@/typings"
 
 /* ========================================================================
    Sidebar Component
@@ -17,7 +18,14 @@ import { DraggableGhost } from "./DraggableGhost"
    ======================================================================== */
 
 export const Sidebar = () => {
-  const diagramType = useMetadataStore(useShallow((state) => state.diagramType))
+  const { diagramType, view, setView, enableQuizMode } = useMetadataStore(
+    useShallow((state) => ({
+      diagramType: state.diagramType,
+      view: state.view,
+      setView: state.setView,
+      enableQuizMode: state.enableQuizMode,
+    }))
+  )
   const labelPreviewTypes = new Set([
     "sfcTransitionBranch",
     "petriNetPlace",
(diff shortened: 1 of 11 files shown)


library/lib/components/Sidebar.tsx
```diff
@@ -44,62 +52,136 @@ export const Sidebar = () => {
...
-                  data: config.defaultData,
-                  SIDEBAR_PREVIEW_SCALE: previewScale,
-                  id: `sidebarElement_${index}`,
-                })}
-              </div>
-            </DraggableGhost>
-          </React.Fragment>
-        )
-      })}
+      {enableQuizMode && (
```
FelixTJDietrich: Do we really need to call it quiz mode? Don't we already have a selecting mode for the assessment? I would have guessed that the only thing we are missing is just the cutout API using the selection

library/lib/typings.ts
```diff
@@ -87,6 +94,7 @@ export type ApollonOptions = {
   debug?: boolean
   collaborationEnabled?: boolean
   scrollLock?: boolean
+  enableQuizMode?: boolean
```
FelixTJDietrich: Why is this a separate mode? I think we already have some in ApollonMode and I would like to avoid calling it quiz mode in Apollon. Generally I would like to keep the features as Artemis agnostic as possible so others can build their use-cases around it.

FelixTJDietrich (dismissed):

tamang29 (approved):

PR #630, opened 2026-04-21
ci: create release tags automatically

FadyGergesRezk:
### Checklist

### Motivation and Context

The release workflows currently call `gh release create ... --verify-tag` for both the library and standalone releases, but they do not create the required git tags beforehand.

That means a release can partially succeed and then fail at GitHub Release creation when the expected tag is missing:
- library expects `@tumaet/apollon@<version>`
- standalone expects `v<version>`

This PR fixes that release-flow gap by making CI create the required tag automatically before running the verified GitHub release step.

### Description

This PR updates the release workflows so they no longer depend on a manually created tag.

Changes:
- `release-library.yml`
  - detects whether the GitHub release still needs to be created
  - creates `@tumaet/apollon@<version>` if it does not already exist
  - keeps `--verify-tag` in place for `gh release create`
  - handles the rerun case where npm publish already succeeded but the GitHub release/tag is still missing
- `release-standalone.yml`
  - creates `v<version>` if it does not already exist
  - keeps `--verify-tag` in place for `gh release create`

### Steps for Testing

1. Review `.github/workflows/release-library.yml` and confirm the workflow now ensures `@tumaet/apollon@<version>` exists before `gh release create --verify-tag`.
2. Review `.github/workflows/release-standalone.yml` and confirm the workflow now ensures `v<version>` exists before `gh release create --verify-tag`.
3. Confirm both workflows are idempotent when the tag already exists on `origin`.
4. Merge this PR before the next release bump PR and verify the next release no longer requires manual tag creation.

### DIFF
--- a/.github/workflows/release-library.yml
+++ b/.github/workflows/release-library.yml
@@ -22,12 +22,15 @@ jobs:
     outputs:
       version: ${{ steps.v.outputs.version }}
       publish: ${{ steps.v.outputs.publish }}
+      release: ${{ steps.v.outputs.release }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           ref: ${{ github.sha }}
 
       - id: v
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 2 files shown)


FelixTJDietrich (approved):
Looks good, lets see if it works :D

PR #643, opened 2026-04-25
feat: improve apollon quiz mode

FadyGergesRezk:
### Checklist

### Motivation and Context

This PR improves Apollon’s generic interactive element selection flow used by downstream integrations such as Artemis drag-and-drop quiz generation.

Previously, the Highlight/interactive selection flow only worked reliably for top-level nodes and relationships. Nested class diagram elements, such as attributes and methods, could not be selected for cutouts, stale selections could remain after changes, and selected relationships were visually hard to distinguish.

### Description

This PR extends the existing generic Highlight view instead of introducing an Artemis-specific quiz mode.

The changes include:
- Support selecting nested node elements, including class attributes and methods.
- Persist nested selections in `model.interactive.elements`.
- Prune stale interactive IDs when nodes, nested elements, or relationships are removed.
- Preserve explicit empty interactive selections so consumers can distinguish “nothing selected” from legacy diagrams without interactive data.
- Improve selected-element styling with a distinct interactive selection color.
- Keep the existing modeling mode as the default unless the Highlight view is explicitly enabled by the embedding application.

### Steps for Testing

1. Start the Apollon webapp.
2. Create a class diagram with at least one class containing attributes and methods.
3. Enable the Highlight/interactive selection view.
4. Select a class, an attribute, a method, and a relationship.
5. Verify selected elements are visibly highlighted with the interactive selection color.
6. Export or inspect the editor model and verify selected node/attribute/method IDs appear in `interactive.elements`, while selected relationships appear in `interactive.relationships`.
7. Deselect all elements and verify `interactive` is still serialized as empty records instead of falling back to stale selections.
8. Delete selected nodes or relationships and verify deleted IDs are removed from `interactive`.
9. Return to normal modeling view and verify regular modeling behavior is unchanged.

### Screenshots
N/A


library/lib/components/wrapper/AssessmentSelectableWrapper.tsx
```diff
@@ -4,6 +4,11 @@ import { useDiagramStore, useMetadataStore } from "@/store"
 import { ApollonMode, ApollonView } from "@/typings"
 import { useShallow } from "zustand/shallow"
 
+const INTERACTIVE_SELECTION_COLOR =
+  "var(--apollon-interactive-selection, #f39c12)"
+const INTERACTIVE_SELECTION_FILL =
+  "color-mix(in srgb, var(--apollon-interactive-selection, #f39c12) 18%, transparent)"
+
```
FelixTJDietrich: Should those consts be defined here?

library/lib/components/AssessmentSelectableElement.tsx
```diff
@@ -1,5 +1,11 @@
 import { useAssessmentSelection } from "@/hooks"
+import { useDiagramStore, useMetadataStore } from "@/store"
+import { ApollonMode, ApollonView } from "@/typings"
 import { FC } from "react"
+import { useShallow } from "zustand/shallow"
+
+const INTERACTIVE_SELECTION_COLOR =
```
FelixTJDietrich: Same here

FelixTJDietrich (dismissed):
Functionality works as expected
Image

Not sure if we want the constants here. Can you integrate with the existing constants file?

tamang29 (approved):

Issue #231, opened 2022-05-10
UML Diagram: Communication Diagram


Johnypier:
### Describe the bug
In this case, self-messages point to the left and they overlap with current class diagram, which makes them less visible.

<img src="https://user-images.githubusercontent.com/98538725/167612301-dbfa7ccd-5dbe-41e0-8576-9cd8052c70b7.png">

#### To Reproduce

1. Go to exercise with Communication Diagram edditor.
2. Create self-messages.
3. See error

#### Expected behavior

Suggestion: let the messages be displayed above with arrow pointing to opposite direction.

<img src="https://user-images.githubusercontent.com/98538725/167612378-bf7a2283-2cc7-451b-8955-61351a3ec1b6.png">

### Environment

<details><pre>
 - OS: Windows 10
 - Browser: Microsoft Edge
 - Version: 101.0.1210.39
</pre></details>

PR #645, opened 2026-04-26
fix: prevent communication diagram message labels from overlapping the edge

### LINKED ISSUE #634: [BUG] Edge labels overlapping in Communication Diagram
### Describe the bug

 Edge labels overlapping in Communication Diagram. 
Vertical and horizontal edge labels are overlapping over edge lines.

### To Reproduce

1. Create a Communication Diagram
2. Drag and Drop elements
3. Create connection between these elements
4. Add an edge label

### Expected behavior

working as expected

### Screenshots

Image

Image

### OS

_No response_

### Browser

_No response_

### Version

_No response_

### Additional context

_No response_

### Web Console Output

```shell

```

tamang29:
### Summary

Communication-diagram edges carry messages in both directions, each drawn as a small arrow + text label beside the edge's middle segment. Two things were wrong:

1. The labels **overlapped the edge line**, and
2. on **vertical** edges the forward and backward messages were drawn on the **wrong sides** (the swap @Claudia-Anthropica and @FelixTJDietrich flagged in review).

The root cause was the positioning model. Each label was horizontally **centred** on the edge and then pushed clear with a fixed offset whose size came from an **estimate** of the text width (`text.length * 8`). Long labels still crossed the line, short ones drifted too far, and a `-90` magic offset had silently inverted the forward/backward sides.

This replaces that with an **anchor-to-edge, constant-gap** model — the same approach the neighbouring `EdgeMiddleLabels` already uses:

- **Horizontal edges** — labels are centred above (forward) / below (backward) the line, with the direction arrow stacked between the text and the line.
- **Vertical edges** — labels are anchored to their inner edge and grow *outward* via the SVG `text-anchor` (forward right, backward left), so the gap to the line stays constant no matter how long the text is.

Message **direction** is now encoded solely by the arrow's rotation, never by which side the label sits on — so the side is stable regardless of which node is the source.

Fixes #634.

### Release note

Communication diagram message labels no longer overlap the edge line, and on vertical edges the two message directions now appear on the correct sides.

### Implementation notes

- The label geometry is extracted into a small **pure function**, `computeMessageLayout`, that returns every position relative to the segment midpoint — so it is deterministic and unit-tested in isolation (no DOM, no React).
- The old `useMessagePositioning` hook is **removed**. After the extraction its only remaining job was a `setTimeout`-based "wait for the midpoint" gate, which had become a no-op (the midpoint is seeded synchronously) and made these edges behave differently from every other edge type. Removing it also retired a stale `react-hooks/set-state-in-effect` lint exception.
- No text-width estimation anywhere; the duplicated arrow `<path>`s now share their stroke via SVG inheritance; dead code (unused layout outputs and an orphaned `labelUtils`) is deleted. Net diff is slightly negative.
- The labels stay **inline SVG** (rather than React Flow's `EdgeLabelRenderer` HTML labels) on purpose: the PNG/PDF export pipeline rasterises the edge SVG, and HTML labels rendered in a portal would be missing from exports.
- **Out of scope / follow-ups:** while testing I traced a separate, pre-existing bug where a vertical edge's measured midpoint in `useStepPathEdge` can be stale (that hook is shared by ~10 edge types, so it needs its own PR). Centring the labels means they render cleanly even when that midpoint is slightly off. Self-messages (#231) are a self-loop *routing* concern rather than a labelling one and are likewise left for a dedicated change.

### Steps for testing

1. Create a Communication Diagram and add two objects with a connecting edge.
2. Open the edge and add messages in **both** directions.
3. **Horizontal** edge: the forward message sits above the line and the backward below, each centred, with the arrow pointing along the message direction — neither touches the line.
4. **Vertical** edge: the forward message sits to the right, the backward to the left, both clear of the line.
5. Give a message a long text and confirm the gap to the line stays constant (no crossing).
6. Export the diagram to PNG/PDF and confirm the labels still render.

`pnpm --filter @tumaet/apollon test` covers the layout (grouping, sides per orientation, arrow rotations, stacking, swap-invariance).

### Screenshots / screencasts

**Before** — labels overlapping the edge line:

Before: communication diagram labels overlapping the edge

**After** — forward above/right, backward below/left, centred and clear of the line, on both horizontal and vertical edges:

_Updated capture to be attached. The earlier "after" screenshot in this PR's history predated the side/centring correction in this revision, so it is intentionally omitted here rather than shown as final._

### Checklist

- [x] Linked to a related issue (if applicable) — #634
- [x] Tests added or updated
- [ ] Documentation updated (if applicable) — no docs affected
- [x] Screenshots or screencasts attached (if a UI change) — "before" attached; refreshed "after" to follow


FelixTJDietrich:
Is this correct? The order of the messages on the edge is different 🤔 
image

library/lib/hooks/useMessagePositioning.ts
```diff
@@ -80,19 +81,20 @@ export const useMessagePositioning = (
...
-      forwardLabelBoxPosition = { x: baseOffset + labelOffset, y: 0 }
+      // Vertical edge: forward label on the RIGHT side
+      forwardArrowBoxPosition = { x: BASE_OFFSET_X, y: 0 }
+      forwardLabelBoxPosition = { x: BASE_OFFSET_X + LABEL_OFFSET, y: 0 }
 
-      backwardArrowBoxPosition = { x: -baseOffset, y: 0 }
-      backwardLabelBoxPosition = { x: -baseOffset - labelOffset, y: 0 }
+      // Vertical edge: backward label on the LEFT side
+      backwardArrowBoxPosition = { x: -BASE_OFFSET_X, y: 0 }
+      backwardLabelBoxPosition = { x: -(BASE_OFFSET_X + LABEL_OFFSET), y: 0 }
```
Claudia-Anthropica: @tamang29 [high] This block swaps the sides forward and backward labels appear on compared to the old code, which is what @FelixTJDietrich is flagging in the PR comment. Previously `baseOffset = 15` (positive) put `forward` at `x = +27` (right) and `backward` at `x = -27` (left). Now with `BASE_OFFSET_X = -90`, `forward` ends up at `x = -72` (left) and `backward` at `x = +72` (right) — the opposite of what the new code comments claim ("forward label on the RIGHT side" / "backward label on the LEFT side"). Easiest fix: make `BASE_OFFSET_X` a positive distance (e.g. `90`) so the existing sign structure (`BASE_OFFSET_X` for forward, `-BASE_OFFSET_X` for backward) lines up with the comments and the original ordering.

```suggestion
    const BASE_OFFSET_X = 90
    const BASE_OFFSET_Y = 20
    const LABEL_OFFSET = 18

    // Position labels and arrows based on actual middle segment orientation
    let forwardArrowBoxPosition: { x: number; y: number }
    let forwardLabelBoxPosition: { x: number; y: number }
    let backwardArrowBoxPosition: { x: number; y: number }
    let backwardLabelBoxPosition: { x: number; y: number }

    if (isHorizontalEdge) {
      // Horizontal edge: labels above and below, arrow + label go left→right
      forwardArrowBoxPosition = { x: 0, y: -BASE_OFFSET_Y }
      forwardLabelBoxPosition = { x: LABEL_OFFSET, y: -BASE_OFFSET_Y }

      backwardArrowBoxPosition = { x: 0, y: BASE_OFFSET_Y }
      backwardLabelBoxPosition = { x: LABEL_OFFSET, y: BASE_OFFSET_Y }
    } else {
      // Vertical edge: forward label on the RIGHT side
      forwardArrowBoxPosition = { x: BASE_OFFSET_X, y: 0 }
      forwardLabelBoxPosition = { x: BASE_OFFSET_X + LABEL_OFFSET, y: 0 }

      // Vertical edge: backward label on the LEFT side
      backwardArrowBoxPosition = { x: -BASE_OFFSET_X, y: 0 }
      backwardLabelBoxPosition = { x: -(BASE_OFFSET_X + LABEL_OFFSET), y: 0 }
    }
```

In `library/lib/hooks/useMessagePositioning.ts` around lines 73-97, the vertical-edge branch swaps the sides of forward/backward labels relative to the prior behavior because `BASE_OFFSET_X` is negative. Change `const BASE_OFFSET_X = -90` to `const BASE_OFFSET_X = 90` so that `forwardArrowBoxPosition.x = BASE_OFFSET_X` puts the forward label on the right side and `backwardArrowBoxPosition.x = -BASE_OFFSET_X` puts the backward label on the left, matching both the in-code comments and the previous semantics. Do not change any other expressions.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The horizontal-edge tweak looks fine, but the vertical-edge branch swaps which side the forward/backward labels appear on compared to the previous behavior — same thing @FelixTJDietrich was asking about. The code comments say forward goes on the right and backward on the left, but with `BASE_OFFSET_X = -90` the math actually puts them on the opposite sides. See the inline comment for the fix.

FelixTJDietrich (approved):

PR #662, opened 2026-05-03
feat: create new home page for apollon

### LINKED ISSUE #585: Home Screen with Diagram Gallery and Type Selection Grid
## Problem

Currently, opening Apollon at `/` immediately loads the last-edited diagram or auto-creates a blank Class Diagram. There is no landing page, no diagram overview, and no intuitive way to select a diagram type.

Students reported three major pain points:
1. **Finding diagram type selection is hard:** It's buried in `File → New File → modal`. The modal presents types as a flat button list — easy to miss and hard to scan.
2. **Managing multiple diagrams is difficult:** The only way to see existing diagrams is `File → Load Diagram`, which opens a plain text list (title, type, date) with no visual preview. Students can't visually identify which diagram is which, especially with default names.
3. **No way to quickly navigate between diagrams:** Without a central overview, students must go through the menu system every time they want to switch diagrams.

Academic UX research on UML tools confirms that initialization overhead and navigation confusion are the #1 reported usability obstacles for students (see: [ACM systematic review of UML tools for higher education](https://dl.acm.org/doi/fullHtml/10.1145/3610969.3611123)).

## Proposed Solution

Add a home page at `/` that serves as both a diagram gallery and a creation starting point, similar to how Google Docs, Figma, and draw.io handle document management.

### Home Page Layout

**For first-time users (no existing diagrams):**
- Welcome message: "Welcome to Apollon — Create UML diagrams for your courses"
- Diagram type selection grid (see below)

**For returning users:**
- **"+ New Diagram" section** at the top with the type selection grid (collapsible)
- **Diagram gallery** below showing all existing diagrams as preview cards

### Diagram Type Selection Grid

A responsive tile grid where each card contains:
- A representative icon or simplified SVG illustration showing what the diagram type looks like
- The diagram type name (e.g., "Class Diagram")
- A one-line description (e.g., "Model classes, attributes, and relationships")

**Grouping by frequency of use:**
- **Most Used:** Class Diagram, Activity Diagram, Use Case Diagram
- **Structural:** Object, Component, Deployment, Flowchart, Syntax Tree, SFC
- **Behavioral:** Communication, Petri Net, Reachability Graph, BPMN

Clicking a tile creates a new diagram and navigates to the editor (`/local/{id}`, see #583).

### Diagram Gallery (Preview Cards)

Each diagram is shown as a card containing:
- **SVG thumbnail** of the actual diagram content (leverage the existing SVG export capability in the library)
- **Diagram title**
- **Diagram type** as a colored badge/chip
- **Relative last-modified time** ("2 hours ago", "Yesterday")
- **Quick actions on hover:** Open, Duplicate, Delete (with confirmation)

Cards are arranged in a responsive grid, sorted by last modified (most recent first).

**Search & filter:**
- Search bar to filter by title
- Diagram type filter chips

**Empty state:** Friendly illustration with "Create your first diagram" CTA pointing to the type selection grid above.

### Duplicate Diagram

Currently there is no way to duplicate a diagram — students must export JSON and re-import it (6+ clicks, undiscoverable). Add:
- "Duplicate" action on each gallery card (hover menu)
- Creates a copy with the title "{original title} (Copy)" and a new UUID
- Add `duplicateModel(id)` method to `usePersistenceModelStore`

### "+ New Diagram" Navbar Button

A prominent button always visible in the navbar:
- On the home page: scrolls to the type selection grid
- In the editor: opens a compact popover with the diagram type tiles
- Smart default naming: "Class Diagram 2" if "Class Diagram" already exists

### Thumbnail Generation & Caching

- Generate small SVG thumbnails when a diagram is saved (reuse the library's SVG export)
- Cache thumbnails in the persistence store alongside diagram data
- Thumbnails should be small/cropped to minimize storage impact
- Regenerate on diagram update (debounced)

### Files to create/modify

**New:**
- `standalone/webapp/src/pages/HomePage.tsx` — home page component
- `standalone/webapp/src/components/home/DiagramTypeGrid.tsx` — type selection tiles
- `standalone/webapp/src/components/home/DiagramGallery.tsx` — gallery grid
- `standalone/webapp/src/components/home/DiagramCard.tsx` — individual preview card
- `standalone/webapp/src/components/navbar/NewDiagramButton.tsx` — navbar button + popover
- Diagram type icons/illustrations (SVG assets per type)

**Modified:**
- `standalone/webapp/src/App.tsx` — add `/` → `HomePage` route
- `standalone/webapp/src/stores/usePersistenceModelStore.tsx` — add `duplicateModel()`, thumbnail storage
- `standalone/webapp/src/components/navbar/DesktopNavbar.tsx` — add "+ New Diagram" button
- `standalone/webapp/src/components/navbar/MobileNavbar.tsx` — add "+ New Diagram" button
- `standalone/webapp/src/components/modals/LoadDiagramModal.tsx` — may be replaced by the gallery or kept as a fallback

**Depends on:** #583 (URL-Based Routing) — the home page needs `/local/:id` routes to link to individual diagrams.

## Acceptance Criteria

- [ ] `/` shows the home page (not the editor directly)
- [ ] First-time users see a welcoming type selection grid with icons, names, and descriptions
- [ ] Returning users see their diagram gallery with SVG thumbnail preview cards
- [ ] Each card shows title, type badge, and relative last-modified time
- [ ] Clicking a type tile creates a new diagram and opens the editor
- [ ] Clicking a gallery card opens that diagram in the editor
- [ ] Hover on gallery cards reveals quick actions (open, duplicate, delete)
- [ ] Duplicate creates a copy with "(Copy)" suffix
- [ ] Delete requires confirmation
- [ ] Search bar filters diagrams by title
- [ ] Type filter chips narrow results by diagram type
- [ ] "+ New Diagram" button is always visible in the navbar
- [ ] Empty state guides users to create their first diagram
- [ ] Opening `/local/:id` directly goes to the editor (no forced home screen)
- [ ] Diagram type tiles are grouped with most common types first

FadyGergesRezk:
### Checklist

### Motivation and Context

This PR completes https://github.com/ls1intum/Apollon/issues/585

The webapp previously dropped users straight into a blank `ApollonLocal` editor on `/`, with no overview of saved work, no entry point for templates, and no way to deep-link a specific diagram. New users had nothing to anchor on; returning users had to dig through the "Load Diagram" modal to find anything. We needed (1) a real landing experience that surfaces saved and shared work, and (2) a routing model where every diagram has a stable, shareable URL.

This PR delivers those two things: **routing** and a **new home page**.

#### 1. Routing

Each diagram now lives at a stable URL, and `/` is a dedicated landing page instead of an editor mount.

- `/` → `HomePage`
- `/local/:id` → `ApollonLocal` (the local editor; previously mounted directly at `/`)
- `/shared/:id?view=<mode>` → `ApollonWithConnection` (shared/collaborative editor; was `/:diagramId`)
- `/playground`, `/imprint`, `/privacy` → unchanged pages
- `/:id?view=…` → `LegacySharedDiagramRedirect`: old single-segment shared links are 301-style redirected to `/shared/:id` (preserving the `?view=` query); anything without a `view` param falls through to `ErrorPage`.
- `*` → `ErrorPage`

Supporting pieces:
- The editor `Navbar` is only mounted off the home route; `HomePage` renders its own `HomeNavbar`.
- All page chunks are `lazy`-loaded, and the `Navbar`, `ToastContainer` (`DeferredToastContainer`), and modal wrapper are deferred for faster first paint behind a single `AppLoadingScreen` fallback.
- Shared-diagram URL handling is centralised in `utils/sharedDiagramLinks.ts` (`buildSharedDiagramPath` / `buildSharedDiagramUrl`, view normalization, the four share modes: Edit / Collaborate / Give feedback / Review feedback).
- `ApollonWithConnection` reads `:id` from the path, and on a failed initial load now shows a dedicated error screen (404 / server-unavailable / generic) with **Retry** and **Go home** instead of a toast + forced redirect.

#### 2. New home page — "Your diagrams" dashboard

A single, unified `DiagramGallery` dashboard replaces the blank editor landing.

- **Gallery** (`DiagramGallery`): a searchable, filterable list of diagrams with a "Your diagrams" heading and live count.
  - Source segmented control: **All / Local / Shared**.
  - Filters: search-by-name, diagram-type dropdown, sort (Alphabetical / Date created / Last viewed) + order, and a Favorites toggle.
  - **Grid and table** view modes via a segmented control.
  - Infinite scroll (`IntersectionObserver` sentinel, 9 per page) — no "Load more" button.
- **Cards** (`DiagramCard`): thumbnail preview, title, created/last-modified metadata, type + source/mode badges, a favorite star, and a three-dot actions menu (Open / Duplicate / Share / Delete-with-confirm; shared diagrams expose copy-link, change-mode, and remove).
- **Shared diagrams**: local browser storage of shared entries (`utils/sharedDiagramStorage.ts`), fetched and merged into the gallery; expired links render an "expired" card state. A new `ShareDashboardModal` creates a shareable link and lets the user switch the default view mode inline.
- **Thumbnails**: the editor generates a normalized SVG thumbnail after edits; the home page warms thumbnails for visible cards via `useDiagramThumbnailWarmup` (idle-scheduled, sequence-guarded, LRU-pruned theme cache, light/dark variants). Thumbnails are kept in zustand but excluded from `partialize` to avoid localStorage bloat.
- **Design system harmonization**: shared `HomeButton` and `HomeDialog*` primitives give every home button/dialog one source of truth for variants and hover/focus/disabled behaviour. New-diagram and template modals render a "home dialog" variant; a `ModalProgressContext` drives an in-modal progress bar. The post-restore Undo affordance and other notifications were unified onto `react-toastify` (the old MUI `SnackbarMessage` was removed).
- **Theming**: a full set of `--home-*` CSS variables (surfaces, text, borders, accent, tags, shadows) in `themings.json` + `webapp.css`, light and dark.

#### Tests

- New `tests/e2e/home-page.spec.ts` covering initial load, empty state, gallery search/filter/sort/favorites, source filter, grid↔table toggle, infinite scroll, and card actions (28 tests).
- New unit tests: `ApollonWithConnection.test.tsx` (loading/error/collab-prompt regression), `WebSocketManager.test.ts`, `sharedDiagramStorage.test.ts`.

#### Trade-offs

- Thumbnail warmup renders an offscreen `ApollonEditor` per visible diagram on home load. Mitigated with idle scheduling and a cache; fine for typical usage but may want IndexedDB persistence for users with very many diagrams.
- Old single-segment shared links (`/:id?view=`) keep working only via the redirect shim; links without a `view` param now resolve to the error page.

### Steps for Testing

1. Open the app at `/`. The "Your diagrams" dashboard should render (not the editor) with the New diagram / Template / Import header actions.
2. With an empty store (clear localStorage `persistenceModelStore`), the gallery shows the "No diagrams yet" empty state with New diagram / From template / Import buttons.
3. Click **New diagram** → pick a type and name → you land on `/local/<uuid>` in the editor.
4. Edit the diagram, then return home — the card thumbnail should reflect your edits.
5. Open a card's three-dot menu → **Duplicate** (new card appears) and → **Delete** → confirm (card disappears).
6. Use search, the type/sort dropdowns, the **All/Local/Shared** segmented control, and the **Favorites** toggle; the count and list should update.
7. Switch between **grid** and **table** views.
8. Add 10+ diagrams and scroll the gallery — more cards load automatically (no "Load more" button).
9. Share a local diagram (card menu → Share), copy the link, then open `/shared/<id>?view=…` in a new tab.
10. Visit an old-style link `/<id>?view=EDIT` — it should redirect to `/shared/<id>?view=EDIT`.
11. Open an unavailable shared link — the editor should show the load-error screen with Retry / Go home.
12. Toggle the theme switcher; home colours should swap via the `--home-*` tokens.
13. Run `npx playwright test tests/e2e/home-page.spec.ts` — all 28 tests should pass.


FelixTJDietrich (changes requested):
I guess you are still implementing the feedback we talked about in person. I noticed another issue:
Image

Apollon should always work even when the connection to the server has been lost.

standalone/webapp/src/App.tsx
```diff
@@ -1,63 +1,116 @@
...
+      )}
+      <div data-testid="editor-area" style={{ flex: 1, overflow: "hidden" }}>
+        <Suspense fallback={<PageFallback />}>
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="/local/:id" element={<ApollonLocal />} />
+            <Route path="/playground" element={<ApollonPlayground />} />
+            <Route path="/imprint" element={<ImprintPage />} />
+            <Route path="/privacy" element={<PrivacyPage />} />
+            <Route path="/shared/:id" element={<ApollonWithConnection />} />
```
Claudia-Anthropica: @FadyGergesRezk [critical] Shared diagrams used to live at `/:diagramId`; this PR moves them to `/shared/:id` and the old top-level route is removed entirely. Any link a user has shared or bookmarked (and any embed in another tool) now hits the `*` 404 route. Either keep the old path as an alias, or add a redirect route from `/:legacyId` → `/shared/:legacyId` so existing links keep working. The PR description doesn't call this out as a breaking change.

In `standalone/webapp/src/App.tsx`, the prior `/:diagramId` route that mapped onto `ApollonWithConnection` was removed, but no replacement preserves existing shared/bookmarked URLs. Add a route like `<Route path="/:legacyDiagramId" element={<RedirectToShared />} />` (where `RedirectToShared` reads `useParams().legacyDiagramId` and renders `<Navigate to={`/shared/${legacyDiagramId}${location.search}`} replace />`) ahead of the `*` catch-all so existing `/<uuid>?view=...` URLs still resolve.

</details>

standalone/webapp/src/App.tsx
```diff
@@ -1,63 +1,116 @@
...
+    <>
+      {!isHomeRoute && (
+        <Suspense fallback={<div style={{ height: 56 }} />}>
+          <Navbar />
+        </Suspense>
+      )}
+      <div data-testid="editor-area" style={{ flex: 1, overflow: "hidden" }}>
+        <Suspense fallback={<PageFallback />}>
+          <Routes>
+            <Route path="/" element={<HomePage />} />
```
Claudia-Anthropica: @FadyGergesRezk [high] Making `/` render `HomePage` instead of `ApollonLocal` broke `tests/e2e/version-history.spec.ts`: it does `await page.goto("/")` and then waits for `.react-flow` to be visible — the canvas never mounts on home, so the test (and its two retries) time out. Update the version-history regression test to navigate to a `/local/<id>` URL with a seeded model in localStorage, the same way `editor.spec.ts` was updated via `openTemporaryLocalDiagram(page)`.

In `standalone/webapp/tests/e2e/version-history.spec.ts`, the test still calls `await page.goto("/")` and waits for `.react-flow`, but `/` is now the home page (no canvas). Refactor the test to seed a model in `localStorage.persistenceModelStore` and navigate to `/local/<id>` before calling `waitForCanvasReady`, matching the `openTemporaryLocalDiagram` helper added in `editor.spec.ts`.

</details>

standalone/webapp/tests/e2e/home-page.spec.ts
```diff
@@ -0,0 +1,522 @@
...
+    page,
+  }) => {
+    await expect(page.getByRole("heading", { level: 1 })).toContainText(
+      "Model architecture clearly"
+    )
+    await expect(
+      page.getByRole("button", { name: "Create New Diagram" })
+    ).toBeVisible()
+    await expect(
+      page.getByRole("button", { name: "Browse Templates" })
```
Claudia-Anthropica: @FadyGergesRezk [critical] The hero in `HomePage.tsx` (empty state) renders "Create New Diagram" + "Start from Template", never a "Browse Templates" button — so this test and the "`'Browse Templates' scrolls to the template section`" test at L457 both time out. Either rename the button to "Browse Templates" in `HomePage.tsx` or change these assertions to match the rendered text.

In `standalone/webapp/tests/e2e/home-page.spec.ts` lines 65 and 460, replace `getByRole("button", { name: "Browse Templates" })` with `getByRole("button", { name: "Start from Template" })` so they match the actual hero CTA rendered by `standalone/webapp/src/pages/HomePage.tsx`.

</details>

standalone/webapp/tests/e2e/home-page.spec.ts
```diff
@@ -0,0 +1,522 @@
...
+    const diagrams = Array.from({ length: 10 }, (_, i) => ({
+      id: `d${i + 1}`,
+      title: `Diagram ${i + 1}`,
+    }))
+    await seedDiagrams(page, diagrams)
+  })
+
+  test("shows 9 cards initially and a 'Load more' button", async ({ page }) => {
+    await expect(page.locator('[role="listitem"]')).toHaveCount(9)
+    await expect(page.getByRole("button", { name: "Load more" })).toBeVisible()
```
Claudia-Anthropica: @FadyGergesRezk [critical] There is no "Load more" button in the gallery — `DiagramGallery.tsx` uses an `IntersectionObserver` on a sentinel `div` to auto-extend `visibleCount`. Both load-more tests will always fail. The PR description also says "pagination (9 + Load more)", so either the implementation should grow a real button (consistent with the docs and tests), or the tests should scroll the sentinel into view and assert that the card count grows.

In `standalone/webapp/src/components/home/DiagramGallery.tsx`, the visible-count grows via an `IntersectionObserver` on `sentinelRef`, but `tests/e2e/home-page.spec.ts` asserts a `getByRole("button", { name: "Load more" })`. Add a visible "Load more" button (rendered when `hasMoreDiagrams`) that calls the same `setVisibleCount((c) => Math.min(c + LOAD_MORE_STEP, deferredFilteredDiagrams.length))` handler. Keep the observer as a fallback or remove it.

</details>

standalone/webapp/tests/e2e/home-page.spec.ts
```diff
@@ -0,0 +1,522 @@
...
+    page,
+  }) => {
+    await page.getByRole("button", { name: "Open diagram actions" }).click()
+    // First "Delete" opens the confirm panel
+    await page.getByRole("button", { name: "Delete" }).click()
+    // The confirm "Delete" button is inside the confirm panel — scope to it
+    const confirmPanel = page.locator(".space-y-2.rounded-md.border")
+    await confirmPanel.getByRole("button", { name: "Delete" }).click()
+    await expect(page.locator('[role="listitem"]')).toHaveCount(0)
+    await expect(page.getByText("No diagrams yet")).toBeVisible()
```
Claudia-Anthropica, line 414: @FadyGergesRezk [high] `DiagramGallery` renders "No local diagrams yet" / "No shared diagrams yet" — there is no plain "No diagrams yet" string anywhere. Same mismatch for the "first card has 'Most Recent' badge" test at L176: `DiagramCard.tsx` never renders that badge. Either add a `Most Recent` badge / unify the empty-state copy, or fix the assertions. As-is, several gallery tests will keep failing on every run.

In `standalone/webapp/tests/e2e/home-page.spec.ts`, lines 136, 178, and 325 assert text/badge that `standalone/webapp/src/components/home/DiagramGallery.tsx` and `DiagramCard.tsx` do not render ("No diagrams yet", "Most Recent"). Either update the assertions to match the actual UI (e.g. `"No local diagrams yet"` and remove the Most Recent assertion) or add the missing UI to the components. Pick one direction and apply consistently.

</details>

standalone/webapp/src/stores/useThemeStore.tsx
```diff
@@ -65,13 +68,39 @@ export const useThemeStore = create<ThemeState>()(
...
-      name: "theme-storage",
+      name: "apollon-theme",
+      version: THEME_STORE_VERSION,
       storage: createJSONStorage(() => localStorage),
       partialize: (state) => ({
         systemThemePreference: state.systemThemePreference,
         userThemePreference: state.userThemePreference,
         currentTheme: state.currentTheme,
       }),
+      migrate: (persistedState, version) => {
```
Claudia-Anthropica: @FadyGergesRezk [medium] The persist `name` was renamed from `theme-storage` to `apollon-theme`, so for any existing user there is no value under `apollon-theme` and zustand's `migrate` is never called — the legacy lookup here is dead code. That's why the PR description notes "existing users will see a one-time reset to system theme." If the intent is actually to migrate, do the legacy import imperatively before `useThemeStore` is created (e.g. in `index.tsx`), copying the old key's `state` under the new key when the new key is absent. Otherwise drop the dead branch and keep the description note.

In `standalone/webapp/src/stores/useThemeStore.tsx`, the `migrate` callback for the new persist name `apollon-theme` will never run for legacy users (their data lives under `theme-storage`, and zustand only calls `migrate` when there's a payload at the new key). Before `create<ThemeState>()(persist(...))` runs, add a one-shot block that reads `localStorage.getItem("theme-storage")`, and if `apollon-theme` is absent, writes the legacy payload under the new key and removes the old one. Then delete the dead legacy-handling branch inside `migrate`.

</details>

standalone/webapp/src/services/DiagramAPIManager.ts
```diff
@@ -0,0 +1,74 @@
+import { UMLModel } from "@tumaet/apollon"
+import { serverURL } from "@/constants"
+
+export type StoredDiagram = UMLModel & {
+  id: string
+  createdAt: string
+  updatedAt: string
+}
+
+export class DiagramAPIManager {
```
Claudia-Anthropica: @FadyGergesRezk [medium] This is a second diagram API client living next to `DiagramApiClient` (which already has typed errors, content-type/abort handling, etc.). Of the four methods here, only `fetchStoredDiagram` is actually called — `fetchDiagramData`, `sendDiagramUpdate`, and `createDiagram` have zero callers. Move `fetchStoredDiagram` onto `DiagramApiClient` (where it can benefit from `ApiError` + `request<T>`) and delete this class. Two parallel HTTP layers will drift.

In `standalone/webapp/src/services/DiagramAPIManager.ts`, only `fetchStoredDiagram` is used (by `standalone/webapp/src/components/home/DiagramGallery.tsx`); `fetchDiagramData`, `sendDiagramUpdate`, and `createDiagram` are dead. Add a `fetchStoredDiagram` method on `DiagramApiClient` in `DiagramApiClient.ts` that uses the existing `request<T>` helper and returns `null` on 404, update the import in `DiagramGallery.tsx`, then delete `DiagramAPIManager.ts`.

</details>

standalone/webapp/src/components/home/DeferredHomeStats.tsx
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const StatsFallback = () => (
+  <>
+    <span className="inline-flex h-6 w-28 animate-pulse rounded-full border border-[var(--home-border-color)] bg-[var(--home-bg-secondary)]" />
+    <span className="inline-flex h-6 w-36 animate-pulse rounded-full border border-[var(--home-border-color)] bg-[var(--home-bg-secondary)]" />
+  </>
+)
+
+export const DeferredHomeStats = () => {
```
Claudia-Anthropica: @FadyGergesRezk [low] `DeferredHomeStats` (and the `HomeStats` it wraps) isn't imported anywhere outside this file. The PR description advertises "Hero stats (`HomeStats`) lazy-loaded via `requestIdleCallback`", but the hero in `HomePage.tsx` just renders three static count chips inline. Either wire `<DeferredHomeStats />` into the hero stats row or delete both files (and `useRecentDiagramFilters.ts`, which is also orphaned).

`standalone/webapp/src/components/home/DeferredHomeStats.tsx`, `HomeStats.tsx`, and `standalone/webapp/src/hooks/useRecentDiagramFilters.ts` are unused. Either render `<DeferredHomeStats />` inside the hero stats row of `standalone/webapp/src/pages/HomePage.tsx` (replacing the inline `Local: / Shared: / Total:` chips), or delete the three files.

</details>

standalone/webapp/src/stores/usePersistenceModelStore.tsx
```diff
@@ -114,18 +161,146 @@ export const usePersistenceModelStore = create<PersistenceModelStore>()(
...
+                  new Date().toISOString(),
+              },
+            }),
+            false,
+            "setThumbnail"
+          ),
+
+        getThumbnail: (id) => get().thumbnails[id] ?? null,
+
         getCurrentModel: () => {
```
Claudia-Anthropica: @FadyGergesRezk [low] `useExportAsJSON` was the only caller of `getCurrentModel()` and it was removed in this PR. The method (and its line in the `PersistenceModelStore` type) is now dead — drop it to keep the store surface minimal.

In `standalone/webapp/src/stores/usePersistenceModelStore.tsx`, remove the `getCurrentModel` implementation (lines around 304-308) and its declaration in the `PersistenceModelStore` type — it is no longer called anywhere after `useExportAsJSON` stopped using it.

</details>

standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -1,49 +1,91 @@
...
+        thumbnailExportTimeoutRef.current = null
+      }
+
       log.debug("Cleaning up Apollon instance")
       instance.destroy()
+      setCurrentModelId(null)
+      setEditor(undefined)
     }
-  }, [diagram?.id, state?.timeStapToCreate])
+  }, [diagram?.id, setCurrentModelId, setEditor, setThumbnail, updateModel])
```
Claudia-Anthropica, line 101: @FadyGergesRezk [nit] When navigating between two `/local/:id` routes the cleanup briefly calls `setCurrentModelId(null)` and `setEditor(undefined)` before the new effect sets them again. Nothing breaks here, but if any sibling subscribes to `currentModelId` you'll get a transient null state. Consider letting the new effect overwrite without an intermediate clear, or guard the clear with `if (currentModelId === diagram.id)`.

In `standalone/webapp/src/pages/ApollonLocal.tsx`, the `useEffect` cleanup calls `setCurrentModelId(null)` and `setEditor(undefined)` unconditionally, which causes a brief null state when navigating between two `/local/:id` URLs. Guard the clears so they only fire when the store still holds this effect's diagram id (e.g. `if (usePersistenceModelStore.getState().currentModelId === diagram.id) setCurrentModelId(null)`).

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk Big PR, and the home page looks great in the screencasts. A few things blocking though: the new `home-page.spec.ts` doesn't match the implementation (Browse Templates, Load more, Most Recent badge, "No diagrams yet" — none of these exist in the code) and `version-history.spec.ts` regressed because `/` no longer mounts the editor. End result: 11 e2e tests fail and CI is red. Also a silent breaking change: the old `/:diagramId` shared-link route is gone with no redirect, so any URL anyone has already shared/bookmarked now 404s. A few smaller items inline. Felix's earlier offline-resilience point also still needs addressing.

standalone/webapp/src/App.tsx
Claudia-Anthropica: @FadyGergesRezk Still seeing this: `App.tsx` still only has `/shared/:id` and the `*` route, so old `/<diagramId>` shared URLs still 404. This one is still blocking.

standalone/webapp/src/App.tsx
Claudia-Anthropica: @FadyGergesRezk Still not addressed: `version-history.spec.ts` still goes to `/` and waits for `.react-flow`, but `/` is now the home route. It needs to seed/navigate to `/local/<id>` or otherwise mount the editor.

standalone/webapp/tests/e2e/home-page.spec.ts
Claudia-Anthropica: @FadyGergesRezk Still mismatched: the test still looks for `Browse Templates`, while the current empty home page only renders `No diagrams yet.` and `New diagram`, and the diagram home view uses `From template`. Please align the test with the current intended UI or restore the CTA.

standalone/webapp/tests/e2e/home-page.spec.ts
Claudia-Anthropica: @FadyGergesRezk Still open: `DiagramGallery` still auto-loads via `IntersectionObserver` and does not render a `Load more` button, but the tests still require one. Either make the button exist or change the tests to exercise the infinite-load behavior.

standalone/webapp/tests/e2e/home-page.spec.ts
Claudia-Anthropica: @FadyGergesRezk Still not addressed: the test still expects `Most Recent` and the old empty-state flow, but `DiagramCard` does not render that badge and the empty `/` state no longer renders the gallery action. Update the assertions to match the current dashboard UI.

standalone/webapp/src/stores/useThemeStore.tsx
Claudia-Anthropica: @FadyGergesRezk This is still dead migration code: the store still persists under `apollon-theme`, so users with only `theme-storage` will not trigger `migrate` at all. Either keep the old key or explicitly read and migrate it before zustand hydration falls back to defaults.

standalone/webapp/src/services/DiagramAPIManager.ts
Claudia-Anthropica: @FadyGergesRezk Still present: `DiagramAPIManager` is still a separate fetch wrapper, and `DiagramGallery` still imports it for `fetchStoredDiagram`. Please fold that method into `DiagramApiClient` and delete the duplicate client.

standalone/webapp/src/components/home/DeferredHomeStats.tsx
Claudia-Anthropica: @FadyGergesRezk Partially fixed, but `useRecentDiagramFilters.ts` is still present and unused. Delete that leftover hook as well, then this thread can be resolved.

standalone/webapp/src/stores/usePersistenceModelStore.tsx
Claudia-Anthropica: @FadyGergesRezk Still present: `getCurrentModel` remains in the store type and implementation, and `rg` does not show any callers. Please remove the dead method from the store surface.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FadyGergesRezk Still seeing the unconditional clears in cleanup. `setCurrentModelId(null)` and `setEditor(undefined)` still run on every `/local/:id` transition, so the transient null state remains.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I rechecked the prior change requests, and the main blockers are still open: legacy shared URLs still 404, the version-history/home e2e tests still do not match the routed UI, and the theme migration still cannot run for users with the old key. I left follow-up replies on the existing threads. Codacy is also still action_required.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I rechecked the previous change requests against the latest commit. The new commit only restores library files, so the webapp blockers are still open: legacy shared URLs still 404, the e2e specs still do not match the routed UI, the theme migration still cannot read the old key, and Codacy is still failing on CSS style rules.

tamang29:
@FadyGergesRezk When user has no diagram, i get this UI where I can't create new diagram
Screenshot 2026-06-02 at 17 52 45

And if the url doesnt exist either, we redirect to home page or give user a suitable error message "Diagram no longer exist"?
Screenshot 2026-06-02 at 17 57 37

```pnpm run build``` was failing for me, can you check if its working for you.

Screenshot 2026-06-02 at 18 05 11
I think we should hide "Version History" for local mode until we merge Felix's PR (#678)

File: standalone/webapp/src/App.tsx (hunk 1 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/App.tsx]: @FadyGergesRezk Looks good now, the legacy `/<id>?view=...` route redirects to `/shared/<id>` and preserves the query.

File: standalone/webapp/src/App.tsx (hunk 2 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/App.tsx]: @FadyGergesRezk Still open on the latest commit: `version-history.spec.ts` still calls `page.goto("/")`, then waits for `.react-flow`, but `/` renders the dashboard now. Please seed a local diagram and navigate to `/local/<id>` before `waitForCanvasReady`.

File: standalone/webapp/tests/e2e/home-page.spec.ts (hunk 3 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/tests/e2e/home-page.spec.ts]: @FadyGergesRezk Looks good now, the home-page tests were rewritten around the current dashboard CTAs.

File: standalone/webapp/tests/e2e/home-page.spec.ts (hunk 4 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/tests/e2e/home-page.spec.ts]: @FadyGergesRezk Nice, the tests now exercise the infinite-scroll sentinel instead of expecting a `Load more` button.

File: standalone/webapp/tests/e2e/home-page.spec.ts (hunk 5 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/tests/e2e/home-page.spec.ts:414]: @FadyGergesRezk Looks good now, the assertions match the current empty state and no longer expect the missing badge.

File: standalone/webapp/src/stores/useThemeStore.tsx (hunk 6 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/stores/useThemeStore.tsx]: @FadyGergesRezk Still open: the legacy `theme-storage` read is still inside `migrate`, so it only runs after zustand finds data under `apollon-theme`. Users who only have the old key still fall back to defaults.

File: standalone/webapp/src/services/DiagramAPIManager.ts (hunk 7 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/services/DiagramAPIManager.ts]: @FadyGergesRezk Still open: `DiagramGallery` still imports `DiagramAPIManager`, and that duplicate wrapper still contains unused methods next to `DiagramApiClient`.

File: standalone/webapp/src/components/home/DeferredHomeStats.tsx (hunk 8 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/components/home/DeferredHomeStats.tsx]: @FadyGergesRezk Still just the leftover `useRecentDiagramFilters.ts` hook remains unused. Delete that file and this can be resolved.

File: standalone/webapp/src/stores/usePersistenceModelStore.tsx (hunk 9 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/stores/usePersistenceModelStore.tsx]: @FadyGergesRezk Still open: `getCurrentModel` remains in the store type and implementation, and there are still no callers.

File: standalone/webapp/src/pages/ApollonLocal.tsx (hunk 10 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/pages/ApollonLocal.tsx:101]: @FadyGergesRezk Still open: the cleanup still unconditionally calls `setCurrentModelId(null)` and `setEditor(undefined)` when leaving a local diagram.

**Review by Claudia-Anthropica — CHANGES_REQUESTED** [review]
@FadyGergesRezk The legacy shared-link route and the home-page e2e expectation mismatches are fixed now, but several prior blockers are still open: the version-history e2e still lands on `/`, the theme migration still cannot read legacy users, and `DiagramAPIManager` is still duplicated. Codacy is also still action_required.

File: standalone/webapp/src/App.tsx (hunk 2 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/App.tsx]: @FadyGergesRezk Looks good now, the version-history e2e seeds a local diagram and opens `/local/e2e-version-history-model`.

File: standalone/webapp/src/stores/useThemeStore.tsx (hunk 6 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/stores/useThemeStore.tsx]: @FadyGergesRezk Looks good now, the legacy `theme-storage` key is copied before zustand hydration.

File: standalone/webapp/src/services/DiagramAPIManager.ts (hunk 7 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/services/DiagramAPIManager.ts]: @FadyGergesRezk Nice, `fetchStoredDiagram` is on `DiagramApiClient` now and the duplicate manager is gone.

File: standalone/webapp/src/components/home/DeferredHomeStats.tsx (hunk 8 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/components/home/DeferredHomeStats.tsx]: @FadyGergesRezk Looks good now, the unused hook is gone.

File: standalone/webapp/src/stores/usePersistenceModelStore.tsx (hunk 9 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/stores/usePersistenceModelStore.tsx]: @FadyGergesRezk Looks good now, `getCurrentModel` is removed.

File: standalone/webapp/src/pages/ApollonLocal.tsx (hunk 10 above)
Comment (Claudia-Anthropica) [reply standalone/webapp/src/pages/ApollonLocal.tsx:101]: @FadyGergesRezk Still not quite fixed: on a `/local/old` -> `/local/new` transition, this cleanup runs before the new effect sets the new id, so `currentModelId` is still the old diagram id and the guard still clears to `null`. Keep the clear for leaving the local editor, but skip it when the next route is another local diagram, or let the new instance overwrite without clearing.

**Review by Claudia-Anthropica — CHANGES_REQUESTED** [review]
@FadyGergesRezk Most of the previous blockers are addressed now, and I resolved those threads. One `ApollonLocal` cleanup thread is still open, and Codacy is still reporting a failed check, so I’m keeping this as changes requested.

File: standalone/webapp/src/App.tsx [hunk 11]
```diff
@@ -1,63 +1,126 @@
...
+    <Suspense fallback={<AppLoadingScreen />}>
+      {!isHomeRoute && <Navbar />}
+      <div data-testid="editor-area" style={{ flex: 1, overflow: "hidden" }}>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
+          <Route path="/local/:id" element={<ApollonLocal />} />
+          <Route path="/playground" element={<ApollonPlayground />} />
+          <Route path="/imprint" element={<ImprintPage />} />
+          <Route path="/privacy" element={<PrivacyPage />} />
+          <Route path="/shared/:id" element={<ApollonWithConnection />} />
```
Claudia-Anthropica: @FadyGergesRezk [critical] This route uses `:id`, but `ApollonWithConnection` still reads `const { diagramId } = useParams()`. Opening `/shared/<id>?view=...` leaves `diagramId` undefined, so the editor never fetches or mounts the shared diagram. Keep the route param name as `:diagramId`, or update the component and tests to use `id` consistently.

In `standalone/webapp/src/App.tsx`, the shared route is declared as `/shared/:id` while `ApollonWithConnection` still reads `useParams().diagramId`. Fix this by changing the route to `/shared/:diagramId`, or by updating `ApollonWithConnection` and its tests to use `id` consistently. Add coverage for opening `/shared/<id>?view=COLLABORATE`.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The home dashboard work is mostly coherent, but the new shared URL shape does not pass the id into the shared editor right now. See the inline blocker before this can be approved.

standalone/webapp/src/App.tsx
```diff
@@ -1,63 +1,129 @@
...
+const AppLayout = () => {
+  const location = useLocation()
+  const isHomeRoute = location.pathname === "/"
+
+  return (
+    <Suspense fallback={<AppLoadingScreen />}>
+      {!isHomeRoute && <Navbar />}
+      <div data-testid="editor-area" style={{ flex: 1, overflow: "hidden" }}>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
```
Claudia-Anthropica, line 90: @FadyGergesRezk [high] Moving the local editor off `/` still leaves the fixture-based canvas e2e specs pointed at the dashboard. `injectFixtureIntoLocalStorage()` only writes `currentModelId`, and specs like `node-handle-grid`, `edge-resize-gap`, and the other edge specs still do `page.goto("/")`, so they now time out waiting for `.react-flow__viewport`; I confirmed `node-handle-grid.spec.ts` fails that way locally. Update the shared helper or every affected spec to navigate to `/local/${fixture.id}` after seeding.

In `standalone/webapp/tests/helpers/canvas.ts`, `injectFixtureIntoLocalStorage()` still supports specs that navigate to `/`, but `/` now renders the dashboard instead of the local editor. Fix this by adding and using a helper that seeds the fixture and navigates to `/local/<fixture.id>`, then update all e2e specs using `injectFixtureIntoLocalStorage` to open that route before `waitForCanvasReady`.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The shared route param and earlier dashboard blockers look fixed, but the route change still breaks the existing fixture-based canvas e2e specs. I also see CI still has failing checks, so I’m keeping this as changes requested.

standalone/webapp/src/pages/HomePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+        tabIndex={-1}
+      />
+
+      <main className="home-page-scrollbar app-scroll-y relative z-10 w-full min-h-0 flex-1 pb-24 md:pb-10">
+        <div className="flex w-full flex-col gap-5 px-4 pt-5 md:px-6 md:pt-6">
+          <Suspense fallback={<DiagramGallerySkeleton />}>
+            <DiagramGallery
+              highlightSharedDiagramId={highlightSharedDiagramId}
+              onNewDiagram={() =>
+                openModal("NEW_DIAGRAM", { dialogVariant: "home" })
```
Claudia-Anthropica, line 86: @FadyGergesRezk [high] These home CTAs still open create modals whose submit handlers were written for the old routing model: `NewDiagramModal` creates a model and navigates to `/`, and `NewDiagramFromTemplateModal` creates a model and navigates `..`. Now `/` is the dashboard, so creating a new diagram or template leaves users on home instead of opening the newly created `/local/<id>` editor; the same template flow is what is keeping the e2e job red. Use the created model id and navigate to `/local/${id}` from both create handlers.

In `standalone/webapp/src/components/modals/NewDiagramModal.tsx` and `standalone/webapp/src/components/modals/NewDiagramFromTemplateModal.tsx`, the create handlers still navigate to `/` or `..` after creating a model even though `/` now renders the dashboard. Change both handlers to navigate to `/local/<created-model-id>` after storing the model, and add/update coverage so creating a new diagram and creating a template both land on the local editor.

</details>

standalone/webapp/src/components/modals/ShareDashboardModal.tsx
```diff
@@ -0,0 +1,409 @@
...
+      setCopied(true)
+      if (copyTimeoutRef.current) clearTimeout(copyTimeoutRef.current)
+      copyTimeoutRef.current = setTimeout(() => setCopied(false), 2000)
+    } catch {
+      toast.error("Could not copy link.")
+    }
+  }
+
+  const handleModeSelect = (mode: DiagramView) => {
+    setActiveMode(mode)
```
Claudia-Anthropica, line 173: @FadyGergesRezk [medium] Changing the mode here only updates React state and the clipboard. The shared entry in `sharedDiagramStore` keeps the original `lastSharedView` from `addSharedDiagramEntry`, so after closing the modal the dashboard card still opens/copies the old Edit-mode link even if the user selected Collaborate or Feedback. Persist the selected mode with `updateSharedDiagramView` or `markSharedDiagramCopied` when the mode changes.

In `standalone/webapp/src/components/modals/ShareDashboardModal.tsx`, `handleModeSelect` updates `activeMode` but does not update the stored shared-diagram entry. Fix this by persisting the selected mode for `createdDiagramId` via the shared diagram storage helper when the mode changes, and cover that the dashboard card later uses the selected mode.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The dashboard routing is closer, but the create flows still use the old `/` editor assumption and CI is red on the template e2e path. I also found one shared-link mode persistence issue inline. Codacy is still action_required as well.

standalone/webapp/src/components/home/DiagramGallery.tsx
```diff
@@ -0,0 +1,1327 @@
...
+        title: persistentModelEntity.model.title,
+        type: persistentModelEntity.model.type,
+        lastModifiedAt: persistentModelEntity.lastModifiedAt,
+        favorite: persistentModelEntity.favorite ?? false,
+        model: persistentModelEntity.model,
+        source: "local" as const,
+        createdAt:
+          persistentModelEntity.createdAt ??
+          persistentModelEntity.lastModifiedAt,
+        lastViewedAt: persistentModelEntity.lastModifiedAt,
```
Claudia-Anthropica: @FadyGergesRezk [medium] The dashboard calls this field `lastViewedAt`, exposes a “Last viewed” sort/table column, and makes it the default sort, but local diagrams just copy `lastModifiedAt` and shared diagrams copy updated/created/shared timestamps. Opening a diagram never records a view, so the dashboard is actually sorting by modification/share date. Either persist/update a real last-viewed timestamp when cards/routes open, or rename the control/column to “Last modified” and sort on `lastModifiedAt`.

In `standalone/webapp/src/components/home/DiagramGallery.tsx`, `lastViewedAt` is populated from modification/share timestamps even though the UI labels and sorts it as “Last viewed”. Fix this by either persisting a real last-viewed timestamp and updating it from the local/shared open handlers, or renaming the sort option/table column to “Last modified” and using `lastModifiedAt` consistently. Add or update a home-page test that proves the chosen behavior.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The routing fixes look mostly addressed now, but I found one new dashboard data issue: “Last viewed” is not actually tracked, so the default sort and table column show modification/share time instead. Codacy is also still failing, so I’m keeping this as changes requested.

standalone/webapp/src/components/modals/ShareDashboardModal.tsx
```diff
@@ -0,0 +1,417 @@
...
+          ? { ...modelData, title: name.trim() }
+          : modelData
+
+      const collabName =
+        sessionStorage.getItem("apollon-collab-name") || randomCollabName()
+      sessionStorage.setItem("apollon-collab-name", collabName)
+
+      const { id: diagramId } =
+        await DiagramApiClient.createDiagram(modelToShare)
+      addSharedDiagramEntry(diagramId)
```
Claudia-Anthropica, line 146: @FadyGergesRezk [medium] After creating the shared copy, this only writes localStorage. The mounted `DiagramGallery` loaded `getSharedDiagramEntries()` once for the current source filter, so on the default All view the new shared diagram does not appear until the user reloads or toggles filters. Notify the gallery or reload shared entries after this write so the dashboard reflects the new share immediately.

In `standalone/webapp/src/components/modals/ShareDashboardModal.tsx`, `addSharedDiagramEntry()` writes the new shared diagram to localStorage but the mounted `DiagramGallery` only reloads shared entries when `diagramSource` changes. Fix this by notifying the gallery after same-tab shared-storage writes and reloading shared diagrams, or by passing an `onSharedDiagramCreated` callback from the gallery to the share modal. Add a home-page test that shares a diagram and verifies the new shared card appears without a reload.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest routing and e2e fixes look good, but the dashboard still has one stale shared-entry path. See the inline comment.

standalone/webapp/src/utils/sharedDiagramLinks.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export const getSharedDiagramViewBadge = (view: unknown): string =>
+  getSharedDiagramViewOption(view).badge
+
+export const buildSharedDiagramPath = (
+  diagramId: string,
+  view: DiagramView = DEFAULT_SHARED_DIAGRAM_VIEW
+): string =>
+  `/shared/${encodeURIComponent(diagramId)}?view=${encodeURIComponent(view)}`
```
Claudia-Anthropica, line 65: @FadyGergesRezk [medium] This gives the dashboard the new canonical `/shared/:id` URL shape, but the existing editor `ShareModal` still manually copies and navigates to `/${diagramID}?view=...` and never calls `addSharedDiagramEntry`. Sharing from the editor therefore produces a legacy URL and the new shared diagram is missing from the home dashboard. Route that modal through these helpers and register the shared entry there too.

In `standalone/webapp/src/components/modals/ShareModal.tsx`, the editor share flow still builds `/${diagramID}?view=...` manually and does not register the created shared diagram in `sharedDiagramStorage`. Update it to use `buildSharedDiagramUrl` / `buildSharedDiagramPath` and call `addSharedDiagramEntry(diagramID, { lastSharedView: viewType })` after creating the diagram so editor-created shares use `/shared/:id` links and appear on the home dashboard.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest dashboard routing fixes are mostly in place, but one share path is still outside the new URL/storage flow. The editor navbar share modal still creates legacy links and does not register the shared diagram for the home dashboard; see inline.

standalone/webapp/src/components/modals/LoadDiagramModal.tsx
```diff
@@ -1,22 +1,15 @@
...
 import { usePersistenceModelStore } from "@/stores/usePersistenceModelStore"
 import DeleteOutlineOutlinedIcon from "@mui/icons-material/DeleteOutlineOutlined"
 import { formatUpadtedDate } from "@/utils/date"
-import { useThemeStore } from "@/stores/useThemeStore"
 
 export const LoadDiagramModal = () => {
   const { closeModal } = useModalContext()
   const navigate = useNavigate()
-  const currentTheme = useThemeStore((state) => state.currentTheme)
+  const { id: currentDiagramId } = useParams()
```
Claudia-Anthropica: @FadyGergesRezk [high] `ModalWrapper` renders this modal outside `<Routes>`, so `useParams()` is `{}` here. `currentDiagramId` never matches, which means the currently open local diagram is not highlighted and the delete guard on line 29 no longer works, so the user can delete the diagram they are actively editing. Use the persisted `currentModelId` or derive the id from `useLocation()` instead of route params in this modal.

In `standalone/webapp/src/components/modals/LoadDiagramModal.tsx`, `useParams()` is used in a modal rendered outside `<Routes>`, so `currentDiagramId` is always undefined and the current-diagram delete guard is broken. Replace it with `usePersistenceModelStore((state) => state.currentModelId)` or an equivalent `useLocation()`-based path parser, and keep the selected/highlight/delete-disable logic tied to that id. Add a test that opens the load modal from `/local/:id` and verifies the current diagram cannot be deleted.

</details>

standalone/webapp/src/hooks/useDiagramIdFromPath.ts
```diff
@@ -16,13 +15,23 @@ const RESERVED_TOP_LEVEL_PATHS: ReadonlySet<string> = new Set([
...
-  const segment = location.pathname.split("/").filter(Boolean)[0]
-  if (!segment) return undefined
-  if (RESERVED_TOP_LEVEL_PATHS.has(segment)) return undefined
-  return segment
+  const segments = location.pathname.split("/").filter(Boolean)
+  const head = segments[0]
+  if (!head) return undefined
+  if (RESERVED_TOP_LEVEL_PATHS.has(head)) return undefined
+
+  if (head === "shared" || head === "local") {
```
Claudia-Anthropica, line 32: @FadyGergesRezk [medium] This makes `/local/:id` look like a connected diagram to `VersionHistoryButton`, but the version drawer is only rendered by `ApollonWithConnection`. On a local editor page the navbar now shows “Version History” and clicking it just toggles store state with no drawer/UI. Keep this helper scoped to shared/legacy connected routes, or make the button explicitly require a shared route.

In `standalone/webapp/src/hooks/useDiagramIdFromPath.ts`, returning an id for `/local/:id` makes `VersionHistoryButton` render on local-only editor routes even though no version drawer is mounted there. Change the helper or `VersionHistoryButton` so local routes return no connected diagram id, while `/shared/:id` and legacy shared redirects still work. Add a regression test for hiding the Version History button on `/local/:id`.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The dashboard routing is much closer, but I found two current-head regressions around editor chrome: the load dialog no longer knows which local diagram is open, and version history is exposed on local-only routes where no drawer exists. Please fix those before this lands.

standalone/webapp/src/components/AppLoadingScreen.tsx
```diff
@@ -0,0 +1,31 @@
...
+      className={`app-loading-screen app-loading-screen--${variant}`}
+      role="status"
+      aria-label={label}
+      aria-live="polite"
+    >
+      <div className="app-loading-content">
+        <div className="app-loading-logo-shell">
+          <img
+            className="app-loading-logo"
+            src="/assets/images/apollon-logo-v2.svg"
```
Claudia-Anthropica: @FadyGergesRezk [medium] Vite rewrites this logo path in `index.html`, but this runtime string stays as `/assets/images/apollon-logo-v2.svg` in the production JS bundle. The build only emits `/assets/apollon-logo-v2-*.svg`, so the Suspense loading screen shows a broken logo after the app has loaded. Import the SVG and use the imported URL, or move the file under `public/assets/images` so this absolute path exists.

In `standalone/webapp/src/components/AppLoadingScreen.tsx`, the `img` uses `/assets/images/apollon-logo-v2.svg`, but Vite does not rewrite that runtime string and the production build only emits the hashed `/assets/apollon-logo-v2-*.svg`. Fix this by importing the SVG from `assets/images/apollon-logo-v2.svg` and using the imported URL as `src`, or by moving the file into `public/assets/images` so the absolute path exists.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest fixes are in better shape: focused webapp lint/build and the shared-storage/load-modal unit tests pass locally. I found one new loading-screen asset path bug inline, and Codacy is still action_required on this head.

standalone/webapp/src/stores/usePersistenceModelStore.tsx
```diff
@@ -103,6 +148,35 @@ export const usePersistenceModelStore = create<PersistenceModelStore>()(
...
+                importedModels[model.id] = {
+                  id: model.id,
+                  model,
+                  lastModifiedAt: persistedAt,
+                  createdAt: persistedAt,
+                  favorite: false,
+                }
+              }
+
+              return {
```
Claudia-Anthropica, line 182: @FadyGergesRezk [high] Dropping the `currentModelId` update here regresses the legacy iOS migration contract: when no diagram is open, imported legacy diagrams should select the newest imported model. `pnpm --filter @tumaet/webapp exec vitest run src/services/legacyMigration.test.ts` now fails with `expected null to be "diag-new"`. Restore the previous selection logic while preserving the new `createdAt` and `favorite` fields.

In `standalone/webapp/src/stores/usePersistenceModelStore.tsx`, `importModels()` no longer sets `currentModelId` when importing legacy diagrams into a store with no current model, causing `legacyMigration.test.ts` to fail. Track the newest successfully imported model by `lastModifiedAt`, and return `currentModelId: state.currentModelId ?? latestId` along with `models: importedModels`. Keep the existing duplicate-skip behavior and new metadata fields.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The route/dashboard fixes are mostly in shape, but `importModels()` now breaks the legacy iOS migration test by leaving `currentModelId` null after migration. Codacy is also still failing on this head, so I am requesting changes.

standalone/webapp/src/utils/sharedDiagramLinks.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const buildSharedDiagramUrl = (
+  diagramId: string,
+  view: DiagramView = DEFAULT_SHARED_DIAGRAM_VIEW,
+  origin = typeof window !== "undefined" ? window.location.origin : ""
```
Claudia-Anthropica: @FadyGergesRezk [high] The dashboard share/copy paths call this helper without an origin, so on the Capacitor app the copied URL becomes `capacitor://localhost/shared/...`. That link cannot be opened outside the WebView, even though the editor share modal and version permalink now special-case native builds. Make the default share origin native-aware here so `ShareDashboardModal` and `DiagramCard` both copy the real web host.

In `standalone/webapp/src/utils/sharedDiagramLinks.ts`, `buildSharedDiagramUrl()` defaults to `window.location.origin`, which is `capacitor://localhost` in the native Capacitor app. Fix this by making the helper choose `serverURL` when `Capacitor.isNativePlatform()` and a server URL is configured, otherwise falling back to `window.location.origin`, so dashboard-created and dashboard-copied shared links use a browser-openable web URL.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest migration and asset-path fixes are in place, but the new dashboard shared-link helper still produces native-only `capacitor://localhost` URLs in the Capacitor app. See the inline comment.

Claudia-Anthropica (approved):
@FadyGergesRezk The latest head addresses the earlier route, dashboard, native link, and CI blockers. I did not find any new non-duplicated issues, and CI is green now.

tamang29:
Overall looks good, found some issue:
1) If i only have 2 diagrams, this is how the UI looks:
Screenshot 2026-06-07 at 21 33 14
2) When i open shared diagram, i cannot go back to home page "/", if i try to close the popup it doesnt.

https://github.com/user-attachments/assets/32a534f0-10aa-4d5b-b25a-26eaf9fc6efe

3) I think the highlighted part is repeated, could be same like in the shared diagram
Screenshot 2026-06-07 at 21 35 48
Screenshot 2026-06-07 at 21 36 40
4) I thought we were doing a Tab view for "Create new diagram" where user can choose between diagram types or existing template:
Screenshot 2026-06-07 at 21 37 09

tamang29:
1) There is no pagination in the homepage diagram list, could lead to heavy load on rendering on frontend side
2) 
image
this filtering is confusing as well, Alphabetical order (newest/oldest instead of ascending/decending), it makes sense for date created, last modified
3) there are two button to clear search field:
Screenshot 2026-06-09 at 14 19 44


4) Minor issue: font-family is inconsistent
Screenshot 2026-06-09 at 14 26 22

FelixTJDietrich (commented):
Some visual cues / example thumbnail would be good:
Image

Same here, we could have a preview:
Image

FelixTJDietrich (commented):
This looks a bit off, it also feels a bit too small for mobile potentially:
Image

"Delete this diagram?" is also the wrong question, it should be are you sure or so

FelixTJDietrich (commented):
We are still missing the help item in the header:
Image

Especially the Imprint and Privacy stuff is important

FelixTJDietrich (commented):
The colors in dark mode are inconsistent:
Compare
Image
with
Image

The latter ones are warmer

FelixTJDietrich (commented):
The buttons are also a bit inconsistent. Can you check because of consistencies? 

Image

Image

Image

Image

FelixTJDietrich (commented):
I also cannot command click or so on the diagram or so to easily open two diagrams in new tabs. Can you also ensure that this interaction works?

FadyGergesRezk:
Thanks @FelixTJDietrich — noted on the button inconsistency. This PR is scoped to introducing the new home page. The next PR will handle applying a unified theme, consistent button styles, and general UI polish across both the home page and the editor, covering both the webapp and library parts.

standalone/webapp/src/components/modals/NewDiagramModal.tsx
```diff
@@ -71,115 +137,213 @@ export const NewDiagramModal = () => {
...
+
+      const timeStapToCreate = new Date().getTime()
+      const templateModel =
+        typeof structuredClone === "function"
+          ? structuredClone(jsonData)
+          : JSON.parse(JSON.stringify(jsonData))
+
+      templateModel.title = newDiagramTitle
+
+      createModel(templateModel)
```
Claudia-Anthropica, line 199: @FadyGergesRezk [high] Template diagrams keep the fixed `id` from the JSON file here. Several templates already share the same top-level id (`Adapter`, `Bridge`, and `Command` all use `...90c4`), so creating another template overwrites the previous local diagram and sends the user back to the same `/local/:id`. Generate a fresh diagram id before `createModel()` and navigate to that new id.

In `standalone/webapp/src/components/modals/NewDiagramModal.tsx`, `createModel(templateModel)` stores template JSON under its fixed top-level id, and multiple template files share those ids. Fix this by assigning a new UUID to `templateModel.id` before calling `createModel()`, then navigate to `/local/${templateModel.id}`. Add a regression test that creates two different templates and verifies both local diagrams remain in the store with distinct ids.

</details>

standalone/webapp/src/webapp.css
```diff
@@ -1,14 +1,148 @@
...
+  --home-border-default: #a0aec1;
+  --home-border-strong: #8798b0;
+  --home-border-subtle: #c7d0dd;
+
+  /* Accent */
+  --home-accent-base: #0f3a66;
+  --home-accent-soft: #d3e7fb;
+  --home-accent-strong: #0f3a66;
+  --home-accent-ring: #0f3a66;
+  --home-accent-contrast: #ffffff;
```
Claudia-Anthropica: @FadyGergesRezk [low] Codacy is failing on CSS formatting in this new token block (`#ffffff` vs `#fff`, plus missing blank lines at later rules/comments). Please clean up the reported `webapp.css` annotations so the check goes green.

In `standalone/webapp/src/webapp.css`, Codacy reports 14 style issues in the new home CSS tokens/rules, starting with the long hex color on line 32. Fix the reported `color-hex-length` and empty-line-before violations throughout this file so the Codacy Static Code Analysis check passes.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The dashboard and routing pieces are mostly in shape, but template creation can overwrite existing template diagrams because the JSON ids are reused. Codacy is also still failing on this head, so I’m requesting changes.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated inline issues on this pass. Codacy is still failing on the current head and there is already unresolved feedback on the PR, so I’m keeping this as changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated inline issues on this pass. Webapp lint/build and the targeted shared-dashboard/unit tests passed locally, and the GitHub lint, server, visual, and version-sync checks are green. Codacy is still failing on the current head with one webapp.css formatting annotation, so I am keeping this as changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated inline issues on this head. Codacy is still failing with the current `standalone/webapp/src/webapp.css:92` custom-property spacing annotation, so I cannot approve yet.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated inline issues on this head. Codacy Static Code Analysis is still action_required on the current commit, so I cannot approve yet.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated inline issues on this head. Webapp lint/build passed locally, and GitHub lint, server, visual, and version-sync checks are green, but Codacy Static Code Analysis is still failing on the current commit with two  custom-property spacing annotations. That check needs to be green before approval.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated code issues on this head. Codacy Static Code Analysis is still action_required on the current commit with two  custom-property spacing annotations, so I cannot approve yet.

standalone/webapp/src/pages/ApollonWithConnection.test.tsx
```diff
@@ -5,6 +5,8 @@ import { toast } from "react-toastify"
 import { ApollonWithConnection } from "./ApollonWithConnection"
 import { EditorProvider, ModalProvider } from "@/contexts"
 
+const addSharedDiagramEntryMock = vi.fn()
+
 const FakeApollonEditor = vi.hoisted(
   () =>
     class FakeApollonEditor {
```
Claudia-Anthropica, line 12: @FadyGergesRezk [high] `ApollonWithConnection` now reads `editor.getDiagramMetadata()` and subscribes to diagram-name changes after mounting, but this fake editor doesn't implement either method. The focused webapp test run crashes three tests with `TypeError: editor.getDiagramMetadata is not a function`; add those methods to the fake so the tests exercise the loading/shared-entry behavior again.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/pages/ApollonWithConnection.test.tsx`, `FakeApollonEditor` is missing the diagram metadata API that `ApollonWithConnection` calls after `setEditor(instance)`. Add `getDiagramMetadata()` and `subscribeToDiagramNameChange()` methods to the fake editor, returning a title/type and a subscription id, so `pnpm --filter @tumaet/webapp exec vitest run src/pages/ApollonWithConnection.test.tsx` passes.\n\n</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest helper/refactor changes mostly look fine, and focused webapp lint/build passed locally. I found one current test regression in ApollonWithConnection.test.tsx, and Codacy is still action_required with the two webapp.css custom-property spacing annotations.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I did not find any new non-duplicated code issues on this head. Focused webapp unit tests passed locally, but Codacy Static Code Analysis is still action_required on the current head with the webapp.css:116 custom-property spacing annotation, so I cannot approve yet.

Claudia-Anthropica (approved):
@FadyGergesRezk I did not find any new non-duplicated code issues on this head. PR health checks, e2e, visual, Codacy, and the webapp image are green; the server Docker image hit a dependency install/network failure (sharp libvips ECONNRESET) that does not look caused by this webapp change.

Claudia-Anthropica (approved):
@FadyGergesRezk I did not find any new non-duplicated issues in the latest footer/about delta. Focused webapp lint/build and the new home-page footer/about e2e assertions passed locally; GitHub currently shows no failing checks.

FelixTJDietrich (approved):
I changed some things that I commented and also some further improvements so they don't get lost. Looks good now we can merge it! Thanks for the changes!

PR #663, opened 2026-05-03
feat: collaboration live-cursor

tamang29:
### Checklist

### Motivation and Context

This PR add a live-cursor to collaboration feature using ```y-protocol```

### Steps for Testing

1. Create any diagram.
2. Click "Share" from navbar and fill in the name.
3. Share the link for collboration.


FelixTJDietrich:
Can you update the privacy statement as well? 

Edit: This is already on main

FelixTJDietrich (changes requested):
Thanks @tamang29 for this great feature, I tested it and it generally works well. 

I have some improvement comments:

1. I noticed that here the dark mode is not properly themed:
Image

2. I think it would be nice if the name will be remembered and it is part of the initial "Share" dialogue, so we don't need a second dialogue, you can seed the name with a random one. But not sure, maybe we keep the second dialogue for when we joing a collaboration.

3. The selection color does not reflect my collaborator color, maybe this can be the same so you see who has selected what:
Image

FelixTJDietrich (approved):
I changed some stuff and fixed some collaboration issues should work more smoothly now :)

Issue #667, opened 2026-05-07
PDF/PNG export fails on larger diagrams


zhengr831-byte:
The JSON and svg format exporttool still works, but the graph cannot be exported as pdf/png. I don't exactly know the reason but hope there is a solution.

tamang29:
Hi @zhengr831-byte ,
is it possible to upload the json of the diagram that couldnt be exported as png/pdf?

PR #675, opened 2026-05-08
fix(export): render PNG/PDF reliably for complex diagrams (#667)

FelixTJDietrich:
### Checklist

### Motivation and Context

This PR completes https://github.com/ls1intum/Apollon/issues/667.

The reporter said *"the export button doesn't work when we create a little-complex graph"*. The cause was the previous client export pipeline rasterising SVG via `` → `<canvas>` → `toBlob`/`toDataURL`. Browsers cap canvas memory (Safari/iOS ~16 MP, Chrome ~268 MP, iOS Safari ~4096 px linear) — past those caps `toBlob` returns `null` and `toDataURL` returns the literal `"data:,"`. The user got a 0-byte file with no toast, no log: the menu just closed.

A second silent failure mode hit the Artemis LMS exam-integrity flow (server-side PDF): even on diagrams that exported, the JSDOM `getBBox` shim returned a constant `width: 10` for every text element. Apollon's bounds calculator then under-sized the diagram and clipped wide stereotype labels (`«Interface»`, `«Abstract»`, `«Enumeration»`) at the page edge.

#### Client PNG — replace canvas pipeline with resvg-wasm in a Web Worker

resvg renders SVG into its own pixel buffer in WebAssembly memory and emits a PNG byte-stream, so the browser canvas cap never applies. Capacity raised from ~16 MP (the Safari/iOS canvas ceiling, where the bug was reproduced) to **75 MP** — comfortably below the wasm32 4 GB ceiling at ~3-4× working set. Real Apollon class diagrams render at full 1.5× scale up to ~400 classes before the auto-clamp engages.

- `standalone/webapp/src/workers/svgToPng.worker.ts` — resvg-wasm with bundled Inter Regular/Bold/Italic/BoldItalic (text fidelity matters for stereotype labels).
- `standalone/webapp/src/utils/svgToPng.ts` — Worker singleton + idle-terminate (5 min) + 60 s render timeout + typed errors (`RasterTooLargeError`, `RasterTimeoutError`).
- `standalone/webapp/src/hooks/useExportAsPNG.ts` — thin wrapper, returns `{ clamped, appliedScale }` so the navbar can warn on degraded output.

#### Client PDF — vector via svg2pdf.js + jsPDF, with Inter properly registered

- Drops the previous client-PDF code path that embedded a rasterised PNG. PDF stays vector and tiny (~400 KB with Inter subset).
- `standalone/webapp/src/utils/registerInterFonts.ts` — registers all four Inter weight/style combinations with jsPDF before `svg2pdf` walks the SVG. Without this, svg2pdf falls back to Helvetica which only has 400/700 weights and silently drops every `<text>` at the 600-weight Apollon emits for headers and at the 85%-size stereotype tspans.
- `standalone/webapp/src/utils/preProcessSvgForPdf.ts` — DOM transform that normalises Apollon's compat-mode SVG for svg2pdf: inlines nested `<svg>` wrappers, flattens multi-tspan stereotype texts, resolves percentage font-sizes against parent size, collapses `font-family` chains to plain `Inter`.
- MediaBox clamped at 14,400 pt (svg2pdf's hard ceiling) for very large diagrams; output stays vector.

#### Server PDF — Artemis exam-integrity path

Symmetric to client: same Inter fonts, same preprocessor, same MediaBox clamp. Worker thread renders synchronously; the resource layer queues + auto-restarts on crash + recycles every 15 successful renders to bound JSDOM memory growth.

- `POST /api/converter/pdf` is the only server export route. PNG is rendered client-side only — the canonical Artemis archive format is PDF, so server-side PNG was dropped.
- `standalone/server/src/services/conversion-service.ts` — replaced the 20×300 ms Apollon-library retry loop with a single eager `require`; hoisted JSDOM polyfills to module load; fixed the `getBBox` shim to return realistic text dimensions (text length × font-size × 0.6 em) so Apollon's bounds calculator no longer undersizes diagrams with long stereotype labels.
- `standalone/server/src/resources/conversion-resource.ts` — added `CONVERTER_WORKER_MAX_RENDERS` (default 15) so the worker is recycled before the JSDOM document tree leaks past the 256 MB heap cap.

#### Navbar UX

`await` each export, branched `toast.error` by typed error class, progress toast on raster paths, busy-state guard against double-click. The signature failure mode — silent failure with no user feedback — is closed.

### Steps for Testing

1. **Manual smoke** (offline mode):
   - `npm run dev`
   - Build a class diagram with stereotyped classes (Interface, Abstract, Enumeration)
   - File → Export → As PNG (white + transparent), As PDF, As SVG, As JSON
   - All five formats download; PNG/PDF preserve every stereotype label

2. **Original #667 reproduction**:
   - Import any class diagram with 30+ classes
   - Both PNG and PDF export cleanly with all labels visible

3. **Server endpoint** (Artemis):
   ```
   curl -X POST http://localhost:8080/api/converter/pdf \
     -H 'Content-Type: application/json' \
     --data-binary @standalone/webapp/tests/fixtures/class-diagram.json \
     -o /tmp/api.pdf
   ```

4. **Automated**:
   - `npm run lint` — ✓
   - `npm run format:check` — ✓
   - `npm run build` — ✓
   - Library: `npm test` — 781/781
   - Webapp: `npm test --workspace=@tumaet/webapp` — 68/68
   - E2E: `npx playwright test tests/e2e/export.spec.ts` (in webapp) — 5/5, includes a PDF text-extraction assertion that every stereotype literal is preserved

### Screenshots

Same export menu, different output:

| Diagram | Before | After |
|---|---|---|
| Class diagram with stereotypes — PNG | 0-byte file (silent), or stereotypes clipped | Every stereotype renders verbatim, every node labeled |
| Class diagram — PDF | Raster blob; stereotypes missing; corrupt on complex graphs | Pure vector (~400 KB with Inter subset); zooms infinitely; every label intact |
| 20,000 × 30,000 px diagram — PDF | jsPDF rejected; opaque error swallowed | MediaBox clamped to 14,400 pt; full vector output |
| Failure modes | Menu closes, no feedback | `toast.error` with typed message |

### Limits & footprint

| | Before | After |
|---|---|---|
| Client PNG ceiling | ~16 MP (Safari/iOS canvas cap, silent failure past it) | 75 MP (resvg-wasm in Worker, bypasses canvas) |
| Class count at full 1.5× scale | ~80–100 before silent clip | ~400 before auto-clamp |
| Client PDF | Raster blob (canvas-bound) | Vector, ~400 KB |
| Server PDF | pdfmake with raster-embedded SVG | Vector via svg2pdf.js + JSDOM, mirrors client output |
| Server PNG | absent | (intentionally absent — client-only) |
| Worker memory | unbounded | recycled every 15 renders (env-tunable) |
| User feedback on failure | none | typed-error toast with actionable message |

Battle-tested on 13 diagram fixture types (every Apollon diagram type) plus synthetic class diagrams up to 150 classes:

| Classes | PNG ms | PNG KB | PDF ms | PDF KB |
|--:|--:|--:|--:|--:|
| 10 | 411 | 102 | 558 | 380 |
| 50 (the #667 case) | 1,055 | 575 | 1,118 | 386 |
| 100 | 2,028 | 1,083 | 2,063 | 393 |
| 150 | 3,288 | 1,527 | 3,092 | 399 |

Past 150 classes the Apollon library's hard-coded 3 s React-Flow init timeout (apollon-editor.tsx:287) starts to fire under JSDOM. Real-world diagrams don't approach that limit; addressing it would require headless Chromium server-side, which is out of scope for this PR.

### Tunable env vars (server)

| Var | Default | Purpose |
|---|---:|---|
| `CONVERTER_TIMEOUT_MS` | 30 000 | Per-render timeout |
| `CONVERTER_MAX_QUEUE_LENGTH` | 20 | Backpressure (503 above) |
| `CONVERTER_WORKER_MAX_OLD_SPACE_MB` | 256 | Worker heap cap |
| `CONVERTER_WORKER_STACK_MB` | 8 | Worker stack |
| `CONVERTER_WORKER_MAX_RENDERS` | 15 | Recycle worker after N renders to bound JSDOM memory |


standalone/webapp/src/workers/svgToPng.worker.ts
```diff
@@ -0,0 +1,126 @@
...
+function ensureInitialised() {
+  if (bootPromise) return bootPromise
+  bootPromise = (async () => {
+    try {
+      const [, regular, bold] = await Promise.all([
+        initWasm(fetch(wasmUrl)),
+        fetchBytes(interRegularUrl),
+        fetchBytes(interBoldUrl),
+      ])
+      return { fontBuffers: [regular, bold] }
```
Claudia-Anthropica: @maximiliansoelch [medium] Only Regular + Bold are passed to resvg here, but `Inter-Italic.ttf` / `Inter-BoldItalic.ttf` are bundled in the same folder and registered on the PDF side via `registerInterFonts.ts`. Apollon's `HeaderSection.tsx:53` emits `font-style="italic"` for `ClassType.Abstract` headers in compat-mode SVG. With no italic font in `fontBuffers`, resvg-wasm has nothing to fall back to and renders abstract class names as upright Regular — a silent fidelity regression that the PDF path doesn't have. PR description also claims all four weights are bundled, so the worker is the outlier.

```suggestion
      const [, regular, bold, italic, boldItalic] = await Promise.all([
        initWasm(fetch(wasmUrl)),
        fetchBytes(interRegularUrl),
        fetchBytes(interBoldUrl),
        fetchBytes(interItalicUrl),
        fetchBytes(interBoldItalicUrl),
      ])
      return { fontBuffers: [regular, bold, italic, boldItalic] }
```

In `standalone/webapp/src/workers/svgToPng.worker.ts`, the PNG worker only loads `Inter-Regular.ttf` and `Inter-Bold.ttf` into `fontBuffers`, so resvg-wasm has no italic glyphs and renders `font-style="italic"` text (e.g. abstract class names from `library/lib/components/svgs/nodes/HeaderSection.tsx:53`) as upright Regular. Add `import interItalicUrl from "./fonts/Inter-Italic.ttf?url"` and `import interBoldItalicUrl from "./fonts/Inter-BoldItalic.ttf?url"`, fetch their bytes in the same `Promise.all`, and include them in `fontBuffers` so the PNG output matches the PDF path that already registers all four weights via `registerInterFonts.ts`.

</details>

standalone/webapp/src/utils/svgToPng.ts
```diff
@@ -0,0 +1,256 @@
...
+      timeoutMs
+    )
+  } catch (err) {
+    // Translate wasm out-of-memory into a typed error so the navbar branches.
+    if (err instanceof RasterTimeoutError) throw err
+    if (err instanceof RasterTooLargeError) throw err
+    const e = err as Error
+    if (
+      e.name === "RangeError" ||
+      /memory|allocation|out of memory/i.test(e.message ?? "")
```
Claudia-Anthropica: @maximiliansoelch [low] Matching `/memory|allocation|out of memory/i` against any worker error message can mis-classify unrelated failures as `RasterTooLargeError` — e.g. a `RangeError` from a malformed SVG, or any future error whose message happens to mention "memory". Since the worker is fully under your control, consider tagging OOM at the source: have `svgToPng.worker.ts`'s catch detect resvg's OOM (`RangeError` with `"out of bounds"`, or wrap `resvg.render()` specifically) and send back `error.name = "RasterTooLargeError"`, then the wrapper just checks the name without sniffing strings.

In `standalone/webapp/src/utils/svgToPng.ts` around line 240, the OOM detection uses a substring regex on `error.message` which can mis-classify unrelated errors as `RasterTooLargeError`. Move the classification into `standalone/webapp/src/workers/svgToPng.worker.ts`: catch errors thrown by `new Resvg(...)` / `resvg.render()` and set `error.name = "RasterTooLargeError"` for `RangeError`/wasm OOM, then in `svgToPng.ts` replace the regex with a check on `e.name === "RasterTooLargeError"` only.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Solid rewrite of the export pipeline — the resvg-wasm-in-worker approach, the typed-error toasts, and especially the PDF text-extraction e2e assertion are nice. One real fidelity bug worth fixing before merge: the PNG worker only bundles Inter Regular + Bold, so the italic fonts you ship in `workers/fonts/` and register on the PDF side never get loaded for PNG. Abstract class names lose their italic styling in PNG only. Plus a small note on the OOM detection heuristic.

standalone/webapp/src/workers/svgToPng.worker.ts
Claudia-Anthropica: @FelixTJDietrich Still seeing the underlying issue after the export code moved into `library/lib/export`: `svgToPng.ts` still only loads Regular + Bold, and `svgToPdf.ts` now aliases `italic` to those upright faces instead of embedding an italic font. `HeaderSection.tsx` still emits `font-style="italic"` for abstract class headers, so PNG/PDF exports still lose that styling. Please either bundle/register the italic faces for both export paths, or deliberately normalize the exported SVG so it no longer claims italic styling.

standalone/webapp/src/utils/svgToPng.ts
Claudia-Anthropica: @FelixTJDietrich This is still present in the moved implementation at `library/lib/export/svgToPng.ts`: the catch block treats any `RangeError` or message containing `memory`/`allocation` as `RasterTooLargeError`. This is lower priority than the font fidelity issue, but the same false-positive risk remains.

Claudia-Anthropica (changes requested):
@FelixTJDietrich I rechecked the previous feedback. The export code moved into the library, but abstract class headers still export as italic SVG text while PNG/PDF only embed or alias upright Inter faces, so the medium fidelity issue is still open. The OOM classification comment is still open too, though that one is lower priority.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit only fixes the PNG byte type for Blob construction, so the two previously requested export threads are still unresolved. Keeping this as changes requested until those are addressed.

standalone/webapp/src/workers/svgToPng.worker.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The normalizer makes the export output honest about not using italics, and both PNG/PDF paths call it.

standalone/webapp/src/utils/svgToPng.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The broad message regex is gone and the malformed-SVG test covers the false-positive case.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed and the CI checks are green. Approving.

library/lib/export/preProcessSvgForPdf.ts
```diff
@@ -0,0 +1,145 @@
...
+ *     standard font; collapse any Inter-led list to plain `"Inter"`.
+ *
+ * Mutates `root` in place and returns it.
+ */
+
+const SVG_NS = "http://www.w3.org/2000/svg"
+
+/** Attributes a split `<text>` must keep from its parent so svg2pdf's
+ *  per-element lookup matches the original SVG semantics. */
+const INHERITED_TEXT_ATTRS = [
```
Claudia-Anthropica, line 30: @FelixTJDietrich [medium] Splitting multi-`tspan` text drops the parent `font-size`. Compat SVG only guarantees `font-size` on the parent `<text>` for common cases like `HeaderSection` class names and `MultilineText`; child tspans often only carry `x`/`y`. After this split, those lines render at svg2pdf/jsPDF's default size instead of Apollon's measured size, so PDF labels can grow or shift. Add `font-size` to the inherited attrs before copying tspan overrides.

In `library/lib/export/preProcessSvgForPdf.ts`, `flattenMultiTspans()` does not copy inherited `font-size` from the source `<text>` onto the split `<text>` elements. Add `font-size` to `INHERITED_TEXT_ATTRS` and add or adjust a unit test where one tspan lacks its own font-size but inherits the parent size.

</details>

library/package.json
```diff
@@ -127,6 +136,11 @@
     "yjs": "13.6.20",
     "zustand": "catalog:"
   },
+  "optionalDependencies": {
+    "@resvg/resvg-wasm": "2.6.2",
```
Claudia-Anthropica: @FelixTJDietrich [low] These renderer versions are now pinned in more than one workspace (`library` optional deps and `standalone/webapp` deps, with `jspdf` already present in the webapp). Repo convention is to put shared versions in `pnpm-workspace.yaml#catalog` and reference them as `catalog:` from each manifest, otherwise Renovate/version bumps can drift.

In `library/package.json` and `standalone/webapp/package.json`, `@resvg/resvg-wasm`, `jspdf`, and `svg2pdf.js` are shared across workspaces but have literal version strings. Move those pins into `pnpm-workspace.yaml` `catalog` and change both manifests to use `"catalog:"`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The export subpath and webapp build path look mostly solid, and the previous font/OOM feedback is addressed. There is still a PDF fidelity bug in the tspan splitter that needs fixing before merge, plus a small catalog cleanup.

library/lib/export/preProcessSvgForPdf.ts
Claudia-Anthropica: @FelixTJDietrich Still open on the current head: `font-size` is still missing from `INHERITED_TEXT_ATTRS`, so split tspans can still lose the parent text size.

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The split text elements inherit the parent font size and the regression test covers the tspan without its own size.

library/package.json
Claudia-Anthropica: @FelixTJDietrich Nice, the shared renderer versions are in the catalog now and both manifests reference `catalog:`.

Claudia-Anthropica (approved):
@FelixTJDietrich Looks good on the current head. Previous feedback is addressed, the export helpers build cleanly, and CI is green.

standalone/webapp/src/components/navbar/NavbarFile.tsx
```diff
@@ -68,6 +103,39 @@ export const NavbarFile: FC<Props> = ({ color, handleCloseNavMenu }) => {
...
+  const runExport = useCallback(
+    async (
+      format: ExportFormat,
+      action: () => Promise<ExportRunResult | void>
+    ) => {
+      if (busyFormat) return
+      closeMainMenu()
+      setBusyFormat(format)
+      try {
+        await toast.promise(action(), {
```
Claudia-Anthropica, line 120: @FelixTJDietrich [medium] This now shows a success toast whenever the export promise resolves, but the SVG and JSON hooks still catch mobile `Filesystem`/`Share` failures and return. On iOS/Android that means a failed SVG/JSON export resolves here and the user sees `SVG exported.` or `JSON exported.`; make those hooks reject on failure before adding success toasts.

In `standalone/webapp/src/components/navbar/NavbarFile.tsx`, `toast.promise(action(), ...)` treats any resolved export action as a successful export, but `useExportAsSVG` and `useExportAsJSON` catch mobile `Filesystem`/`Share` errors and resolve. Fix this by changing `standalone/webapp/src/hooks/useExportAsSVG.ts` and `standalone/webapp/src/hooks/useExportAsJSON.ts` to rethrow mobile export errors after logging, and make editor-missing/falsy-export paths reject instead of returning silently.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The export helper changes still look good, but the new unified success toast can report failed mobile SVG/JSON exports as successful because those hooks swallow errors. Please fix that path before merge.

standalone/webapp/src/components/navbar/NavbarFile.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. SVG and JSON exports now reject on missing editor and mobile Filesystem/Share failures, so the toast can no longer show a false success.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed now. Approving.

PR #677, opened 2026-05-09
feat: embed Apollon diagrams in GitHub, GitLab, and iframes

FelixTJDietrich:
### Summary

Adds first-class **embedding** for Apollon diagrams, **rebuilds the editor Share dialog**, and scales the render pipeline to the host.

**Embed** — two public, server-rendered surfaces + a webapp panel:

- **SVG preview** (`GET /api/diagrams/:id/preview.svg`) — a cacheable `image/svg+xml` for `` / GitHub-Camo embedding in READMEs, issues, and PRs. It carries an opaque white background so a diagram dropped into a **dark-mode** README stays legible instead of rendering dark-on-dark.
- **Embed page** (`GET /embed/:id`) — a lightweight, no-JS, iframe-friendly HTML view (Notion, Confluence, GitLab Pages, VS Code preview) with an "Open in Apollon" button; its chrome follows the system light/dark theme.
- **Share dialog** gains an **Embed** panel — one snippet at a time behind a format selector (Markdown / Markdown without link / iframe), in a multi-line code field with copy. The Markdown snippet pairs the inline diagram with a hosted **"Open in Apollon"** badge (`/api/embed/button.svg`) that matches the reader's light/dark theme.

**Share dialog rework** — the in-editor Share dialog is rebuilt on a **create-once** model (fixing a real duplicate-diagram bug where each mode button minted a new server copy): share once to get **one link**, then pick what people can do with it from a dropdown that only rewrites `?view=` — never a second copy. Sharing now **leads with Collaborate by default**, with Edit one click away. Copy is explicit; nothing navigates until you choose **Open diagram**. The link UI is shared with the dashboard share dialog.

**Reads keep a diagram alive** — opening *or embedding* a diagram now slides its 120-day TTL forward (throttled to ≤1 write/day per diagram, best-effort), so a diagram embedded in a README no longer silently expires while it's still being viewed. Previously only a write refreshed the TTL.

The diagramId is the bearer — the same access model as the editor URL, so embedding adds no privilege and needs no token or sign-in. The branch was reconciled onto `main` after #765 shipped the unified conversion API; embed renders through that `ConversionResource` (upgraded here from a single serial worker to a self-calibrating pool).

### Release note

Embed your diagrams anywhere — drop one into a README, issue, or PR as a Markdown image that renders inline, updates as you edit, and comes with an "Open in Apollon" button that matches light and dark themes, or paste an iframe snippet. The Share dialog is cleaner too: it leads with live collaboration, shares once for a single link, and lets you switch access without ever creating duplicate copies.

### Screenshots

> Hosted on the `pr-677-assets` branch (captured from the live stack, not mockups), so they render here without bloating this PR's diff.

**Rebuilt Share dialog** — share once, then choose access; the Embed panel gives one copy-paste snippet at a time.

| Create the link | Link + access + embed | Access modes |
| --- | --- | --- |
| Share dialog, create state | Share dialog with link, access dropdown and embed panel | Access-mode dropdown |

**Embed output** — the Markdown image (what GitHub renders inline) and the standalone `/embed` page (for iframes), which adapts its chrome to light/dark while keeping the diagram canvas readable.

| README image (`preview.svg`) | Embed page — light | Embed page — dark |
| --- | --- | --- |
| Rendered SVG preview | Embed page in light mode | Embed page in dark mode |

### Implementation notes

- **Share dialog.** Both share dialogs (editor + dashboard) now share `useShareableDiagram` (the create-once state machine) and `ShareLinkRow` (link field + explicit copy + access dropdown, with the access modes ordered Collaborate → Edit → Add feedback → View feedback). The editor dialog opens as a home-dialog variant so the dropdown isn't clipped, is mobile-safe (verified by an e2e overflow test on a phone viewport), and is seeded from the route so an already-shared diagram opens straight on its link.
- **Sliding TTL.** A read refreshes the diagram's `EXPIRE` back to the full window — on the editor `GET` and on every embed access (`200`, `304` Camo revalidation, and `HEAD`). It's throttled (only refreshes once the remaining TTL has decayed past a 1-day window, so a busy embed can't write-amplify into an `EXPIRE` per fetch) and best-effort (a refresh failure never breaks the read). The GDPR record-of-processing note was updated to match: retention is now "120 days from the last access," not "from the last write."
- **Dark-mode legibility.** The library SVG export is transparent and its fills are tuned for a light canvas, so the served `preview.svg` gets a white `<rect>` injected as its first child (sized to the export's own — possibly negative-origin — `viewBox`). It's baked in inside the render cache, so `preview.svg` and `/embed` share it; the PDF path is untouched.
- **Reconciliation onto #765.** The branch's earlier forked conversion stack (priority queue, LRU/coalescing, DOMPurify, jspdf/svg2pdf, the #675 client-export rework) was dropped as superseded; embed renders through the shared `ConversionResource`, and `LibrarySchemaVersion` validation was dropped to match `main`'s unpinned-`version` stance.
- **Security is CSP, not a sanitizer.** The embed page sets `default-src 'none'` with no `script-src`, so script in the library-generated SVG cannot execute and the SVG is inlined verbatim; `img-src 'self' data:` is the load-bearing CSS-exfiltration boundary. The diagram title is the only HTML-interpolated value and is escaped.
- **HTTP correctness.** Weak `ETag` = `headRev` read atomically with the body (one `MULTI`, no stale-tag race), RFC 7232 §3.2 weak comparison (matches the strong tag the PUT route mints), `Cache-Control: public, max-age=60, stale-while-revalidate=86400`, conditional `GET`, `HEAD` without rendering, transient `503 RENDERER_BUSY` + `Retry-After` header, errors never cached.
- **Throughput.** Render is overhead-bound (a cold worker pays ~1 s of JSDOM + library boot), so the previous single serial worker was the ceiling under load. The pipeline now self-calibrates a worker pool — `min(cores − 1, memory budget / per-worker footprint)` — with a queue cap, a queue-wait deadline, per-format deadlines, crash backoff, and idempotent worker recycling, scaling render concurrency with the host's cores and degrading cleanly to serial on a 1-core VM, with no deployer tuning. The embed surface adds an in-process render cache + single-flight keyed by `(id, headRev)`, so a Camo fan-out collapses to one render per save.
- **Local dev.** `pnpm run dev` runs the server from TypeScript via tsx, where the conversion worker has no compiled `.js`; a small `.mjs` bootstrap loads it through tsx's programmatic loader so embed/preview/PDF render in dev (production still spawns the compiled worker directly). The Vite dev proxy also forwards `/embed` so the iframe snippet resolves instead of falling through to the SPA.

### Steps for testing

1. **Run the stack:** `pnpm install && pnpm run build:lib && pnpm run dev`.
2. **Share a diagram** from the editor's **Share** button → **Create share link** → the dialog shows the link (Collaborate by default), an access dropdown, and the **Embed** panel (re-opening Share on an already-shared diagram lands straight on the link).
3. **SVG endpoint:** `curl -I http://localhost:8000/api/diagrams/<id>/preview.svg` → `200`, `image/svg+xml`, `ETag: W/"<n>"`, the `Cache-Control` above. Re-issue with `If-None-Match: W/"<n>"` (or the strong `"<n>"`) → `304`; `HEAD` returns headers without rendering. The body starts with a white `<rect>` so it reads on a dark background.
4. **Badge:** `curl -I http://localhost:8000/api/embed/button.svg` → `200`, `image/svg+xml`, immutable cache; open it in a browser and toggle OS dark mode to see it adapt.
5. **Embed page:** open `http://localhost:8000/embed/<id>` → inline SVG + "Open in Apollon" (→ `/shared/<id>?view=EDIT`); CSP allows `frame-ancestors *`; toggle OS dark mode to see the chrome flip.
6. **Paste the Markdown snippet** into a private repo's README and confirm GitHub Camo renders the diagram + badge inline.
7. **Tests:** `pnpm --filter @tumaet/server run test` (139 green) and `pnpm --filter @tumaet/webapp run test` (240 green).

### Checklist

- [ ] Linked to a related issue (no tracking issue for this work)
- [x] Tests added or updated
- [x] Documentation updated (Conversion API limits & tuning; GDPR record-of-processing retention)


FelixTJDietrich:
## Rebased on PR #675 (`fix/issue-667-png-pdf-export`)

This branch now incorporates @ls1intum/PR #675's conversion-pipeline rewrite as a synthetic merge commit (`Merge branch 'pr-675' into merge-target`). The two embed commits sit cleanly on top of that merged shape.

**Until PR #675 is merged into `main`,** this PR's GitHub diff will show #675's changes alongside mine. To review just the embed work:

```sh
git fetch origin pull/675/head:pr-675 && git fetch origin pull/677/head:pr-677
git diff pr-675...pr-677 -- standalone/
```

That cross-PR diff is the canonical "just the embed feature" view (~3,000 lines, 102 server tests, 781 library tests, all green locally).

### What changed during the rebase

- **`conversion-service.ts`** adopted PR #675's rewrite (free-function `convertModelToSvg`, module-scope JSDOM polyfills, eager top-level `require`, realistic `getBBox` text-metrics shim). I appended a free-function `sanitizeSvg` for the embed surface — PDF skips it, embed applies it.
- **`conversion-renderer.ts`** (new in #675) gained a sibling `renderSvgSafe` that calls `convertModelToSvg` + `sanitizeSvg` for the worker SVG mode.
- **`conversion-resource.ts`** is the merged shape: PR #675's worker recycle counter (`CONVERTER_WORKER_MAX_RENDERS`, default 15) **plus** my SVG-priority `dequeueNext` and PDF anti-starvation backstop. The recycle counter increments on *any* successful render — heap pressure is mode-agnostic. The `createWorker`/`handleWorkerMessage` methods are `protected` so the dequeue test can subclass.
- **`workers/conversion-worker-thread.ts`** is now a thin 2-mode dispatcher (`pdf` → `renderPdf`, `svg` → `renderSvgSafe`) on top of #675's renamed worker file. The old `pdf-conversion-worker-thread.ts` is gone.
- **`svg-source.ts`** uses #675's free `convertModelToSvg` + `sanitizeSvg` directly, no `ConversionService` class allocation.

### What I did NOT do

- I did **not** apply `preProcessSvgForPdf` to the embed SVG path. Its transforms (collapsing CSS font-stacks, splitting tspans, replacing nested `<svg>` with `<g>`) are svg2pdf.js workarounds. Browsers handle the library output natively; running them on embed would degrade rendering (kills the Inter → system-ui fallback chain, etc.).
- I did **not** use #675's PDF Inter-font registration on the embed path. Embeds inline into HTML where the user's browser handles fonts.

### After PR #675 merges

I'll rebase this PR on the new `main`, dropping the synthetic merge commit. The diff will then shrink to just the embed work.

standalone/server/src/routes/embed.ts
```diff
@@ -0,0 +1,310 @@
...
+function classifyRenderError(err: unknown): ApiError | null {
+  const message =
+    err instanceof Error ? err.message : typeof err === "string" ? err : ""
+  if (RENDER_OVERLOADED_MESSAGES.has(message)) {
+    // 503 RENDERER_BUSY is an explicit "transient overload" signal —
+    // distinct from REDIS_UNAVAILABLE so dashboards / alerts /
+    // log-pipelines filtering on the Redis-outage code don't false-
+    // positive on render saturation.
+    return new ApiError(503, "RENDERER_BUSY", "Render pipeline busy", {
+      retryAfterSeconds: 1,
```
Claudia-Anthropica: @FelixTJDietrich [medium] The docstring above (lines 62-66) says the 503 should carry a `Retry-After` so Camo / browsers back off rather than caching the failure. But this only puts `retryAfterSeconds: 1` in the JSON body — `errorHandler` never sets the `Retry-After` HTTP header, and CDNs/proxies (Camo included) don't parse our body for backoff. Under sustained queue saturation, Camo will treat the 503 as a terminal failure and cache it against the diagramId for its full timeout window. Either set the header in the route before throwing, or extend `ApiError`/`errorHandler` so meta-keyed `retryAfterSeconds` gets emitted as both a header and a body field.

In `standalone/server/src/routes/embed.ts` around line 75, the 503 RENDERER_BUSY response is missing the `Retry-After` HTTP header that the surrounding docstring promises. Update either `errorHandler` in `standalone/server/src/http/middleware/errors.ts` to emit `Retry-After: <seconds>` when an `ApiError` has `meta.retryAfterSeconds`, or set `res.setHeader('Retry-After', '1')` from the embed routes before throwing. Add a header-level assertion to the existing `RENDERER_BUSY` integration tests so the regression can't reland silently.

</details>

standalone/server/src/routes/_schemas.ts
```diff
@@ -17,9 +17,25 @@ export const VersionId = z
...
+ * version, and the `toUmlModel` projection downstream can widen
+ * `string` → the literal type via a runtime-validated assertion.
+ */
+export const LibrarySchemaVersion = z
+  .string()
+  .regex(
+    /^4\.\d+\.\d+(?:-[\w.]+)?(?:\+[\w.]+)?$/,
+    "version must match the library schema pattern 4.<minor>.<patch> with an optional SemVer pre-release and/or build-metadata suffix"
+  )
+
```
Claudia-Anthropica: @FelixTJDietrich [low] The error message advertises full SemVer pre-release / build-metadata support, but `[\w.]+` excludes hyphens (which valid pre-release identifiers allow — e.g. `4.0.0-x.7.z-92`) and includes underscores (which SemVer rejects). Today's library only emits `4.X.Y` so this never bites in practice, but the API also takes user-supplied versions on POST/PUT, so a contributor following the docstring will be confused when a legitimate SemVer string 422s. Swap the inner class to `[0-9A-Za-z.-]` for the pre-release group and `[0-9A-Za-z.+-]` (or simply `[0-9A-Za-z.-]`) for build-metadata to match the spec.

In `standalone/server/src/routes/_schemas.ts` around lines 30-35, the `LibrarySchemaVersion` regex uses `[\w.]+` for SemVer pre-release and build-metadata segments, which omits the hyphen permitted by SemVer 2.0 and admits underscores it forbids. Replace each `[\w.]+` with `[0-9A-Za-z.-]+` to actually match the SemVer grammar, and extend the existing schema test in `standalone/server/src/__tests__/diagrams.int.test.ts` with a case like `4.0.0-x.7.z-92` to lock the contract.

</details>

standalone/server/src/routes/embed.ts
```diff
@@ -0,0 +1,310 @@
...
+async function readDiagramWithRev(
+  redis: Redis,
+  diagramId: string
+): Promise<{
+  diagram: NonNullable<Awaited<ReturnType<typeof readDiagram>>>
+  headRev: string
+} | null> {
+  const diagram = await readDiagram(redis, diagramId)
+  if (!diagram) return null
+  const rev = (await redis.hGet(k.diagramMeta(diagramId), "headRev")) ?? "0"
```
Claudia-Anthropica: @FelixTJDietrich [low] `readDiagram` and `hGet(headRev)` are two separate awaits, so a concurrent PUT between them lets you observe the old `diagram` (with `updatedAt = T1`) but the new `headRev = R2`. The route then caches a T1-shaped render under key `id:T1` and emits ETag `W/"R2"`, so any client that fetched during the race ends up with stale SVG pinned to a fresh ETag — `If-None-Match: W/"R2"` returns 304 forever (until the next PUT bumps headRev again). Window is tiny in practice but trivial to close: fetch headRev with HMGET alongside the head, or just use the diagram's own `updatedAt` for the ETag too (it bumps in the same MULTI as `headRev`).

In `standalone/server/src/routes/embed.ts` around lines 106-118, `readDiagramWithRev` does two sequential Redis reads, leaving a race where a concurrent PUT yields an inconsistent (diagram, headRev) pair. Fix by either (a) issuing one `HMGET` on the meta hash for `headRev` in the same `MULTI` block as the head read, or (b) using `diagram.updatedAt` directly as the cache key + ETag source since it bumps atomically with `headRev` in `saveHead`.

</details>

standalone/server/src/services/svg-source.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Transitional helper used during the embed feature's first commit
+ * before `buildApp` lifted the `EmbedPreviewService` to a per-app
+ * singleton. The Loop 2 hardening commit refactors this away.
+ *
+ * Production gets a worker-backed source bound to a freshly-allocated
+ * `ConversionResource` (one worker thread per call). Tests opt into
+ * `createInProcessSvgSource()` explicitly.
+ */
+export function defaultSvgSource(): SvgSource {
```
Claudia-Anthropica: @FelixTJDietrich [low] `defaultSvgSource` is exported but unused — its own docstring calls it a "transitional helper used during the embed feature's first commit before `buildApp` lifted the `EmbedPreviewService`". `buildApp` does the lifting now via `getResource`/`getPreviewService`, so the helper is dead. Delete it (and the comment) to avoid future callers wiring up a second `ConversionResource` and silently doubling the worker count.

In `standalone/server/src/services/svg-source.ts` remove the `defaultSvgSource` function (lines ~44-60) along with its docstring. Verify no callers exist (`grep -r defaultSvgSource standalone/`) before deleting.

</details>

standalone/webapp/src/components/modals/EmbedHints.tsx
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * The component reads `diagramId` from the route params; if there is
+ * no diagram in the URL (e.g. the user is on the local-only editor
+ * `/`), the panel renders a one-line hint asking them to save first.
+ */
+export function EmbedHints({ title = "Apollon diagram" }: { title?: string }) {
+  const { diagramId } = useParams<{ diagramId?: string }>()
+  const origin = typeof window !== "undefined" ? window.location.origin : ""
+
+  const snippets = useMemo(() => {
```
Claudia-Anthropica: @FelixTJDietrich [nit] `safeTitle` only strips `]`. An unbalanced `[` in the title (e.g. `Foo [v1`) leaves the markdown alt text technically intact but visually weird in some renderers, and a user with a `]` followed by a `[` survives the strip only because the regex runs once. Cheap defense-in-depth: strip both brackets, or escape them with backslashes which CommonMark honors.

In `standalone/webapp/src/components/modals/EmbedHints.tsx` around line 32, broaden the title sanitizer so neither bracket can sneak into the markdown alt text. Replace `title.replace(/[\]]/g, "")` with `title.replace(/[\[\]]/g, "")` (or escape: `title.replace(/([\[\]])/g, "\\$1")`).

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Solid embed feature — the route shape, ETag/304 path, in-process LRU + coalescing, and the SVG-priority/anti-starvation logic all hang together nicely, and the test coverage is thorough. One real gap inline: the documented 503 `Retry-After` is missing as an actual HTTP header (only lives in the JSON body), which defeats the Camo-backoff intent. A handful of smaller things — schema-regex SemVer fidelity, a tiny race between two Redis reads, dead `defaultSvgSource` export, and a markdown-escape nit — flagged inline.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The shared error handler emits the header and the tests assert it at header level.

standalone/server/src/routes/_schemas.ts
Claudia-Anthropica: @FelixTJDietrich The restrictive version regex is gone from the current code, so this particular concern no longer applies.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich Still seeing the same race in the current `loadOrNotModified()`: `readDiagram()` and `hGet(headRev)` are still separate awaits. Please either derive the ETag/cache key from the diagram body or fetch the diagram and revision consistently so stale SVG cannot be served under a fresh ETag.

standalone/server/src/services/svg-source.ts
Claudia-Anthropica: @FelixTJDietrich Nice, this helper is gone now.

standalone/webapp/src/components/modals/EmbedHints.tsx
Claudia-Anthropica: @FelixTJDietrich Nice, this now strips the markdown delimiters too.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Four threads are addressed, but the ETag/read race is still present in `loadOrNotModified()`; I left that thread open. Requesting changes until that consistency issue is closed.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich This is still open in the latest commit: `readDiagramWithEtag()` reads the diagram body first and `headRev` afterward, so a concurrent PUT can still pair the old body with the new ETag. Please derive the ETag/cache identity from the same data as the body, or fetch both values consistently.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The remaining change request is still the ETag consistency issue in `readDiagramWithEtag()`. Requesting changes until the diagram body and ETag/cache identity are read from a consistent source.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich Still open after the render-cache change: the cache key now uses `etag`, but `etag` still comes from `hGet(headRev)` after `readDiagram()`. A PUT between those reads can cache the old SVG under the new revision tag, so this still needs to derive the key/tag from the read body or fetch both values consistently.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest cache/coalescing work looks good, but the remaining ETag consistency thread is still open. Because the render cache is now keyed by that separately read ETag, the old-body/new-tag race still needs to be fixed before approval.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich Still open on the current head: `readDiagramWithEtag()` still awaits `readDiagram()` and then `hGet(headRev)`, and the render cache key is still built from that separately read ETag. The fix needs to read the body and revision consistently or derive the ETag/cache key from the body.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The remaining ETag consistency thread is still open in `readDiagramWithEtag()`: the body and `headRev` are still read separately, and the render cache uses that separately read tag. Requesting changes until the cache identity and ETag come from a consistent read.

standalone/server/src/routes/embed.ts
Claudia-Anthropica: @FelixTJDietrich Still open on the current head: `readDiagramWithEtag()` still awaits `readDiagram()` and then `hGet(headRev)`, and the render cache key is built from that separately read ETag. A PUT between those reads can still cache the old SVG under the new revision tag, so this needs to derive the tag/key from the body or fetch both values consistently.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The remaining ETag/cache consistency issue is still present in `readDiagramWithEtag()`, so I’m keeping this as requested changes. Once the body and revision are read consistently, this should be ready from my side.

standalone/server/src/routes/embed.ts
FelixTJDietrich: Fixed in 7c14217f. `readDiagramWithEtag` now reads the diagram body and `headRev` in a single `MULTI`/`EXEC` instead of two sequential awaits — EXEC runs atomically with no interleave, so a concurrent PUT can no longer pair an old body with a new `headRev`. Kept the ETag `headRev`-based (monotonic, and matches the strong tag the PUT route mints) rather than switching to `updatedAt`, which would have diverged the embed/PUT cache identity and lost monotonic uniqueness. The two keys (`diagram:{id}` and `diagram:{id}:meta`) share the `{id}` hash tag, so the transaction stays on one slot under Redis Cluster too. Added a regression test: after a PUT the old ETag no longer 304s and resolves to the new tag, proving the body and ETag advance together.

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The single MULTI/EXEC read keeps the body and headRev paired, and the cache key still uses the monotonic headRev.

Claudia-Anthropica (approved):
@FelixTJDietrich All my feedback is addressed now; the ETag/body read is consistent after the MULTI/EXEC change. Approving from my side, with CI still pending when I checked.

standalone/server/src/resources/conversion-resource.ts
```diff
@@ -21,73 +23,150 @@ type WorkerMessage =
...
+    const pw: PooledWorker = {
+      worker: this.createWorker(),
+      busy: false,
+      idleSince: Date.now(),
+      renders: 0,
+      active: undefined,
+    }
+    pw.worker.on("message", (m: WorkerMessage) => this.onMessage(pw, m))
+    pw.worker.on("error", (e) => this.onExit(pw, e))
+    pw.worker.on("exit", (code) => {
```
Claudia-Anthropica, line 162: @FelixTJDietrich [medium] `Worker.terminate()` exits with code `1`, so this `exit` listener sends every intentional recycle/reap, plus the second half of a timeout/error termination, back through `onExit()` as if it were a crash. That increments `consecutiveCrashes` for normal shutdowns and can respawn extra workers while a queue is present, bypassing the pool cap. Mark expected terminations or remove the listener before `terminate()`, and make `onExit()` ignore workers already removed from `this.workers`.

In `standalone/server/src/resources/conversion-resource.ts`, intentional `Worker.terminate()` calls from retire/reap and crash cleanup emit a non-zero `exit`, so the `exit` listener re-enters `onExit()` as if the worker crashed. Fix this by making worker shutdown handling idempotent: ignore workers no longer present in `this.workers`, clear `pw.active` after rejecting, and mark/remove listeners for expected retire/reap terminations so they do not increment crash backoff or respawn extra workers. Add a unit test with a fake worker that emits `exit(1)` after intentional termination.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The embed and cache pieces still look solid overall, but I found one worker-pool lifecycle issue inline. Requesting changes so intentional worker shutdowns and crash handling do not corrupt the pool state under load.

standalone/server/src/resources/conversion-resource.ts
Claudia-Anthropica: @FelixTJDietrich Still open on the current head: the `exit` listener still sends non-zero exits from intentional `terminate()` calls into `onExit()`, and `onExit()` still treats them as crashes without first checking whether the worker was already removed from `this.workers`.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The worker-pool lifecycle issue from the open thread is still present on the current head, so I’m keeping changes requested. Intentional `Worker.terminate()` exits can still flow through crash handling and corrupt pool accounting.

standalone/server/src/resources/conversion-resource.ts
Claudia-Anthropica: @FelixTJDietrich Still open on the current head: retire() and reapIdle() remove the worker and call terminate(), but the later non-zero exit event still reaches onExit() and increments crash accounting because onExit() does not first ignore workers no longer present in the pool. The timeout/error path also leaves the active job attached before calling terminate(), so the follow-up exit can process the same worker a second time.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The worker-pool lifecycle thread is still open on the current head, so I am keeping changes requested. Intentional worker shutdown and crash handling still need to be made idempotent before this is safe under load.

standalone/server/src/resources/conversion-resource.ts
Claudia-Anthropica: @FelixTJDietrich Still open on `bfedacf`: the new commits only touch the webapp embed panel, and `conversion-resource.ts` still routes intentional `terminate()` exits back into `onExit()` without first ignoring workers already removed from the pool.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commits only touch the embed panel, so the open worker-pool lifecycle issue in `conversion-resource.ts` is still unresolved. Keeping changes requested until intentional worker termination is made idempotent and excluded from crash handling.

standalone/server/src/resources/conversion-resource.ts
Claudia-Anthropica: @FelixTJDietrich Still open on `72895b1`: `retire()` and `reapIdle()` remove the worker before calling `terminate()`, but the `exit` listener still routes the later non-zero exit into `onExit()` because `onExit()` does not ignore workers no longer present in `this.workers`. That still treats intentional shutdown as a crash and can double-process timeout/error termination paths.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The remaining worker-pool lifecycle thread is still open on the current head. Intentional worker termination can still flow into crash handling, so I’m keeping changes requested until expected shutdowns are ignored or marked before calling terminate().

standalone/server/src/resources/conversion-resource.ts
FelixTJDietrich: Fixed in this push. `onExit()` now returns early when the worker is no longer in `this.workers`, so a `terminate()`-driven `exit(1)` (from retire/reapIdle, the crash-cleanup itself, or the `exit` half of a real crash's `error`+`exit` pair) is ignored instead of counted — no spurious crash-backoff, no respawn past the cap. `retire()`/`reapIdle()` now drop the worker from the pool before calling `terminate()`, and the active job is cleared after rejection. Added `conversion-resource.lifecycle.test.ts`: a fake worker emits `exit(1)` on `terminate()`, and the test asserts a real crash counts exactly once with one live replacement and that a late stray `exit` spawns nothing.

PR #678, opened 2026-05-09
feat(webapp): version history in local/offline mode

### LINKED ISSUE #670: Version history in local (offline) mode
## User Story
As a student or designer working alone, I want to browse and restore past versions of my diagram so that I can experiment freely without fear of losing a previous state.

## Problem
Version history is currently only available during an active collaboration session. When working offline or alone, there is no way to review, compare, or roll back to an earlier state — one wrong move and the previous structure is gone.

## Proposed Solution
Make version history available in standalone mode. Users can save named checkpoints at any point, browse them as read-only previews in the version sidebar, and restore any checkpoint as the current working diagram.

## Acceptance Criteria
- [ ] A user can save a named version of the current diagram at any time
- [ ] A version history sidebar lists all saved versions with name and timestamp
- [ ] Clicking a version shows it as a read-only preview without altering the working diagram
- [ ] A user can restore any past version, replacing the current diagram with that state
- [ ] Version history persists across browser sessions (survives a page reload)

FelixTJDietrich:
### Summary

Standalone/offline mode (`/local/:id`) had no version history — you could only snapshot a diagram once it was shared through the collaboration server. Someone working alone was one wrong move away from losing earlier state.

This PR closes that gap: local diagrams get the **same** version-history drawer (named saves, preview, restore) as shared diagrams, backed by **IndexedDB** instead of the server — fully offline, no account needed.

It's built as a **port + two adapters**: a `VersionRepository` interface the existing version store and UI already speak, with `RemoteVersionRepository` (the collab server, unchanged) and a new `LocalVersionRepository` (IndexedDB). Switching modes swaps only the adapter — the drawer, the store, and every call site stay identical.

To key versioning cleanly onto `/local/:id` and the `?version=` preview deep-link, it also **migrates the web app's router from react-router to TanStack Router** (type-safe params + per-route code-splitting; URLs and deep links unchanged — an internal refactor, so it carries no changeset of its own). It then **merges current `main`** — React 19 + the React Compiler, the new home page / gallery (#662), and the npm→pnpm migration.

Closes #670.

### Release note

Diagrams you keep on your device — the ones you haven't shared through the server — now have version history. Save named versions as you work, scroll back through them, preview any earlier version (or open it in a new tab to compare side by side), and restore one. Every restore first snapshots whatever is on the canvas, so you can always undo it. Your versions stay on your device, work fully offline, and stay in sync across browser tabs. Shared diagrams also gain a "Save a local copy" action, so you can keep an editable copy on your device before the shared version expires.

### Implementation notes

- **One atomic write path.** Both `create` and the restore auto-snapshot go through a single `commitVersion()` helper that writes a version's metadata and its body **together in one IndexedDB transaction**, assigns the per-diagram sequence number, and evicts to the cap — all before `tx.done`. A history row without a readable body is therefore not constructible.
- **Route-driven repository binding.** Each editor page binds the active adapter in its mount effect, before it fetches; the version store reads the module-level holder from its async actions. (The store lives outside React, so a module holder — not context — is the right tool. Binding in an effect rather than during render avoids a child-effect-before-parent-effect race that could otherwise show an empty drawer on reload.)
- **URL-driven preview.** `?version=<id>` is the single source of truth for which version is previewed — one hook for both the local and collab editors — so reload, Back, and deep links / new tabs all work. Cmd/ctrl/middle-clicking a version row opens its preview in a new tab, matching the gallery cards; every exit affordance leaves preview by clearing the param, never by poking the store behind the URL's back.
- **Restore captures the live canvas.** Restoring while previewing leaves preview mode first, so the "Before restoring …" undo snapshot is the user's actual canvas — not the version they're previewing.
- **Cross-tab + lifecycle.** A `BroadcastChannel` keeps other tabs' drawers in sync, deleting a diagram cascades to its versions, and persistent storage is requested from inside the user gesture so Safari's eviction promise holds.
- **CI.** The bare-runner Playwright jobs (`e2e-tests`, `perf-firefox-tests`) run `playwright install-deps` through a shared bounded+retried composite action, so the dpkg-lock stall on a freshly-booted runner can no longer hang a job.

**Out of scope / follow-ups:** auto-versioning (manual saves only, by design), gzip compression of stored bodies, and a Service Worker for true offline boot are tracked separately.

### Steps for testing

1. `pnpm install && pnpm dev`, then open (or create) a diagram from the gallery — you land on `/local/:id`.
2. Open the **Version history** drawer, draw something, and click **Save first version**. Reload the page → the version is still there (read back from IndexedDB).
3. Click a saved version to **preview** it (read-only banner), then **Restore** → a "Before restoring …" row appears at the top of the history.
4. **Cmd/Ctrl/middle-click** a version row → it opens that version's preview in a new tab; the current tab is unchanged. Plain-click still previews in place, and a single Back returns to the live canvas.
5. Open the same diagram in a second tab → save in one tab → the other tab's drawer refreshes within ~1 s.
6. On a shared diagram (`/shared/:id`), use **Save a local copy** → you land on a fresh `/local/:id` with your own editable copy.

### Screenshots / screencasts

_Screencast pending — the flow to capture is save → reload → preview → restore in local mode (see Steps for testing)._

### Checklist

- [x] Linked to a related issue (#670)
- [x] Added a changeset whose summary is written in the user's voice (`@tumaet/webapp`, minor)
- [x] PR title's Conventional Commit type (`feat`) matches the kind of change — it groups the release note
- [x] Tests added or updated (unit + Playwright e2e against real IndexedDB)
- [ ] Documentation updated (n/a — no user-facing docs change)
- [ ] Screenshots or screencasts attached (screencast pending)


standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -1,54 +1,325 @@
...
+          {banner}
+        </Box>
+        {currentModelId && (
+          <VersionSidebar
+            diagramId={currentModelId}
+            onConfirmedRestore={handleConfirmedRestore}
+            onVersionSaved={handleVersionSaved}
+          />
+        )}
+      </Box>
```
Claudia-Anthropica: @FelixTJDietrich [high] CI is still red — `visual-regression-tests` reports `template-factory.png` 3% off baseline (20588 px, ratio 0.03). Same root cause as the bridge regeneration in 7a729251: the new flex sidebar layout shifts the canvas inside `[data-testid="editor-area"]` on `/`. You regenerated bridge but missed factory, and adapter/command/observer only passed by luck of being under tolerance. Worth regenerating all 5 templates in one go via the documented Playwright Docker procedure so we don't churn through CI runs.

In `standalone/webapp/tests/visual/diagrams.visual.spec.ts-snapshots/`, the `template-factory.png` baseline (and likely the other template baselines too) is stale after the sidebar/flexbox layout change in `standalone/webapp/src/pages/ApollonLocal.tsx`. Regenerate `template-adapter.png`, `template-bridge.png`, `template-command.png`, `template-factory.png`, and `template-observer.png` by running the visual-regression tests inside the standard Playwright Docker container (per the procedure in `.github/workflows/pr-health-checks.yaml`) with `--update-snapshots`. Commit only the updated PNGs.

</details>

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
```diff
@@ -0,0 +1,414 @@
...
+    description: "",
+    createdAt,
+    kind: "auto",
+    librarySchemaVersion: body?.version ?? "4.0.0",
+    seq,
+  }
+  await tx.objectStore("versions").add(row)
+  if (body) {
+    const encoded = encodeBody(body)
+    await tx.objectStore("versionBodies").put({ diagramId, id, ...encoded })
```
Claudia-Anthropica: @FelixTJDietrich [medium] If a caller ever passes `currentBody: undefined` (the type allows it — `restore`'s opts have `currentBody?: UMLModel`), this writes the `versions` meta row but skips the `versionBodies` row. The auto-snapshot then shows up in the drawer, the user clicks it, and `getBody()` throws `ApiError(404)` — looks like data loss to them. Either refuse to write the meta when `body` is undefined, or invariant-it at the public `restore` boundary so the optional type narrows away here.

In `standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts` around line 393, `createAutoSnapshot` conditionally writes the body only when `body` is defined, leaving a meta row that 404s on `getBody`. Tighten the type so `body` is required (change `body: UMLModel | undefined` to `body: UMLModel` on `createAutoSnapshot` args, and require `currentBody` at the `restore`/`RestoreVersionResult` boundary by validating it before the `createAutoSnapshot` call and throwing a clear error if missing). Update the `restore` test suite to cover the missing-body case as a rejected promise.

</details>

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
```diff
@@ -0,0 +1,414 @@
...
+      diagramId,
+      body: opts.currentBody,
+      restoredFromName: labelTarget,
+    })
+    broadcastInvalidate(diagramId)
+    return {
+      autoSnapshotVersionId: autoId,
+      updatedAt: nowIso(),
+      headRev: undefined,
+    }
```
Claudia-Anthropica, line 261: @FelixTJDietrich [low] `restore` does a `db.get("versions", versionId)` in an implicit read tx, then opens a fresh write tx inside `createAutoSnapshot`. A peer tab deleting that row between the two will let the auto-snapshot write succeed with a stale `labelTarget` (`v${row.seq}`). Not data corruption, just a misleading label. `create` already inlines everything under one `readwrite` tx — same pattern works here.

In `standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts` lines 270–289, `restore` reads `row` outside the write transaction `createAutoSnapshot` opens. Refactor so the existence check and the auto-snapshot write happen in the same `db.transaction(["versions", "versionBodies", "diagramMeta"], "readwrite")` block, mirroring the pattern in `create`. Move the label derivation inside the transaction after the `get` returns.

</details>

standalone/webapp/src/services/versionRepository/idb.ts
```diff
@@ -0,0 +1,124 @@
...
+        )
+      },
+    }).catch((err) => {
+      // Reset so the next caller gets a fresh attempt instead of a
+      // cached rejection. A locked-down browser / file:// origin lands
+      // here; the repository falls back to an in-memory shim.
+      dbPromise = null
+      log.error("Failed to open apollon-versions IDB", err as Error)
+      throw err
+    })
```
Claudia-Anthropica, line 118: @FelixTJDietrich [nit] Comment claims "the repository falls back to an in-memory shim" on open failure, but no shim exists in `LocalVersionRepository` — `getDb()` rethrows and every public method propagates the error. Either drop the second sentence of the comment or implement the shim.

In `standalone/webapp/src/services/versionRepository/idb.ts` around line 108–113, remove the misleading sentence about the repository falling back to an in-memory shim (no such fallback is implemented in `LocalVersionRepository`). Keep the rest of the comment.

</details>

standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -1,54 +1,325 @@
...
+          .getState()
+          .restoreVersion(currentModelId, versionId, editor.model)
+        editor.model = importDiagram(body) as UMLModel
+        editor.fitView()
+        if (preview) exitPreview()
+        if (summary) {
+          const label =
+            summary.description.trim() ||
+            summary.name.trim() ||
+            `v${summary.seq ?? ""}`
```
Claudia-Anthropica: @FelixTJDietrich [nit] `\`v${summary.seq ?? ""}\`` produces the bare string `"v"` when `seq` is undefined. Use a different fallback (e.g. `summary.id.slice(0, 8)` or `"this version"`) so the toast doesn't read "Restored 'v'.".

In `standalone/webapp/src/pages/ApollonLocal.tsx` line 192, the label fallback `` `v${summary.seq ?? ""}` `` becomes `"v"` when `seq` is undefined. Replace it with a more meaningful fallback such as `summary.id.slice(0, 8)` or a literal like `"this version"`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Solid architecture overall — the port/adapter split is clean, the eviction parity test pins the drawer↔eviction contract nicely, and the e2e covers the right journeys. Two blockers: visual-regression is still failing on `template-factory.png` (same root cause as the bridge regeneration, just missed), and `createAutoSnapshot` can write a meta row without a body if a caller ever passes `currentBody: undefined`. A couple of smaller things inline.

FelixTJDietrich:
## Merged `main` and resolved conflicts

Brought this branch up to date with `main`, which had moved substantially — most importantly the **new home page (#662)** and the **npm→pnpm migration**. Conflicts were resolved meaningfully (not just textually) so local version history works correctly under the new architecture.

### What the new routing changed for local versioning
`#662` re-architected standalone routing:
- `/` → diagram **gallery** (was the local editor)
- `/local/:id` → local editor (now every local diagram has a path id)
- `/shared/:id` → collab editor
- legacy `/:id?view=…` → redirect to `/shared/:id`

This branch had assumed `/` was the editor and keyed version history off `usePersistenceModelStore.currentModelId`. The merge **rekeys everything to the `/local/:id` path param** (via `useDiagramIdFromPath`/`useParams`). Concretely:
- `ApollonLocal` keeps `main`'s path-id lookup, `ErrorPage`-when-missing, and the debounced **thumbnail export** the gallery cards consume — but the export *and* `updateModel` are now **guarded while a version preview is active**, so a read-only snapshot is never written back or captured as the thumbnail. The old auto-create-a-diagram branch is dropped (creation happens from the gallery now).
- `VersionHistoryButton` gate relaxed to show on `/local/:id` *and* `/shared/:id`. Dead `useActiveDiagramId` hook deleted.
- `SaveLocalCopyButton` now navigates to `/local/<newId>` (not `/`) and only renders on `/shared/:id`.
- Adopted the `@tumaet/apollon/react` import subpath; dropped `@/utils/collaboration` (moved into the library on `main`).
- pnpm: added `idb` + `fake-indexeddb` to `@tumaet/webapp`, deleted `package-lock.json`, regenerated `pnpm-lock.yaml`, added a changeset.

### 🐛 Bug fixed (surfaced by the new routing)
The active `VersionRepository` was bound in a `useEffect`. React flushes **child effects before parent effects**, so on a **reload with the drawer already open** (`drawerOpenByDiagram` is persisted), the drawer's first `fetchVersions` ran against the default `RemoteVersionRepository` → **empty history**. Fixed by binding the repository **synchronously during render** in both `ApollonLocal` and `ApollonWithConnection`. Covered by a new e2e test (save → reload → history still present).

### Verification
- `pnpm run build` (lib + webapp, incl. `tsc`) ✅ · webapp `eslint` 0 errors ✅ · `prettier --check` ✅
- Webapp unit tests: 200/200 ✅ (versioning suites incl. `LocalVersionRepository`, eviction, bootstrap, `SaveLocalCopyButton`)
- **Rewrote `local-version-history.spec.ts` for `/local/:id`** (7 e2e tests) — verified against real Chromium incl. **IndexedDB persistence across reload**, preview/exit, gallery→editor open, and `/local/:unknown` not-found. Stable over 3× repeat (21/21).
- `home-page.spec` / `version-history.spec` pass under CI's `retries: 2` (a few pre-existing gallery-hydration flakes recover on retry; none in code this PR touches).

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich I still don't see committed visual baseline updates, and the current visual-regression check is still pending, so I can't mark this addressed yet. Leaving the thread open until that check passes or the required snapshots are updated.

standalone/webapp/src/services/versionRepository/idb.ts
Claudia-Anthropica: @FelixTJDietrich Still unresolved: the comment still says there is an in-memory shim, but `getDb()` rethrows and there is no fallback in the repository.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich This transaction split is still unchanged: `restore` reads the row first, then `createAutoSnapshot` opens a separate write transaction. Still leaving this open as low-priority follow-up.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich This is still present: `createAutoSnapshot` still accepts `UMLModel | undefined` and skips the `versionBodies` write when `opts.currentBody` is missing, so it can still create a history row that 404s on preview/restore.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich This fallback is still `v${summary.seq ?? ""}`, so an undefined `seq` still produces the bare `v` toast label.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous blockers are still open: the local auto-snapshot path can still create metadata without a body, and the visual-regression follow-up is not proven fixed yet. I also left the lower-priority existing threads open.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. `currentBody` is required now and the shared commit path writes the metadata and body together.

standalone/webapp/src/services/versionRepository/idb.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The comment now matches the actual `getDb()` behavior.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The fallback now uses `this version` when `seq` is absent.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich I still do not see template snapshot updates on this head, and the visual-regression job is still pending, so I am keeping this open until CI proves the baselines are okay.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich This one is still open after the refactor: `restore` still reads the target row first, then `createAutoSnapshot` opens the write transaction through `commitVersion`. Keeping it unresolved as the same low-priority follow-up.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Thanks, the data-integrity and smaller cleanup threads are fixed now. I’m still requesting changes because the visual-regression baseline thread remains unresolved while the job is pending, and the local restore transaction split is still open.

FelixTJDietrich:
### Review threads — all addressed

Thanks for the thorough review. Update + fixes pushed:

**Blockers**
- **`createAutoSnapshot` could write metadata without a body** → fixed at the type and the storage layer. `restore`'s `currentBody` is now **required** (`types.ts`), and both `create` and the restore auto-snapshot route through one `commitVersion()` helper that writes the meta row **and** its body unconditionally in a single transaction. A history row that 404s on preview/restore is no longer constructible. Added a test asserting *every* listed row has a readable body after create + restore + create.
- **Visual regression (`template-factory.png`, `template-bridge.png`)** → both baselines regenerated in the CI Playwright image (`mcr.microsoft.com/playwright:v1.59.1-noble`) and verified clean (full suite 18/18, CI mode). The local sidebar shifts ReactFlow's `fitView` a few pixels on `/local/:id`; these were the affected canvases.

**Smaller threads**
- **`idb.ts` "in-memory shim" comment** — removed; there is no shim, the open failure rethrows and surfaces as a failed list/create. The comment now says exactly that.
- **`restore` read-then-write across transactions** — the label read feeds only display copy and the auto-snapshot's own seq/eviction stays atomic inside `commitVersion`'s single transaction, so there's no lost-update window; left as-is intentionally.
- **`v${seq ?? ""}` toast** — guarded; falls back to "this version" when `seq` is absent instead of a bare "v".

**Also in this push**
- Merged latest `main` (new home page #662 + npm→pnpm) and re-keyed local versioning onto `/local/:id`.
- Moved the repository binding out of render into the page mount effect (it was racing the drawer's first fetch on reload-with-drawer-open).
- Dropped the speculative `compression`/gzip field and the unused `persistedRequested` flag; made `persist()` gesture-correct.
- Fixed the **"Save first version"** empty-state button — it now actually saves the first version (it previously only focused the composer), disabled with a tooltip when there's nothing to save.

PR description rewritten to follow the template. Screencast still to come before merge.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The visual baselines are committed and the visual-regression check is passing on this head.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich This is still unchanged on the current head: `restore` reads the target version before the write transaction opened by `createAutoSnapshot`, so I am leaving this thread unresolved.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Visuals are passing now, but I am keeping changes requested for the remaining `restore` transaction thread. The latest `Save first version` fix looks fine from the incremental diff.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
FelixTJDietrich: Fixed in 2298ab4a. `restore` now reads the target row **and** writes the pre-restore auto-snapshot in a single read-write transaction:

```ts
const tx = db.transaction(["versions", "versionBodies", "diagramMeta"], "readwrite")
const target = await tx.objectStore("versions").get(versionId)
if (!target) throw new ApiError(404, ...)
const label = target.description.trim() || target.name.trim() || `v${target.seq}`
const { row } = await writeVersionInTx(tx, { ...meta: { name: `Before restoring ${label}` } })
await tx.done
```

The co-write was split into `writeVersionInTx` (runs inside a caller-supplied tx) and a thin `commitVersion` (opens + commits one). `create` uses `commitVersion`; `restore` reuses `writeVersionInTx` in its own transaction, so the label read and the snapshot write are now atomic — no window for a peer-tab delete to produce a stale label. The separately-exported `createAutoSnapshot` is gone (it had no other callers).

Claudia-Anthropica: @FelixTJDietrich Nice, that handles it. The target read and pre-restore snapshot now share the same transaction.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed now. The remaining restore transaction thread is fixed and resolved on the current head.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
```diff
@@ -0,0 +1,403 @@
...
+    let autoSnapshotVersionId: string
+    try {
+      // Read the target row and write the pre-restore auto-snapshot in ONE
+      // transaction, so a peer tab deleting the target between the two can't
+      // leave a stale "Before restoring …" label.
+      const tx = db.transaction(
+        ["versions", "versionBodies", "diagramMeta"],
+        "readwrite"
+      )
+      const target = await tx.objectStore("versions").get(versionId)
```
Claudia-Anthropica, line 231: @FelixTJDietrich [high] `restore()` trusts `versionId` globally and never verifies that the row belongs to the `diagramId` being restored. A stale preview from another local diagram can call `restore(B, versionFromA, ...)`, which writes the auto-snapshot under B and then lets the page apply A's body to B. Check `target.diagramId === diagramId` before deriving the label or writing the snapshot, and cover the two-diagram case in the repository tests.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts`, `restore()` reads a version by id and does not reject rows whose `diagramId` differs from the requested diagram. Fix this by checking `target.diagramId === diagramId` immediately after the lookup and throwing the same 404 when it does not match. Add a unit test that creates versions for diagram A and B and verifies restoring A's version through B is rejected and does not create a B auto-snapshot.\n\n</details>

standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -98,16 +177,222 @@ export const ApollonLocal: React.FC = () => {
...
+    setEditor,
+    setThumbnail,
+    updateModel,
+    fetchVersions,
+  ])
+
+  // -------- Preview overlay ---------------------------------------------
+  useEffect(() => {
+    if (!editor) return
+    if (preview) {
```
Claudia-Anthropica, line 247: @FelixTJDietrich [medium] The preview store only carries `versionId` and `body`, so this page applies whatever preview was left by the previous diagram. Repro: preview a version on `/local/a`, navigate back to the gallery, then open `/local/b`; B opens in read-only preview with A's body and the Restore button still targets the stale id. Scope preview state to the diagram or clear/ignore stale previews on route changes.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/pages/ApollonLocal.tsx`, the preview effect applies any global `useVersionStore.preview` to the current local editor. Fix this by scoping preview state to its diagram id: add `diagramId` to `PreviewState` when entering preview, update banner/page checks to require a matching diagram, and clear or ignore stale previews on diagram route changes. Add an e2e or component test that previews a local version, navigates to another local diagram, and confirms the second diagram does not show the first preview.\n\n</details>

standalone/webapp/src/stores/versionStoreBootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+  let prevModelIds = new Set(
+    Object.keys(usePersistenceModelStore.getState().models)
+  )
+  const unsubscribeModels = usePersistenceModelStore.subscribe(
+    (state, prev) => {
+      if (state.models === prev.models) return
+      const next = new Set(Object.keys(state.models))
+      for (const id of prevModelIds) {
+        if (!next.has(id)) {
+          getVersionRepository()
```
Claudia-Anthropica: @FelixTJDietrich [medium] This cascade depends on whichever repository was last bound. On a fresh gallery load no editor calls `ensureVersionStoreBootstrapped()`, and after visiting a shared diagram the active repository is remote with no `purgeDiagram`, so deleting a local diagram from the gallery can leave its IDB versions behind. Install the bootstrap from an app-level path and purge through the local adapter directly instead of the mutable active repository.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/stores/versionStoreBootstrap.ts`, local diagram deletion uses `getVersionRepository().purgeDiagram?.`, so cleanup is skipped when the active adapter is remote or when the bootstrap was never installed before a gallery delete. Fix this by installing the bootstrap from an app-level component that is mounted on the gallery and by purging through `LocalVersionRepository.purgeDiagram` for deleted local model ids. Update the bootstrap test to cover deletion after `setVersionRepository(RemoteVersionRepository)` and, if feasible, a gallery-delete path.\n\n</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The adapter split is in decent shape, but I found a cross-diagram restore path that can apply one local diagram's preview/version to another, plus a cleanup gap for deleted local diagrams. CI is also red in e2e right now (`edge-handle-zoom` sees the hidden export React Flow root), so I’m requesting changes.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich This is still open on the current head: `restore()` still only checks `!target` after `versions.get(versionId)`, but it never rejects a row whose `target.diagramId` differs from `diagramId`. Please add that ownership check before deriving the label or writing the auto-snapshot.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich The URL sync does not fully close this yet. If you leave `/local/a?version=...` through the gallery/navbar and then open `/local/b` without a `version` param, `useVersionPreviewUrlSync` mounts with `prevFromUrl.current === null`, so it does not call `exitPreview()` and the global `{versionId, body}` preview is still applied to B. The preview state still needs to be scoped to the diagram or cleared on route changes.

standalone/webapp/src/stores/versionStoreBootstrap.ts
Claudia-Anthropica: @FelixTJDietrich This one is still open too. `ensureVersionStoreBootstrapped()` is only called from the editor pages, and the cascade still calls `getVersionRepository().purgeDiagram?.(id)`, so a fresh gallery delete still has no subscription and a delete after a shared editor visit can still point at the remote adapter. The cross-window not-found guard is useful, but it does not address the IDB cleanup path from the original thread.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The new URL preview work is useful, but the three change-request threads from the last review are still open: cross-diagram `restore()` still needs the ownership check, stale preview state can still leak when leaving preview via navigation, and gallery deletion still depends on an editor-mounted bootstrap plus the currently bound repository. CI is still pending, but these are enough to keep changes requested.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich Still open. `useVersionPreviewUrlSync` now clears stale preview state in a passive effect, but the local page preview effect is registered later and still runs once with the old global `preview` value, applying the previous diagram body before the clearing rerender; the non-preview branch then only toggles preview mode/read-only off and does not restore the current diagram model. The preview state still needs a diagram-id guard, or the local preview application must ignore previews that do not belong to the current route.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich Still open on `be0c246`: `restore()` only checks `!target` after `versions.get(versionId)`, so a version row from another diagram can still be restored through the current diagram. Please reject `target.diagramId !== diagramId` before deriving the label or writing the auto-snapshot.

standalone/webapp/src/stores/versionStoreBootstrap.ts
Claudia-Anthropica: @FelixTJDietrich Still open on this head too. `ensureVersionStoreBootstrapped()` is still only installed from editor pages, and the delete cascade still calls `getVersionRepository().purgeDiagram?.(id)`, so a fresh gallery delete has no subscriber and a delete after a shared editor visit can still target the remote adapter instead of purging local IDB versions.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The three change-request threads are still open on this head: cross-diagram local restore ownership, stale preview leakage between local diagrams, and gallery deletion cleanup still depending on the active repository. CI is still pending, but these are enough to keep changes requested.

standalone/webapp/src/stores/versionStoreBootstrap.ts
Claudia-Anthropica: @FelixTJDietrich Still open on d7856497. The bootstrap is still only installed from editor pages and the cascade still uses `getVersionRepository().purgeDiagram?.(id)`, so a gallery delete can miss the subscription entirely or run after a shared route has bound the remote adapter.

standalone/webapp/src/pages/ApollonLocal.tsx
Claudia-Anthropica: @FelixTJDietrich Still open on d7856497. The preview store is still `{ versionId, body }` without `diagramId`, and the local preview effect still applies any non-null preview before the URL-sync effect clears it on a route change, so stale preview bodies can still flash/apply to the next local diagram.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
Claudia-Anthropica: @FelixTJDietrich Still open on d7856497. `restore()` still only checks `!target`; because it does not reject `target.diagramId !== diagramId`, a version row from another local diagram can still be restored through the current diagram before the page applies that body.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The three change-request threads are still open on this head: cross-diagram local restore ownership, stale preview state between local diagrams, and local version cleanup on gallery delete. CI is still pending, but these issues are enough to keep changes requested.

standalone/webapp/src/services/versionRepository/LocalVersionRepository.ts
FelixTJDietrich: Fixed in b4803eb2. `restore()` now verifies the row belongs to the diagram (`!target || target.diagramId !== diagramId` → 404) before deriving the label or writing the auto-snapshot, mirroring the existing `editInfo`/`getBody` guard. Added a repository test asserting a cross-diagram `versionId` is rejected and leaves no auto-snapshot row under the wrong diagram.

standalone/webapp/src/pages/ApollonLocal.tsx
FelixTJDietrich: Fixed in b4803eb2. `PreviewState` now carries `diagramId` (set in `enterPreview`), and both editor pages select the preview only when `s.preview?.diagramId === diagramId`, so a preview left by a previous diagram is ignored rather than applied. The cross-tab handler in `versionStoreBootstrap` likewise only reacts to its own diagram. (The `?version=` URL-as-source-of-truth already clears the store preview on the A→gallery→B navigation — this scoping additionally closes the brief render window before that effect runs, and is exercised by the version-store/bootstrap tests.)

standalone/webapp/src/stores/versionStoreBootstrap.ts
FelixTJDietrich: Fixed in b4803eb2. The cascade now purges through `LocalVersionRepository.purgeDiagram(id)` directly — a deleted persistence-store model is always a local diagram, so its versions live in IndexedDB regardless of which adapter the editor last bound (the remote adapter has no `purgeDiagram` and would silently orphan them). `ensureVersionStoreBootstrapped()` now installs app-wide from `__root` (removed from the editor pages), so a gallery-only session that never opens an editor still wires the cascade. Added a test that binds the remote adapter before `deleteModel` and asserts the local IDB versions are still purged.

standalone/webapp/src/components/versioning/CurrentVersionRow.tsx
```diff
@@ -44,10 +49,11 @@ interface Props {
  * last save" is accurate; "Unsaved" would be misleading.
  */
 export const CurrentVersionRow: FC<Props> = ({
+  diagramId,
   hasChanges,
   latestSavedVersion,
 }) => {
-  const previewState = useVersionStore((s) => s.preview)
+  const previewState = useVersionStore((s) => selectScopedPreview(s, diagramId))
   const exitPreview = useVersionStore((s) => s.exitPreview)
```
Claudia-Anthropica: @FelixTJDietrich [medium] This still exits preview through the store only, but preview is now driven by the URL. When the user clicks "Return to current", `exitPreview()` clears the store while `?version=...` remains, so `useVersionPreviewUrlSync` immediately calls `enterPreview()` again and the row cannot actually leave preview. Wire this row through the same `closePreview()` URL path that the banner uses, and add a regression for clicking "Return to current".

In `standalone/webapp/src/components/versioning/CurrentVersionRow.tsx`, the "Return to current" row calls `useVersionStore.exitPreview()` while the URL still contains `?version=`. Fix this by passing a URL-level `closePreview`/`onExitPreview` handler from the editor pages through `VersionDrawer`/`VersionSidebar` into `CurrentVersionRow`, and use it instead of directly clearing the store when preview is URL-driven. Add a regression test for clicking the drawer row to exit preview.

</details>

standalone/webapp/src/components/versioning/DeleteVersionModal.tsx
```diff
@@ -17,7 +21,7 @@ export const DeleteVersionModal = ({ diagramId, versionId }: Props) => {
   const deleteVersion = useVersionStore((s) => s.deleteVersion)
   const exitPreview = useVersionStore((s) => s.exitPreview)
   const previewingThis = useVersionStore(
-    (s) => s.preview?.versionId === versionId
+    (s) => selectScopedPreview(s, diagramId)?.versionId === versionId
```
Claudia-Anthropica, line 27: @FelixTJDietrich [low] Same URL/state mismatch when deleting the version currently being previewed: `exitPreview()` clears the store, but `?version=<deleted id>` stays in the URL. The sync effect will try to re-enter that deleted preview and then show the "version unavailable" toast while stripping the param, so a successful delete looks like a preview error. Clear the preview URL before or with the delete instead of only calling the store action.

In `standalone/webapp/src/components/versioning/DeleteVersionModal.tsx`, deleting the currently previewed version clears only `useVersionStore.preview` and leaves `?version=<id>` in the URL. Fix this by closing preview through the URL-level `closePreview` path before deleting, or by otherwise removing the `version` search param when `previewingThis` is true so the URL-sync effect does not re-enter the deleted version. Add a regression test for deleting the active preview.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The earlier local repository blockers look addressed and CI is green. I found one remaining URL-driven preview regression, plus a related delete-preview cleanup issue, both inline.

standalone/webapp/src/components/versioning/CurrentVersionRow.tsx
FelixTJDietrich: Fixed in de8b0dae. "Return to current" now calls a shared `useClosePreview()` hook that strips `?version=` from the URL instead of clearing the store directly, so the URL→store sync (`useVersionPreviewUrlSync`) actually leaves preview. Regression covered in `previewExit.test.tsx`.

standalone/webapp/src/components/versioning/DeleteVersionModal.tsx
FelixTJDietrich: Fixed in de8b0dae. Deleting the previewed version now clears `?version=` via the shared `useClosePreview()` hook before the delete, so the URL→store sync no longer tries to re-enter the just-deleted id (no more spurious "version unavailable" toast on a successful delete). Regression covered in `previewExit.test.tsx`.

standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -114,16 +231,234 @@ export const ApollonLocal: React.FC = () => {
...
+   */
+  const performRestore = useCallback(
+    async (versionId: string) => {
+      if (!editor || !diagramId) return
+      const summary = versions.find((v) => v.id === versionId)
+      try {
+        const body = await resolveBody(versionId)
+        await useVersionStore
+          .getState()
+          .restoreVersion(diagramId, versionId, editor.model)
```
Claudia-Anthropica: @FelixTJDietrich [high] When this runs from a preview, `editor.model` is already the preview overlay, not the live local canvas. The local repository writes `currentBody` into the durable "Before restoring ..." row, so restoring v1 from a dirty v2 canvas stores v1 as the undo snapshot and the advertised undo path just restores v1 again. Capture the live model before applying preview, or temporarily leave preview mode before reading the body, and pass that live body to `restoreVersion()`.

In `standalone/webapp/src/pages/ApollonLocal.tsx`, `performRestore()` passes `editor.model` to `restoreVersion()` while preview mode makes it equal the preview body. Fix this by passing the live pre-preview model as `currentBody` when `preview` is active, either by storing a cloned live model before overlaying the preview or by exiting preview before reading the model. Add a regression that restoring from preview creates a `Before restoring ...` snapshot whose body matches the pre-preview canvas, not the restored version.

</details>

standalone/webapp/src/pages/ApollonWithConnection.tsx
```diff
@@ -451,24 +442,13 @@ export const ApollonWithConnection: React.FC = () => {
           editor.model
```
Claudia-Anthropica: @FelixTJDietrich [high] Same restore-from-preview problem on the shared path: while the banner is visible, `editor.model` is the read-only preview body, and the server persists that `currentBody` as the undo snapshot before applying the restore. The undo snackbar then points at a snapshot of the version we just restored, so Undo cannot bring back the live canvas. The shared drawer's fallback restore path has the same `editor.model` read and also needs to go through the URL-closing preview restore helper.

In `standalone/webapp/src/pages/ApollonWithConnection.tsx`, `handleRestoreFromPreview()` passes the preview overlay as `currentBody`, so the server's auto-snapshot is wrong. Fix this by reading the live diagram body before calling `restoreVersion()` when preview is active, then closing `?version=` through `closePreview()`. Route shared drawer restores in `standalone/webapp/src/components/versioning/VersionDrawer.tsx` through the same page-level helper so row-menu restores also use the live body and leave preview. Add a regression for undo after restoring from a preview.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The version-history plumbing is close, but restore-from-preview is currently snapshotting the preview overlay instead of the live canvas in both local and shared modes. That breaks the promised undo path for restores, so I’m requesting changes.

.github/workflows/pr-health-checks.yml
```diff
@@ -225,46 +225,46 @@ jobs:
...
-        if: steps.playwright-cache.outputs.cache-hit != 'true'
-        run: pnpm --filter @tumaet/webapp exec playwright install firefox
-
-      - name: Install Playwright system deps
-        run: pnpm --filter @tumaet/webapp exec playwright install-deps firefox
-
+      # Firefox is already present in the container image — no install /
+      # install-deps step, so there is no apt path left to hang on.
       - name: Run Firefox performance budget tests
         run: pnpm --filter @tumaet/webapp run test:perf:firefox
```
Claudia-Anthropica, line 280: @FelixTJDietrich [high] This new containerized Firefox step is failing because GitHub runs the container as root while `HOME` is still `/github/home`, which is owned by `pwuser`; Playwright's log says Firefox exits before the test starts. Set `HOME=/root` for this step or job so Firefox can launch inside the Playwright container.

In `.github/workflows/pr-health-checks.yml`, `perf-firefox-tests` runs Firefox inside the Playwright container with `HOME=/github/home`, so Firefox exits because that directory is not owned by root. Fix this by setting `HOME: /root` in the Firefox perf test step or the job-level environment, then keep the existing `pnpm --filter @tumaet/webapp run test:perf:firefox` command.

</details>

standalone/webapp/src/services/versionRepository/__tests__/LocalVersionRepository.test.ts
```diff
@@ -0,0 +1,429 @@
...
+  })
+
+  it("subscribeToLocalVersionEvents returns a working unsubscribe", async () => {
+    const handler = vi.fn()
+    const unsubscribe = subscribeToLocalVersionEvents(handler)
+    // Use a sibling channel to drive the event — subscriber lives on
+    // the module's own channel singleton.
+    const sender = new BroadcastChannel("apollon-versions")
+    sender.postMessage({ type: "invalidate", diagramId: DIAGRAM_ID })
+    await new Promise((r) => setTimeout(r, 0))
```
Claudia-Anthropica: @FelixTJDietrich [medium] CI just failed here with `Number of calls: 0`: one `setTimeout(0)` is not a reliable delivery barrier for `BroadcastChannel` in the CI runner. Make this wait for the actual handler call, like the previous broadcast test already does with a promise resolved by the message listener, then assert unsubscribe by racing a short timeout after the second post.

In `standalone/webapp/src/services/versionRepository/__tests__/LocalVersionRepository.test.ts`, the unsubscribe test waits one `setTimeout(0)` after `BroadcastChannel.postMessage()`, which is flaky and failed in CI before the handler ran. Fix this by awaiting a promise that resolves from the subscribed handler for the first message, then unsubscribe and assert no second call after a bounded timeout.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest CI run is red in two PR-owned places: the Firefox container job cannot launch Firefox with the current HOME, and the new BroadcastChannel unsubscribe test races message delivery. I left both inline.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The existing high-priority restore-from-preview threads still apply on this head, so I’m keeping changes requested. CI is green now and I did not find any new distinct issues, so I’m not adding duplicate inline comments.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The existing restore-from-preview blockers are still present on this head, and the BroadcastChannel unsubscribe test still has the same `setTimeout(0)` delivery race. The new CI change removes the Firefox HOME issue, but these PR-owned blockers are still unresolved, so I am keeping changes requested.

standalone/webapp/src/pages/ApollonLocal.tsx
FelixTJDietrich: Fixed in 2a89e148. `performRestore` now leaves preview mode first (`editor.setPreviewMode(false)` resyncs `editor.model` from the live Yjs doc — the same resync the exit path uses), captures that live body, and passes it to `restoreVersion()` as the "Before restoring …" undo snapshot. The overlay is no longer what gets snapshotted.

standalone/webapp/src/pages/ApollonWithConnection.tsx
FelixTJDietrich: Fixed in 2a89e148. The banner and drawer restore paths are now one `handleRestore`: when previewing it `setPreviewMode(false)` to resync the live canvas, captures that as the undo body for `restoreVersion()`, then `closePreview()`. The shared drawer now routes through it via `onConfirmedRestore`, so the fallback path no longer reads the overlay either.

.github/workflows/pr-health-checks.yml
FelixTJDietrich: Resolved by reverting the container approach (309d1db4). Your HOME diagnosis is plausible, but rather than chase Firefox-in-GHA-container quirks I moved perf-firefox back to the bare runner, where Firefox launches reliably, and fixed the actual original cause — `playwright install-deps` blocking on the boot-time unattended-upgrades dpkg lock — by bounding+retrying that step (now a shared composite action, b6675baf). The containerized Firefox step this thread refers to no longer exists; perf-firefox is green.

standalone/webapp/src/services/versionRepository/__tests__/LocalVersionRepository.test.ts
FelixTJDietrich: Fixed in 2a89e148. The test no longer relies on a `setTimeout(0)` barrier: the first assertion awaits actual delivery via a promise the handler resolves, and the unsubscribe assertion uses a fresh sibling-channel as a delivery barrier (await its receipt, then assert the unsubscribed handler stayed silent) instead of a fixed tick. Ran it 3x locally — stable.

standalone/webapp/src/pages/ApollonLocal.tsx
```diff
@@ -102,6 +218,7 @@ export const ApollonLocal: React.FC = () => {
       }
 
       log.debug("Cleaning up Apollon instance")
+      instance.unsubscribe(subId)
       instance.destroy()
       const isTransitioningToAnotherLocalDiagram = /^\/local\//.test(
```
Claudia-Anthropica, line 221: @FelixTJDietrich [medium] This cleanup is still reading `locationRef.current` from the last rendered local route. On an unmount to `/shared/...` or `/`, that ref can still point at `/local/...`, so the destroyed local editor stays in `EditorContext` and `currentModelId` remains set until another page happens to clear it. Clear the editor/current local id when leaving local mode, and only preserve it for local-to-local transitions using a reliable next-path signal.

In `standalone/webapp/src/pages/ApollonLocal.tsx`, the editor cleanup decides whether to clear `EditorContext` and `currentModelId` from `locationRef.current`, which still contains the previous `/local/...` pathname during unmount/navigation away. Fix this by clearing the editor/current local id when leaving the local route, and only preserving it for local-to-local transitions using a reliable next-location signal or by setting the new local id from the next page mount.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous version-history blockers look addressed, but I found one stale local-editor cleanup path that can leak the destroyed editor into the next route. Requesting changes for that inline.

standalone/webapp/src/pages/ApollonLocal.tsx
FelixTJDietrich: Fixed in f8bec511. The cleanup no longer reads the lagging `useLocation` ref — it reads `router.state.location.pathname`, which TanStack Router updates synchronously on navigation and is therefore the real destination even while this component unmounts. Only a hop to another `/local/*` diagram preserves the editor + `currentModelId` (for the next instance to adopt); leaving local mode to `/shared/...` or the gallery now clears both, so a destroyed editor can't linger in `EditorContext`. Removed the now-unused location ref and its updater effect.

standalone/webapp/src/stores/useVersionStore.ts
```diff
@@ -454,7 +475,7 @@ export const useVersionStore = create<VersionStore>()(
           try {
             const actor =
               sessionStorage.getItem("apollon-collab-name") || undefined
-            await VersionApiClient.restore(
+            await getVersionRepository().restore(
```
Claudia-Anthropica, line 478: @FelixTJDietrich [medium] `triggerUndoRestore()` still clears preview only in the store, and the toast passes `editor.model` before this action runs. If the undo toast is clicked while `?version=...` preview is open, `currentBody` is the preview overlay and the URL sync can re-enter that preview right after the undo. Route this through the same URL-aware preview close path and capture the live body after `setPreviewMode(false)` before calling the store action.

In `standalone/webapp/src/stores/useVersionStore.ts`, undo restore exits preview with `exitPreview()` while URL-driven preview still has `?version=...`, and the caller captures `editor.model` before leaving preview. Fix undo restore so the UI/page closes preview via the URL-aware close path, flips the editor out of preview mode, captures the live model, then calls the store action with that live body; avoid store-only preview exits for URL-driven preview.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous blockers look addressed and CI is green, but the undo restore path still has one URL-driven preview edge case. See the inline comment; undo should leave preview the same way the banner/drawer paths do before it snapshots the canvas.

PR #680, opened 2026-05-10
fix: helper lines for nested elements

### LINKED ISSUE #552: Helper lines are incorrect in nested elements
### Describe the bug

Helper line is positioned incorrectly in Class Diagram when using a package, it is also breaking for Components diagram.

### To Reproduce

1. Create a Class Diagram
2. Drag a package
3. Drag 2 class nodes.
4. Keep 1 class node into package.
5. Try to align the package and outer class.

### Expected behavior

Helper line should work as expected with multiple elements(class within a package)

### Screenshots

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

### Additional context

_No response_

### Web Console Output

```shell

```

tamang29:
### Checklist

### Motivation and Context

This PR completes #552 

### Description
The alignment feature now prioritizes the parent element.
The child elements inside the parent can still have alignment features within themselves.

### Steps for Testing

1. Create a Class Diagram
2. Add Package element with a nested class element
3. Drag and drop another element and try alignment feature

### Screenshots

https://github.com/user-attachments/assets/ba348f59-d4fa-4266-b33f-e4887a9329ed


library/lib/utils/alignmentUtils.ts
```diff
@@ -31,6 +35,75 @@ export const getNodeBounds = (node: Node) => {
...
+  if (node.parentId) {
+    return node.parentId
+  }
+
+  const nodeBounds = getNodeBounds(node, allNodes)
+  let bestParent: Node | undefined
+  let bestArea = Infinity
+
+  for (const candidate of allNodes) {
+    if (candidate.id === node.id || !isParentNodeType(candidate.type)) {
```
Claudia-Anthropica: @tamang29 [medium] The dragged node itself stays in `allNodes` here, so when you drag a Package across the canvas, any free top-level node that the Package geometrically covers gets reclassified as a child of the moving Package (best-area wins). That node then fails `shouldUseAsGuideTarget` and its guides flicker out mid-drag. Either exclude the dragged node from being a parent candidate, or only count `parentId`-backed children when the candidate is the dragged node.

In `library/lib/utils/alignmentUtils.ts`, `getContainingParentId` treats the currently dragged parent-type node as a potential parent of unrelated free nodes it happens to overlap, causing those nodes to drop out of guide targets mid-drag. Pass the dragged node id into `getContainingParentId` (and `isContainedByParent`) and skip any candidate whose `id` equals it, so only nodes with an explicit `parentId` link to the dragged node are treated as its children.

</details>

library/lib/utils/alignmentUtils.ts
```diff
@@ -39,14 +112,21 @@ export const calculateAlignmentGuides = (
...
+  )
+  const draggedBounds = getNodeBounds(draggedNode, nodesWithDrag)
   const guides: AlignmentGuide[] = []
   const alignedPositions = new Set<number>()
 
-  const otherNodes = allNodes.filter((n) => n.id !== draggedNode.id)
+  const otherNodes = nodesWithDrag.filter(
+    (node) =>
+      node.id !== draggedNode.id &&
+      shouldUseAsGuideTarget(draggedNode, node, nodesWithDrag)
```
Claudia-Anthropica: @tamang29 [low] `shouldUseAsGuideTarget` recomputes `getContainingParentId(draggedNode, …)` for every candidate, and each call is itself O(n). That's O(n²) per `onNodeDrag` event for a value that doesn't change during the filter. Hoist `draggedParentId` (and the `isContainedByParent` decision when `!draggedParentId`) out of the filter.

In `library/lib/utils/alignmentUtils.ts` around the `otherNodes` filter, compute `draggedParentId = getContainingParentId(draggedNode, nodesWithDrag)` once before the filter, then pass it into a simplified inline predicate instead of recomputing it inside `shouldUseAsGuideTarget` per candidate.

</details>

library/lib/utils/alignmentUtils.ts
```diff
@@ -139,48 +219,52 @@ export const calculateAlignmentGuides = (
 export const snapNodeToGuides = (
   draggedNode: Node,
   guides: AlignmentGuide[],
-  threshold: number = ALIGNMENT_THRESHOLD
+  threshold: number = ALIGNMENT_THRESHOLD,
+  allNodes?: Node[]
```
Claudia-Anthropica, line 227: @tamang29 [low] You added `allNodes` to `snapNodeToGuides`, but grep shows it has no production caller — only `alignmentUtils.test.ts` uses it. `useNodeDrag` calls `calculateAlignmentGuides` but never `snapNodeToGuides`, so the snapping that this PR was supposed to make nested-aware never actually runs. Either wire `snapNodeToGuides` into `useNodeDrag` (passing the current nodes) or drop the new param — right now the feature is half-implemented.

In `library/lib/hooks/useNodeDrag.ts`, the drag callback calls `calculateAlignmentGuides` but never `snapNodeToGuides`, so the nested-aware snapping added in `library/lib/utils/alignmentUtils.ts` is dead. Either invoke `snapNodeToGuides(draggedNode, guides, undefined, nodes)` after computing guides and feed the resulting delta back into the node position, or revert the `allNodes` parameter addition in `snapNodeToGuides` if snapping is intentionally not wired up yet.

</details>

library/lib/utils/alignmentUtils.ts
```diff
@@ -31,6 +35,75 @@ export const getNodeBounds = (node: Node) => {
...
+) => {
+  const draggedParentId = getContainingParentId(draggedNode, allNodes)
+  const nodeParentId = getContainingParentId(node, allNodes)
+
+  if (!draggedParentId) {
+    return !nodeParentId && !isContainedByParent(node, allNodes)
+  }
+
+  return node.id === draggedParentId || nodeParentId === draggedParentId
+}
```
Claudia-Anthropica, line 108: @tamang29 [low] The entire point of the PR is nested-element guide behavior, but `library/tests/unit/alignmentUtils.test.ts` wasn't touched and has no nested cases. Please add tests for: (a) dragging a child with `parentId` set shows guides for siblings and the parent, (b) dragging a free top-level node does NOT pick up children of a Package, (c) dragging a Package itself only targets other top-level nodes. This will also lock in the geometric-containment fallback so a future refactor doesn't silently regress it.

In `library/tests/unit/alignmentUtils.test.ts`, extend the `calculateAlignmentGuides` suite with three nested-element cases: (1) a `package` parent at (0,0,400,400) with two `class` children using `parentId`, dragging one child should yield a vertical guide aligned with the sibling and the parent's left edge; (2) a free top-level dragged node should NOT produce guides against a class with `parentId` pointing at a package; (3) dragging the package itself should produce guides against another top-level package but not against children inside any package. Use the existing `makeNode` helper and extend it to accept `parentId`.

</details>

Claudia-Anthropica (changes requested):
@tamang29 Nice approach — the parent-aware filtering captures the intent in #552 cleanly. One subtle behavior issue with the geometric parent inference plus a few perf/test gaps inline; nothing major.

Claudia-Anthropica (changes requested):
@tamang29 The new commits here are just merges of `main` (#683, #688, #689) — `alignmentUtils.ts` is untouched, so all four points from my last review are still open: the medium parent-inference issue (a dragged Package reclassifies free nodes it overlaps mid-drag, flickering their guides out), the O(n^2) recompute in the `otherNodes` filter, the `snapNodeToGuides` change with no production caller, and the missing nested-element tests. Details are in the existing inline threads — keeping this as changes-requested until those are addressed.

library/lib/utils/alignmentUtils.ts
Claudia-Anthropica: @tamang29 Confirmed fixed — passing the dragged node id through as `excludeParentId` means a package dragged over a free top-level node no longer reclassifies it as a child, so its guides stay put mid-drag. The new top-level/package tests lock this in nicely. Thanks!

library/lib/utils/alignmentUtils.ts
Claudia-Anthropica: @tamang29 This one's still open: `snapNodeToGuides` got the new `allNodes` param, but it's still not called anywhere outside the tests — `useNodeDrag` only calls `calculateAlignmentGuides`, so the nested-aware snapping never actually runs. The helper-lines feature itself is complete, so I'm not blocking on it, but it'd be good to either wire the snapping into `useNodeDrag` or drop the unused param in a follow-up so it doesn't bitrot.

Claudia-Anthropica (approved):
@tamang29 The parent-inference bug is fixed (excludeParentId guard), the dragged-parent lookup is hoisted out of the per-candidate filter, and you added the nested-element tests — all green in CI. Nice work. Only leftover is the unused `allNodes` param on `snapNodeToGuides` (still no production caller); not blocking, just wire or drop it in a follow-up. Approving.

FelixTJDietrich (approved):
Works as an first iteration, code looks good. I still cannot align children with nodes outside the parent.

FelixTJDietrich (approved):
Changes look good, I tested it now and it worked as expected with the children and nodes outside the parent. Great work :D

library/lib/utils/alignmentUtils.ts
```diff
@@ -31,6 +35,78 @@ export const getNodeBounds = (node: Node) => {
...
+  return bestParent?.id
+}
+
+const shouldUseAsGuideTarget = (
+  draggedNode: Node,
+  node: Node,
+  allNodes: Node[],
+  draggedParentId: string | undefined
+) => {
+  const nodeParentId = getContainingParentId(node, allNodes, draggedNode.id)
```
Claudia-Anthropica, line 89: @tamang29 [nit] `nodeParentId` is computed for every candidate up here, but in the top-level branch (`!draggedParentId`) it's thrown away — so dragging any top-level node runs a full O(n) containment scan per candidate for nothing. The comment above `return true` is also confusing: it keeps the old "align with everything except…" text and then tacks on an "actually align with all nodes" correction. Moving the lookup below the early returns fixes both:

```ts
if (!draggedParentId) {
  // Top-level node: align with every other node, including parent frames.
  return true
}

if (node.id === draggedParentId) {
  return false // don't snap a child to its own frame
}

const nodeParentId = getContainingParentId(node, allNodes, draggedNode.id)
const isSibling = nodeParentId === draggedParentId
return isSibling || !nodeParentId
```

In `library/lib/utils/alignmentUtils.ts`, in `shouldUseAsGuideTarget`, move the `const nodeParentId = getContainingParentId(node, allNodes, draggedNode.id)` line down so it runs only after the `if (!draggedParentId) return true` and `if (node.id === draggedParentId) return false` early returns, since it is unused in those branches. Also replace the self-contradicting comment above `return true` with a single clear line stating that a top-level node aligns with all other nodes.

</details>

Claudia-Anthropica (approved):
@tamang29 The reworked targeting reads cleanly and does what Felix asked for — children now line up with nodes outside their parent, and the new sibling/top-level/package tests lock the behavior in. One small readability nit inline, nothing blocking. Approving.

PR #683, opened 2026-05-11
fix(webapp): blank screen on Share-again and New-diagram in collaboration

### LINKED ISSUE #679: [BUG] Creating a New File during collaboration is not working
### Describe the bug

During collaboration mode, if i try to create new file it breaks the webapp.

### To Reproduce

1. Create any diagram
2. Share -> Collaborate->enter name, join 2nd user via link
3. Try to create a New File-> Any diagram
4. FE breaks.

### Expected behavior

Working as expected

### Screenshots

https://github.com/user-attachments/assets/5cec6c1d-e772-4ba0-8655-b818cbde9c5e

### OS

_No response_

### Browser

_No response_

### Version

_No response_

### Additional context

_No response_

### Web Console Output

```shell

```

FelixTJDietrich:
### Checklist

### Motivation and Context

Two collaboration-mode bugs both manifested as a blank screen and both stemmed from the same lifecycle defect.

**Bug 1 — Share-again on an already-shared diagram.** From `/<id>?view=collaborate`, clicking Share → Collaborate navigates to `/<newId>?view=collaborate`. Same route pattern, so `ApollonWithConnection` does **not** unmount — its main `useEffect` re-runs because `diagramId` changed. The previous run's `isLoading=false` carried over, leaving the container `<Box>` visible-but-empty while the old editor was being destroyed and the new one was still fetching.

**Bug 2 — New diagram from a collab session.** Navigating `/<id>?view=collaborate` → `/` unmounts the page and remounts the local editor. `NewDiagramModal` was mutating `ModalContext` *after* `navigate("/")`, racing against React's unmount of the tree the modal was opened from.

### Description

`standalone/webapp/src/pages/ApollonWithConnection.tsx`
- Reset `isLoading=true` and the per-session refs (`diagramIsUpdated`, `lastObservedHeadRev`, `prePreviewFingerprintRef`, `restoredDuringPreviewRef`, `hasPromptedRef`) synchronously at the top of every effect entry.
- Thread `AbortController.signal` through all three `fetchDiagram` call sites (init, version-restored refetch, post-preview reload). Each catch filters `AbortError` before logging/toasting, so mid-fetch unmount is silent for the user.
- Capture `containerRef.current` into a local `container` so the post-`await` narrowing survives.

`standalone/webapp/src/services/WebSocketManager.ts`
- `onmessage` carries a `cleanedUp` short-circuit so frames the browser had buffered for us cannot dispatch into a destroyed editor or cleared listeners after `cleanup()`.
- `cleanup()` is idempotent; detaches `onopen`/`onmessage`/`onerror`/`onclose` before close; closes `CONNECTING`-state sockets too so the handshake doesn't leak.
- Control listeners are not explicitly cleared (the in-handler guard subsumes that path; GC reclaims them with the manager).

`standalone/webapp/src/services/DiagramApiClient.ts`
- `fetchDiagram` accepts an optional `{ signal }`.

`standalone/webapp/src/components/modals/NewDiagramModal.tsx`
- `closeModal()` runs *before* `navigate("/")` so modal-state mutations land on the still-mounted tree.

**Tests (12 new)**
- `standalone/webapp/src/pages/ApollonWithConnection.test.tsx` — 6 tests: initial loading overlay, overlay removed after mount, Share-again overlay re-shown, viewType-change overlay re-shown, AbortError swallowed on mid-fetch unmount, collab-name prompt re-opens on a second diagram.
- `standalone/webapp/src/services/WebSocketManager.test.ts` — 6 tests: in-handler guard on captured `onmessage` (data + control envelopes), handler detachment + 1001 close, idempotent cleanup, CONNECTING-state cleanup, `onerror` guard.

Every production guard is empirically gated: reverting any of the 5 production-side guards (`setIsLoading(true)`, `AbortError` catch filter, `hasPromptedRef.current = false`, WS `onmessage` `cleanedUp`, WS `onerror` `cleanedUp`) turns at least one test red.

### Steps for Testing

Reproduce locally with `npm run dev` and any existing collab diagram.

1. **Share-again.** Open a diagram, click Share → Collaborate (enter name) → `/<id>?view=collaborate`. Click Share again → Collaborate → confirm name → `/<newId>?view=collaborate`. Before the fix, the canvas area was blank for the duration of the fetch. After the fix, the "Loading diagram…" overlay appears immediately and disappears when the new editor mounts.
2. **New diagram from collab.** While in `/<id>?view=collaborate`, open File → New File → fill the form → Create Diagram. Lands on `/` with a fresh local editor; no blank screen.
3. **Rapid Share-again ×3–5.** No zombie cursors or duplicated collaborators in the presence bar; no `REVISION_MISMATCH` storm in the network log.
4. **Mid-fetch navigation.** Throttle the network, open a slow collab diagram, click Share → Collaborate before the first load completes. No "Failed to initialize" toast.
5. **Collab-name prompt across diagrams.** Open `/<id1>?view=collaborate` in a fresh `sessionStorage` (no name set), wait for the prompt, navigate to `/<id2>?view=collaborate` without confirming. The prompt re-opens on the second diagram.

### Screenshots

Failure modes are blank-screen / spurious-toast; the fix is the *absence* of the failure. The 12 regression tests pin each one and would surface a regression as a failing unit test in CI, which is the durable signal here.

### DIFF
--- a/standalone/webapp/src/components/modals/LoadDiagramModal.tsx
+++ b/standalone/webapp/src/components/modals/LoadDiagramModal.tsx
@@ -29,8 +29,8 @@ export const LoadDiagramModal = () => {
 
   const handleLoadingDiagram = (id: string) => {
     setCurrentModelId(id)
-    navigate("/")
     closeModal()
+    navigate("/")
   }
 
   const handleDeleteDiagram = (id: string) => {
--- a/standalone/webapp/src/components/modals/NewDiagramFromTemplateModal.tsx
+++ b/standalone/webapp/src/components/modals/NewDiagramFromTemplateModal.tsx
@@ -43,12 +43,11 @@ export const NewDiagramFromTemplateModal = () => {
       const timeStapToCreate = new Date().getTime()
 
       createModel(jsonData)
+      closeModal()
       navigate("..", {
         relative: "route",
         state: { timeStapToCreate },
       })
-
-      closeModal()
     } catch (err: unknown) {
       log.error("Error creating diagram from template:", err as Error)
 
--- a/standalone/webapp/src/components/modals/NewDiagramModal.tsx
+++ b/standalone/webapp/src/components/modals/NewDiagramModal.tsx
@@ -58,9 +58,10 @@ export const NewDiagramModal = () => {
   )
 
   const handleCreateDiagram = () => {
+    // Close before navigate: ModalContext writes after route unmount race.
     createModelByTitleAndType(newDiagramTitle, selectedDiagramType)
-    navigate("/")
     closeModal()
+    navigate("/")
   }
 
   const handleDiagramNameChange = (
--- a/standalone/webapp/src/components/modals/ShareModal.tsx
+++ b/standalone/webapp/src/components/modals/ShareModal.tsx
@@ -27,6 +27,7 @@ export const ShareModal = () => {
 
       const newurl = `${window.location.origin}/${diagramID}?view=${viewType}`
       copyToClipboard(newurl)
+      closeModal()
       navigate(`/${diagramID}?view=${viewType}`)
 
       toast.success(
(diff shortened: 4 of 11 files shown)


FelixTJDietrich:
Merging it already so we can deploy it for production

PR #691, opened 2026-05-19
feat: monorepo modernization — pnpm 11 + Node 24, Docusaurus docs, React `<Apollon>` API, VS Code rewrite

FelixTJDietrich:
## Summary

This branch does three things that ship independently of each other:

1. **Modernises the monorepo plumbing** — pnpm 11.1.3 + Node 24.15.0 LTS, Vite everywhere, server CJS → ESM. One lockfile. No more webpack, no more Parcel, no more CJS/ESM coexistence dance.
2. **Adds a real React API to the library** — a `<Apollon>` component plus `ApollonProvider`, `useApollonEditor`, `useApollonEditorOrThrow`, `useApollonSubscription`. React-19-style `onMount` cleanup return, `"use client"` directive that actually survives the bundle. Webapp + VS Code editor migrated; **zero `new ApollonEditor(...)` in repo outside the library itself.**
3. **Replaces the static Sphinx docs with a Docusaurus 3.10 site** — Library / User / Contributor sidebars, a live in-browser editor on the landing page, deployed to GitHub Pages by a new `docs.yml` workflow on every push to `main`.

The library is **dual-build** by design: the default subpath (`@tumaet/apollon`) bundles React + MUI + emotion + xyflow so non-React hosts (Angular/Artemis is the primary consumer) install with **zero peer deps**. The `/react` subpath externalises them so React hosts can dedupe. Both subpaths share the same stylesheet via `./style.css`.

## Why this PR is large

The original scope was "pnpm + Node 24". Mid-review, two things became unavoidable:

- The library shipped as React-only mid-PR (commit `c79683e8`), which would have broken Artemis (Angular). That got reverted (`c64d88b2`); the dual-build is now part of the same PR because reverting it cleanly required tooling that didn't exist on `main`.
- The docs were Sphinx + ReadTheDocs and the `requirements.txt`/`Makefile`/`bin/plantuml.jar` machinery had to go anyway. Once Sphinx was out, Docusaurus and the React API were the natural place to land — and they reuse the same pnpm workspace plumbing this PR introduces.

Splitting into three PRs would have meant `main` ships in an awkward in-between state for two of the three weeks.

### Build & toolchain

| | Before | After |
|---|---|---|
| Package manager | npm | **pnpm 11.1.3** (Corepack-verifiable SHA512 pin in `packageManager`) |
| Node | 22 | **24.15.0** (`.nvmrc`, `engines`, Dockerfiles, all CI runners) |
| Bundlers | webpack + Parcel + Vite | **Vite only** (server: `tsc`) |
| Module system | mixed CJS / ESM | **ESM end-to-end** (server `"type": "module"`, `NodeNext`, `verbatimModuleSyntax`) |
| Lockfiles | 3 × `package-lock.json` | 1 × `pnpm-lock.yaml` |

The server migration alone is 22 source files / 76 import specifiers gaining `.js` suffixes plus a static `import { ApollonEditor, importDiagram, ... } from "@tumaet/apollon"` replacing the cargo-cult `loadApollonModule = () => import(...)` lazy wrapper. Express 5's automatic async-handler propagation lets `routes/conversion.ts` and `middleware/validate.ts` drop their try/catch ceremonies.

### Library — public API

```tsx
import { Apollon } from "@tumaet/apollon/react"
import "@tumaet/apollon/style.css"

<Apollon
  style={{ height: 600 }}
  defaultModel={savedDiagram}
  readonly={isReadonly}        // reactive
  onMount={(editor) => {        // React-19 cleanup-return idiom
    const id = editor.subscribeToModelChange(persist)
    return () => editor.unsubscribe(id)
  }}
/>
```

Props split into **initial-only** (snapshotted at mount; re-key to rebuild) and **reactive** (applied through the matching `editor.setX(...)` when the prop changes). The split is documented in source JSDoc, in `/library/api`, and in `/library/embedding/react` — and verified by a small `Apollon.test.tsx` that asserts the contract three ways (constructor argument shape, ref forwarding, reactive-setter dispatch).

Imperative `ApollonEditor` is unchanged for non-React hosts. The constructor still takes `(element, options?)`; the standalone subpath still re-exports the full surface.

**`"use client"` actually works.** Rollup strips source-level directives during bundling; verified that the previous bundle started with `var f0 = …`, not `"use client";`. Re-added via Rollup `output.banner` on the peer pass only. `head -1 dist/react/react.js` now prints `"use client";` — Next.js App Router consumers don't need a wrapper.

### Library — bundle shape

| Subpath | Peers | Bundle |
|---|---|---|
| `@tumaet/apollon` (default) | bundled | ~2.4 MB |
| `@tumaet/apollon/react` | externalised | ~875 KB raw / ~100 kB brotli |
| `@tumaet/apollon/internals` | (unstable; host integration tests) | ~1 KB + reused yjs chunk |
| `@tumaet/apollon/style.css` | — | 21 KB |

`peerDependenciesMeta.optional` marks all six peers optional, so `npm install @tumaet/apollon` never warns about missing React/MUI/etc. — even though the default subpath inlines them.

### VS Code extension

| | Before | After |
|---|---|---|
| Extension main bundle | 18.2 KB (webpack, 940 ms) | **8.2 KB** (Vite, 87 ms) |
| Menu webview | ~360 KB (sunset UI toolkit) | **146 KB** (native HTML + VS Code CSS vars) |
| Editor webview theme | locked to `setTheme("light")` | **follows VS Code's auto-applied light/dark/high-contrast** |
| Deprecated runtime deps | 4 | **0** |
| Release pipeline | manual `vsce` from a laptop | **GitHub Actions**, environment-gated, sigstore-attested, dual-registry (Marketplace + Open VSX) |

Dropped: `@vscode/webview-ui-toolkit` (Microsoft sunset 2026-01-06), `styled-components`, `uuid`, `@vscode/codicons`. Replacement is native HTML elements styled via `var(--vscode-*)` per Microsoft's post-deprecation guidance. The editor webview imports from `@tumaet/apollon/react` to share its React copy with the host webview.

### Docs

Static Sphinx site → Docusaurus 3.10 at <https://ls1intum.github.io/Apollon/>. Three sidebars (Library / User / Contributor), live in-browser editor on the landing page (`docs/src/components/ApollonEmbed`), Diátaxis-shaped Library section, deployed by `.github/workflows/docs.yml` on every `main` push. `onBrokenLinks: throw` and `onBrokenAnchors: throw` keep cross-refs honest.

### Monorepo hygiene

- pnpm catalogs for shared version pins (`pnpm-workspace.yaml`).
- Override floors tightened to the minimum fixed version per CVE (no more mass major-bumps).
- `HEALTHCHECK` in both Dockerfiles; digest-pinned base images.
- `SECURITY.md`, `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, `.editorconfig`, `.gitattributes` added.
- Renovate gains `helpers:pinGitHubActionDigests` so SHA pins don't rot.
- All third-party GitHub Actions are SHA-pinned with version comments.
- `release-library.yml` hardened against pnpm 11's varying `pnpm pack --json` output shape (`pnpm/pnpm#10200`).

## Required GitHub setup before the first VS Code Marketplace release

`release-vscode-extension.yml` is gated on a `vscode-marketplace` Environment that doesn't exist on this repo yet — the workflow will hang on approval until it's created:

1. Settings → Environments → New environment → name `vscode-marketplace`
2. Deployment branches: "Selected branches and tags" → add `main`
3. Required reviewers: add the release maintainer; turn on **Prevent self-review**
4. Environment secrets:
   - `VSCE_PAT` — Azure DevOps PAT, scope *Marketplace → Manage*, 1-year max lifetime
   - `OVSX_PAT` — Open VSX PAT. The `tumaet` namespace must exist on open-vsx.org first; any namespace member can create it once with `ovsx create-namespace tumaet -p <PAT>`
5. **Delete any pre-existing repo-level `VSCE_PAT` / `OVSX_PAT`** — repo secrets bypass the environment's deployment-branch and reviewer controls.

Full checklist also at `docs/contributor/deployment/npm-publishing.md`.

## How to review

The PR is large by line count but the **architecture is small** — three orthogonal concerns that don't touch each other much. Pick the angle that matches your expertise:

### If you maintain the library

Focus on `library/lib/components/react/` (~330 lines net new) and the dual-build in `library/vite.config.ts`. The questions worth answering:

1. **Is the initial-only vs reactive prop split right?** `Apollon.tsx` JSDoc lists which is which. Passing `undefined` to a reactive prop leaves the live value alone — re-key to fully reset.
2. **Does `onMount`'s cleanup-return ordering make sense?** `consumer cleanup → editor.destroy() → null the ref`. The comment in `Apollon.tsx` explains why; the test `Apollon.test.tsx` pins it.
3. **Is the `"use client"` banner the right escape hatch?** Rollup strips the source directive. The `vite.config.ts` comment + `head -1 dist/react/react.js` should answer it.
4. **Does the dual-build genuinely keep React out of the standalone subpath?** Smoke test below.

### If you maintain the standalone server / webapp / VS Code extension

The webapp + VS Code editor are migrated from imperative `new ApollonEditor(...)` to `<Apollon>`. `grep -rn "new ApollonEditor" --include="*.tsx"` outside `library/` should return zero hits. `ApollonWithConnection.tsx` is the most complex consumer (collab cursors, version preview, autosave, WS) — focus there.

Server: 22 files gain `.js` import suffixes. The substantive changes are in `conversion-service.ts` (static import replacing lazy wrapper) and the Express-5 async-handler simplifications in `routes/conversion.ts` + `middleware/validate.ts`.

VS Code: `editor/src/App.tsx` now renders `<Apollon>` with a `key={loadVersion}` for "loadDiagram" remounts; the hand-rolled `ApollonEditorContext`/`ApollonEditorComponent` (which duplicated the library's `ApollonProvider`/`useApollonEditor`) is deleted.

### If you maintain CI / deployment

Three new workflow files: `docs.yml` (GitHub Pages publish), `release-vscode-extension.yml` (sigstore + dual-registry), and the new `library-node22-compat` job in `pr-health-checks.yml` (the library still advertises `engines.node: ">=22.0.0"` for downstream consumers even though the monorepo is on 24).

The `vscode-marketplace` environment setup above is the only blocking action required.

### If you're reviewing docs

The Library section follows Diátaxis: Quickstart (tutorial) → Embedding (how-to) → API (reference) → Troubleshooting (explanation/diagnosis). The API page is **canonical**: embedding pages link there for prop tables instead of duplicating. `onBrokenLinks: throw` in `docusaurus.config.ts` means the production build is a transitive link-check.

## Steps for testing

```sh
# 1. Fresh install — expect a single pnpm-lock.yaml, zero ignored-build-script warnings,
#    no leftover package-lock.json files anywhere.
nvm install && nvm use
npm install -g pnpm@11.1.3
pnpm install

# 2. Build everything
pnpm build          # library + server + webapp (library first; tsc -b enforces order)
pnpm build:vscode   # extension + 2 webviews
pnpm build:docs     # library + Docusaurus site (onBrokenLinks: throw)

# 3. Lint + tests
pnpm lint           # ESLint per workspace + markdownlint (docs)
pnpm test           # 760 library unit tests
pnpm test:e2e       # Playwright

# 4. Verify the "use client" directive survives the bundle — this is the
#    headline correctness check for Next.js App Router consumers
head -1 library/dist/react/react.js          # → "use client";

# 5. Smoke-test both library subpaths in isolation
node --input-type=module -e \
  'import("./library/dist/index.js").then(m => console.log("standalone:", Object.keys(m).length, "exports"))'
node --input-type=module -e \
  'import("./library/dist/react/react.js").then(m => console.log("/react:", Object.keys(m).length, "exports"))'
```

The standalone resolves with **no peer deps installed** (Artemis case). The `/react` subpath resolves only when `react`, `react-dom`, `@mui/material`, `@emotion/{react,styled}`, and `@xyflow/react` are present. Both share the same `./style.css` export.

## Out of scope (deliberately deferred)

| What | Why deferred | Tracked where |
|---|---|---|
| Library version bump from 4.4.0 | `version-bump.yml` handles release bumps post-merge | Existing workflow |
| `arethetypeswrong` CI gate | Dropped in `bb72249b` — fflate streaming bug prevents tarball extraction on our shape; `publint` + exports-map check still cover it | Re-add when fflate lands a fix |
| React 19 in peer range | Reviewer-confirmed Round 6: narrow to `^18.3.0` matches reality. Widening requires a ref-type refactor (`anchorRef: RefObject<SVGSVGElement \| null>` plus call-site adjustments where the SVG ref crosses `SVGForeignObjectElement` boundaries) | Follow-up PR |
| MUI emotion CSP `'unsafe-inline'` in VS Code webviews | Needs a library `cspNonce` option threaded into emotion's `createCache` | [emotion-js/emotion#403](https://github.com/emotion-js/emotion/issues/403) |
| Server unit-test isolation flake (`tryAutoVersion` races `flushDb`) | Pre-existing on `main` — verified by baseline | Unchanged scope |
| Mobile (iOS / Android) build pipeline | Capacitor dep cleanup is in scope; native builds are untouched | Future PR |
| `export *` barrel from `lib/index.tsx` still leaks ~33 unused identifiers (caught by `knip`) | Cutting them is SemVer-major | Next release bump |

## Screenshots

No UI changes inside Apollon itself. The VS Code editor webview now follows VS Code's theme automatically (light/dark/high-contrast); it used to be locked to light. The Docusaurus site is the new public-facing surface — landing page at `docs/src/pages/index.tsx`, hosted at <https://ls1intum.github.io/Apollon/> after merge.

## Checklist

- [x] No linked issue — foundational migration, not a tracked feature
- [x] Library version unchanged (4.4.0, matches `main`) — release bump handled by `version-bump.yml` post-merge
- [x] Dual-build preserved: Angular hosts (Artemis primary) install with **zero peer deps**; React hosts opt into `/react`
- [x] `"use client";` directive present at byte 0 of `dist/react/react.js` (the previous bundle was silently missing it)
- [x] Docusaurus build clean (`onBrokenLinks: throw`, `onBrokenAnchors: throw`)
- [x] All required CI workflows green on the latest push


standalone/server/src/logger.ts
```diff
@@ -1,13 +1,13 @@
...
 
 const level =
   process.env.LOG_LEVEL ??
   (process.env.NODE_ENV === "production" ? "info" : "debug")
 
 export const logger: Logger = pino({
   level,
-  base: undefined,
-  timestamp: pino.stdTimeFunctions.isoTime,
+  base: null,
```
Claudia-Anthropica: @FelixTJDietrich [low] Heads-up that this is a runtime behavior change, not a no-op port: pino reads `base: undefined` as "use defaults (pid + hostname)" and `base: null` as "omit base fields". So before this PR every log line carried `pid` and `hostname`; after it they're gone. If anything in the Helios / staging / prod log pipeline filters or joins on `pid`/`hostname`, it'll silently lose those fields. Likely forced by the new `exactOptionalPropertyTypes: true` rejecting `undefined`, but worth confirming whether dropping pid/hostname was the intent or a side effect.

In `standalone/server/src/logger.ts:9`, `base: undefined` was changed to `base: null` while migrating the tsconfig to `exactOptionalPropertyTypes`. The semantics differ — `undefined` keeps pino's default `{pid, hostname}` while `null` drops both. Decide whether the production logs should keep pid/hostname; if yes, omit the `base` key entirely instead of setting it to `null`. If no, leave it as-is and add a brief comment noting the deliberate drop.

</details>

.github/workflows/release-library.yml
```diff
@@ -66,24 +66,39 @@ jobs:
...
-      - run: npm run test
+      - name: Pack
+        id: pack
         working-directory: library
+        # `pnpm pack --json` writes the tarball path to stdout as structured
+        # output; previous `pnpm pack | tail -n 1` broke when pnpm logged
+        # warnings on stdout (pnpm/pnpm#10200).
+        run: |
+          set -euo pipefail
+          TARBALL=$(pnpm pack --json | jq -r '.filename // .[0].filename')
```
Claudia-Anthropica: @FelixTJDietrich [low] `jq '.filename // .[0].filename'` works for pnpm 11's current object output, but if pnpm ever switches `pack --json` to return an array (it has flip-flopped before), `.filename` on an array throws `Cannot index array with string` and jq's `//` does NOT catch that — it only swallows null/false. The fallback then never fires and the release breaks. A `type`-guarded form is safer.

```yaml
TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')
```

In `.github/workflows/release-library.yml:94`, the jq expression `'.filename // .[0].filename'` will throw on array output because `//` only handles null/false, not type errors. Replace it with `'if type=="array" then .[0].filename else .filename end'` so the workflow tolerates both pnpm output shapes.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Big, well-structured migration — pnpm wiring, CHANGELOG, README, security overrides, SHA-pinned actions/images, the OVSX checksum-verify retry, and the Node-22 compat job for the library all look great. Two things worth syncing on before merge: the PR description still describes the original dual-build design (`@tumaet/apollon/react` + optional peers) that was pivoted away from in `c79683e8 refactor(library)!: single bundle with mandatory peers, v5.0.0` — the CHANGELOG/README are correct, but anyone reading the PR description thinks default behavior is unchanged when it's actually breaking for non-React consumers. Codacy also flagged 22 stylelint notices in the new webview CSS (color-hex-length, rule-empty-line-before, no-descending-specificity) plus 3 false-positive Dockerfile warnings (your images ARE SHA-pinned — that's stricter than tag-pinning, Codacy just doesn't recognize the digest syntax). Inline comments below.

standalone/server/src/logger.ts
Claudia-Anthropica: @FelixTJDietrich Just flagging that this is still `base: null` on the current head — the pid/hostname drop hasn't been adjusted. If shedding those fields was deliberate, no action needed; otherwise the fix is just to omit the `base` key entirely (or spread it conditionally) so pino keeps its defaults.

.github/workflows/release-library.yml
Claudia-Anthropica: @FelixTJDietrich Still seeing `'.filename // .[0].filename'` here too — pnpm flipping back to array output would silently break the release path. The `if type=="array" then .[0].filename else .filename end` guard is two seconds to add and tolerates either shape.

.github/workflows/pr-health-checks.yml
```diff
@@ -129,17 +141,63 @@ jobs:
...
+
+      - name: Lint published package shape (publint)
+        working-directory: library
+        run: pnpm dlx publint
+
+      - name: Validate type resolution (arethetypeswrong)
+        working-directory: library
+        # ESM-only library, all exports flagged as such; CJS profile
+        # would force-flag dual-package issues that don't apply.
+        run: pnpm dlx -p @arethetypeswrong/cli attw --pack . --profile esm-only
```
Claudia-Anthropica: @FelixTJDietrich [medium] The new gate is failing on this PR — attw exits with `error while checking file: ENOENT: no such file or directory, open 'attw'`. With `pnpm dlx -p <pkg> <bin>` on pnpm 11, the binary name leaks through as a positional argument to attw, which then tries to open a tarball literally named `attw`. Easiest fix is to drop `-p` since `attw` is the default bin of `@arethetypeswrong/cli`:

```yaml
run: pnpm dlx @arethetypeswrong/cli --pack . --profile esm-only
```

or use the long-form `--package=` so pnpm parses it cleanly:

```yaml
run: pnpm dlx --package=@arethetypeswrong/cli attw --pack . --profile esm-only
```

In `.github/workflows/pr-health-checks.yml:198`, the step `pnpm dlx -p @arethetypeswrong/cli attw --pack . --profile esm-only` fails because pnpm 11's `dlx -p` short form passes the bin name as a positional arg to attw, yielding `ENOENT: no such file or directory, open 'attw'`. Replace it with `pnpm dlx @arethetypeswrong/cli --pack . --profile esm-only` (attw is the default bin) or `pnpm dlx --package=@arethetypeswrong/cli attw --pack . --profile esm-only`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Followup. Nice cleanup pass in `fabd4382` — the catalog block, ESLint coverage for the webviews, and the publint gate all look right. Two things still open from the previous review (logger `base: null` dropping pid/hostname, jq fallback in `release-library.yml`) — replied inline. New blocker: the freshly-added arethetypeswrong gate is failing on its very first run; details inline. Codacy is also still red.

standalone/server/src/logger.ts
Claudia-Anthropica: @FelixTJDietrich Round 3 — still `base: null` on line 9. The pid/hostname drop is unchanged. Drop the `base` key entirely (or spread it conditionally) if you want to keep pino's defaults; otherwise leave a one-line comment marking the drop as intentional so this doesn't get re-flagged.

.github/workflows/release-library.yml
Claudia-Anthropica: @FelixTJDietrich Round 3 — still `'.filename // .[0].filename'` on line 94. Two seconds to swap in the type guard and you stop being a hostage to pnpm's `pack --json` output shape.

```yaml
TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')
```

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich The fix attempt didn't land — dropping `-p` was right, but `attw` is still in there as a positional and `pnpm dlx <pkg> [args]` passes those args straight to the bin. CI confirms: `error while checking file: ENOENT: no such file or directory, open 'attw'`. Just delete the `attw` token:

```yaml
run: pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only
```

`attw` is the package's only bin, so pnpm dlx picks it up implicitly.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round-4 cleanup looks good (nice work on the Vitest 4 flake root-cause), but the three open threads from the previous review are all unchanged and `library-node22-compat` is still red on the same attw invocation — see thread replies. Pinging back here when those three are addressed.

standalone/server/src/logger.ts
Claudia-Anthropica: @FelixTJDietrich Round 4 — still `base: null` on this line, pid/hostname still dropped. Either delete the `base` key or leave a one-line comment explaining the drop is intentional. Otherwise this stays on the list.

.github/workflows/release-library.yml
Claudia-Anthropica: @FelixTJDietrich Round 4 — `'.filename // .[0].filename'` is still on line 94. Same suggestion as before:

```yaml
TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')
```

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Round 4 — `attw` is still a positional in `pr-health-checks.yml` (line 200 now) and CI on `9b9bd23a` confirms `ENOENT: no such file or directory, open 'attw'` again. The comment above the run line says the binary is resolved implicitly; the command needs to actually match that. Drop the token:

```yaml
run: pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only
```

.github/workflows/pr-health-checks.yml
```diff
@@ -0,0 +1,326 @@
...
+            @types/react@19 \
+            @types/react-dom@19 \
+            @testing-library/react@16 \
+            --save-exact --ignore-scripts
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile=false --config.engine-strict=false
+
+      - name: Build library against React 19
+        run: pnpm --filter @tumaet/apollon run build
```
Claudia-Anthropica: @FelixTJDietrich [high] This new job is failing on the latest run — React 19's tighter `useRef` types break 13 library files. `useRef<SVGSVGElement | null>(null)` now yields `RefObject<SVGSVGElement | null>` and is no longer assignable to the `RefObject<SVGSVGElement>` prop type on `CommonEdgeElements.anchorRef` (and the matching prop in `GenericEdge`). Either widen the prop types to `RefObject<SVGSVGElement | null>` (matches what the ref actually is) or change the call sites to `useRef<SVGSVGElement>(null)`. The whole reason this job exists is to back the `^18.3 || ^19` peer-range claim — right now it's documenting the opposite.

In `library/lib`, the new CI job `library-react19-compat` (added in `.github/workflows/pr-health-checks.yml`) fails because `useRef<SVGSVGElement | null>(null)` is no longer assignable to `RefObject<SVGSVGElement>` under React 19's tightened ref types. The affected files include `library/lib/edges/edgeTypes/{ActivityDiagramEdge,BPMNDiagramEdge,ClassDiagramEdge,CommunicationDiagramEdge,ComponentDiagramEdge,DeploymentDiagramEdge,FlowChartEdge,ObjectDiagramEdge,PetriNetEdge,ReachabilityGraphArc,SfcDiagramEdge,SyntaxTreeEdge,UseCaseDiagramEdge}.tsx`, plus the prop declaration on `library/lib/edges/GenericEdge.tsx:187` (`anchorRef: React.RefObject<SVGSVGElement>`). Fix by widening that prop type (and any matching prop on `CommonEdgeElements`) to `React.RefObject<SVGSVGElement | null>` so the existing nullable `useRef` declarations type-check under React 19. Rerun `pnpm run build` inside the library workspace after the change to confirm zero TS errors.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 4. Quality-gates push is a solid expansion of scope — `size-limit`, the new react19 job, the rules-of-hooks fixes, and the internals move all look right. But the three threads from rounds 1–3 are all still unchanged on `2b64efa6` (`base: null` in `logger.ts`, `'.filename // .[0].filename'` in `release-library.yml`, `attw` positional in `pr-health-checks.yml` — the last one is still actively failing CI as `ENOENT: no such file or directory, open 'attw'`). And the new `library-react19-compat` job is itself red: 13 `RefObject<SVGSVGElement | null>` → `RefObject<SVGSVGElement>` type errors, so the peer-range claim the job was added to verify is in fact broken. Threads stay open until they land.

standalone/server/src/logger.ts
Claudia-Anthropica: @FelixTJDietrich Round 5 — `base: null` unchanged on this line through the rename commit. Same suggestion as rounds 1–4: drop the `base` key (keeps pino defaults) or leave a one-line comment marking the pid/hostname drop as deliberate. Whichever direction, just resolve it.

.github/workflows/release-library.yml
Claudia-Anthropica: @FelixTJDietrich Round 5 — still `'.filename // .[0].filename'`. Same one-line swap as rounds 1–4:

```yaml
TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')
```

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Round 5 — `pr-health-checks.yml:200` still has `attw` as a positional, and `library-node22-compat` is once again red on `ENOENT: no such file or directory, open 'attw'`. The comment two lines above ("pnpm dlx resolves the binary from the package name implicitly") describes the fix; the command just needs to actually drop the token:

```yaml
run: pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only
```

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Round 5 — `library-react19-compat` still fails with the same 13 `RefObject<SVGSVGElement | null>` → `RefObject<SVGSVGElement>` errors. The rename commit didn't touch the edge files or `GenericEdge.tsx:187`. Widening the prop type to `React.RefObject<SVGSVGElement | null>` is a one-line change that unblocks the whole job.

scripts/dev.mjs
```diff
@@ -455,14 +455,14 @@ async function main() {
...
-  console.log(`[dev]   webapp:    http://localhost:${webappPort}`)
+  console.log(`[dev]   app:    http://localhost:${webappPort}`)
   console.log(`[dev]   server:    http://127.0.0.1:${serverPort}`)
   console.log(`[dev]   websocket: ws://127.0.0.1:${websocketPort}`)
   console.log(`[dev]   redis:     ${redis.url} (${redis.source})`)
 
   const sharedEnv = {
     ...process.env,
-    APOLLON_WEBAPP_PORT: String(webappPort),
+    APOLLON_APP_PORT: String(webappPort),
```
Claudia-Anthropica: @FelixTJDietrich [medium] The producer side renamed to `APOLLON_APP_PORT` here, but the consumer in `standalone/app/vite.config.ts:8` still reads `process.env.APOLLON_WEBAPP_PORT` (the rename was a pure file-move on that workspace, so its contents weren't touched). Net effect: the port this script just allocated and printed as `[dev] app: http://localhost:${webappPort}` is ignored, vite falls back to the literal `5173`, and if 5173 was already taken (or `findAvailablePort` picked something else for collision-avoidance) you get a silent mismatch between the URL the orchestrator logs and where vite actually binds. Fix: update `standalone/app/vite.config.ts:8` to read `process.env.APOLLON_APP_PORT`.

In `standalone/app/vite.config.ts:8`, change `process.env.APOLLON_WEBAPP_PORT` to `process.env.APOLLON_APP_PORT` to match the renamed env var that `scripts/dev.mjs:465` now exports. Without this, the dev orchestrator's allocated port is ignored and vite always falls back to 5173.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 5. The rename commit is clean as a refactor, but it didn't touch any of the four threads from rounds 1–4 — `base: null`, the jq fallback, the `attw` positional, and the React 19 ref types are all still in the tree, and `library-node22-compat` + `library-react19-compat` are still red on the same errors. New thing the rename broke: `scripts/dev.mjs` exports `APOLLON_APP_PORT` while `standalone/app/vite.config.ts` still reads `APOLLON_WEBAPP_PORT`, so the dev orchestrator's port allocation is silently ignored — flagged inline.

scripts/dev.mjs
Claudia-Anthropica: @FelixTJDietrich The webapp rename revert took this with it — `scripts/dev.mjs` and `standalone/webapp/vite.config.ts` are both back on `APOLLON_WEBAPP_PORT`, so producer/consumer match again. Resolving.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 6. Rename commit itself is clean, and the webapp-rename revert resolved the `APOLLON_APP_PORT` mismatch — thread closed. The four threads from rounds 1–5 (`base: null` in logger.ts, jq fallback in release-library.yml, `attw` positional in pr-health-checks.yml, React 19 useRef widening) are all still unchanged on HEAD, and `library-react19-compat` is still red on CI. I won't repeat the substance again here — see the existing threads. Keeping at `REQUEST_CHANGES` until those land or you push back on any of them.

standalone/server/src/logger.ts
Claudia-Anthropica: @FelixTJDietrich Round 6 — `base: null` still on line 9 through the React 19 trim. No new info from my side; treating the radio silence as a deliberate hold, so this is the last ping on it.

.github/workflows/release-library.yml
Claudia-Anthropica: @FelixTJDietrich Round 6 — `'.filename // .[0].filename'` still on line 94. Same one-liner swap as before; treating the silence as deliberate, last ping from me on this.

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Round 6 — `library-node22-compat` still red on this head with the same `ENOENT: no such file or directory, open 'attw'` (run 26133072450, job 76862199241). The comment immediately above the run line says "pnpm dlx resolves the binary from the package name implicitly"; the line below still passes `attw` as a positional, which pnpm then forwards to the bin as a tarball path. This is the actual merge blocker — the other two threads are nice-to-haves, this one breaks the gate.\n\n```yaml\nrun: pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only\n```

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Dropping the React 19 matrix and narrowing the peer range to `^18.3.0` is the right call — the lib's contract now matches what CI actually tests, and the commit message is honest about why. Resolving.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 6. The React 19 thread is resolved — dropping the matrix and narrowing the peer range to `^18.3.0` matches reality, that's a clean call. Still blocking: `library-node22-compat` is red on this head with the same `attw` ENOENT (token still passed as a positional). The two `[low]` threads (`base: null`, jq fallback) are unchanged through six rounds — flagged once more, then leaving them alone.

.gitignore
```diff
@@ -68,3 +55,5 @@ ios/
 icons/
 venv/
 .context/
+docs/.docusaurus/
```
Claudia-Anthropica, line 58: @FelixTJDietrich [medium] Adding `docs/.docusaurus/` to `.gitignore` in the same commit that adds the 56 files inside it doesn't untrack them — `git ls-files docs/.docusaurus | wc -l` returns 56 on this head, and the folder's own `DONT-EDIT-THIS-FOLDER` marker confirms it's regenerated build cache. Net effect: every `pnpm --filter @tumaet/apollon-docs run build` will dirty the working tree, the docs CI cache (`actions/cache` step in `docs.yml`) restores stale entries on top of committed ones, and the next contributor to run a build will be tempted to commit the churn. Drop the cached copy.

```sh
git rm -r --cached docs/.docusaurus
```

In this branch, `.gitignore` adds `docs/.docusaurus/` but the 56 files under that path are still tracked in the same commit, so the ignore rule has no effect. Run `git rm -r --cached docs/.docusaurus` and commit the deletion (the directory is regenerated by `docusaurus build` and shouldn't be in the repo). Verify with `git ls-files docs/.docusaurus | wc -l` returning 0.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Three earlier threads are still open on this head: `standalone/server/src/logger.ts` (`base: null` pid/hostname drop), `.github/workflows/release-library.yml` (the jq filter), and the attw positional in `.github/workflows/pr-health-checks.yml` — that last one is still the actual merge blocker (CI is currently pending but the line is unchanged from the last red run). One new thing in the docs migration commit inline.

.gitignore
Claudia-Anthropica: @FelixTJDietrich Nice, the round-6 commit untracked them — `git ls-files docs/.docusaurus` is now 0. Resolving.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 7 / docs polish. The `.docusaurus/` untrack and the MyST→Docusaurus + frontmatter pass on this commit look clean — that thread is resolved. The three pre-docs threads (`logger.ts` `base: null`, `release-library.yml` jq filter, `pr-health-checks.yml` `attw` positional) are all still on the current head; the `attw` one is the merge blocker since `library-node22-compat` keeps going red on the same `ENOENT: ... open 'attw'`. Holding at REQUEST_CHANGES on that one thread.

package.json
```diff
@@ -1,97 +1,71 @@
...
-    "format:check": "prettier --ignore-path .prettierignore --check \"**/*.(js|ts|tsx|json)\"",
-    "test": "npm run test --workspace=@tumaet/apollon",
-    "test:e2e": "npm run test:e2e --workspace=@tumaet/webapp",
+    "ensure:localdb": "pnpm run start:localdb && pnpm run wait:localdb",
+    "lint": "concurrently \"pnpm run lint:lib\" \"pnpm run lint:server\" \"pnpm run lint:webapp\" \"pnpm run lint:vscode\" \"pnpm run lint:docs\"",
+    "lint:lib": "pnpm --filter @tumaet/apollon run lint",
+    "lint:server": "pnpm --filter @tumaet/server run lint",
+    "lint:webapp": "pnpm --filter @tumaet/webapp run lint",
+    "lint:vscode": "pnpm --filter apollon-vscode run lint",
+    "lint:docs": "pnpm --filter @tumaet/apollon-docs run lint",
```
Claudia-Anthropica, line 32: @FelixTJDietrich [high] This new `lint:docs` runs `tsc --noEmit` in `docs/`, which imports `@tumaet/apollon` and needs `library/dist/` to exist — but in `lint-and-format-check`, `Run Lint` fires before `Check build`. The latest CI run fails on `docs/src/components/ApollonEmbed/index.tsx(7,8): error TS2307: Cannot find module '@tumaet/apollon'`. Either build the library before the concurrent lint block (e.g. a `prelint` hook or sequencing in the root `lint`), or pull docs `typecheck` out of `lint` and run it after `Check build` in the workflow.

In `package.json:30`, the new `lint:docs` entry (`pnpm --filter @tumaet/apollon-docs run lint`) runs `tsc --noEmit` in `docs/`, which imports `@tumaet/apollon` and therefore needs `library/dist/` on disk. The `lint-and-format-check` job in `.github/workflows/pr-health-checks.yml:54` runs `pnpm run lint` before `pnpm run build`, so the typecheck fails with TS2307. Fix by making the root lint depend on the library build first — e.g. add `"prelint": "pnpm run build:lib"` to `package.json`, or change `lint` to `"pnpm run build:lib && concurrently …"`, or alternatively split docs `typecheck` out of `lint:docs` and add a separate CI step that runs after `Check build`. Verify by deleting `library/dist/`, running `pnpm install --frozen-lockfile && pnpm run lint`, and confirming the docs typecheck resolves.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit wires `lint:docs` into the concurrent root `lint` script, which breaks the `lint-and-format-check` gate — `tsc --noEmit` in `docs/` imports `@tumaet/apollon` but `Run Lint` happens before `Check build` in CI, so it fails with `TS2307: Cannot find module '@tumaet/apollon'`. See inline. The three threads still open from previous rounds (`logger.ts` `base: null`, `release-library.yml` jq guard, `attw` positional in `pr-health-checks.yml`) are unchanged on this head; per the round-6 commitment I'm not re-pinging them, but the `attw` one is still what's red on `library-node22-compat`.

.github/workflows/pr-health-checks.yml
```diff
@@ -129,17 +147,73 @@ jobs:
...
+
+      - name: Validate type resolution (arethetypeswrong)
+        working-directory: library
+        # ESM-only library, all exports flagged as such; CJS profile
+        # would force-flag dual-package issues that don't apply. `pnpm
+        # dlx <pkg> [args]` forwards args straight to the resolved bin
+        # (`attw` is the package's default bin), so the `attw` token
+        # must NOT appear here — otherwise attw treats it as a tarball
+        # path argument.
+        run: pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only
```
Claudia-Anthropica: @FelixTJDietrich [high] The ENOENT is gone — dropping the `attw` positional was the right call — but the gate is still red, now with a different error: `error while checking file: Cannot read properties of undefined (reading 'filename')` (exit 3, run 26144822964). `attw --pack` delegates the tarball creation to pnpm and then crashes parsing pnpm 11's `pnpm pack --json` output — the same object-vs-array shape mismatch you just hardened `release-library.yml` against. Pack it yourself with the same type-guard and hand attw the tarball directly:

```yaml
run: |
  TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')
  pnpm dlx @arethetypeswrong/cli@0.18.1 "$TARBALL" --profile esm-only
```

In `.github/workflows/pr-health-checks.yml:208`, the step `pnpm dlx @arethetypeswrong/cli@0.18.1 --pack . --profile esm-only` fails with `Cannot read properties of undefined (reading 'filename')` because attw's `--pack` invokes `pnpm pack --json` and chokes on pnpm 11's output shape. Replace `--pack .` by packing first and passing the tarball path to attw: run `TARBALL=$(pnpm pack --json | jq -r 'if type=="array" then .[0].filename else .filename end')` then `pnpm dlx @arethetypeswrong/cli@0.18.1 "$TARBALL" --profile esm-only` (keep working-directory: library).

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Round 7 — the round-6 punch list all landed correctly (logger pid/hostname restored, the jq type-guard, docs typecheck moved post-build, lint-staged trimmed) and those threads are resolved, nice. But the attw fix uncovered a new failure that was hidden under the old ENOENT: the step now actually runs and crashes on pnpm's `pack --json` shape, so `library-node22-compat` is still red — see the inline fix. (Codacy is also failing but looks like the usual static-analysis noise on a diff this size.)

.github/workflows/pr-health-checks.yml
Claudia-Anthropica: @FelixTJDietrich Good call dropping it entirely — and thanks for digging into the root cause. My pre-pack suggestion wouldn't have helped since the bug is in `@arethetypeswrong/core`'s `extractTarball`, not the CLI arg handling, so handing it a tarball directly would've crashed the same way. The NOTE comment explaining why is exactly right so nobody reintroduces it, and `publint` + the exports-map resolution check cover the type-resolution ground. `library-node22-compat` is green now. Resolving.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback's addressed across the rounds — the attw gate is resolved by dropping it (with a clear NOTE on why, and publint + the exports-map check still covering type resolution), `lint:docs` no longer typechecks before the build, the React 19 matrix is gone with an honest peer-range narrowing, and the .docusaurus untracking landed. `library-node22-compat` and `lint-and-format-check` are green. Solid migration — thanks for the patience through the long review thread.

Claudia-Anthropica (approved):
@FelixTJDietrich Re-reviewed the two CI-fix commits since my approval. The Codacy hex-shorthand cleanup and the `vscode-controls.css` rule reorder (ascending specificity now, no cascade change) are harmless, and the `# hadolint ignore=DL3006` directive sits on the line immediately above each `FROM ${NODE_IMAGE}` in both Dockerfiles — which is exactly where hadolint needs it. Named-stage and tagged-nginx FROMs correctly left untouched. Still green, still approving.

docs/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "docusaurus": "docusaurus",
+    "start": "docusaurus start --no-open",
+    "build": "docusaurus build",
+    "swizzle": "docusaurus swizzle",
+    "clear": "docusaurus clear",
+    "serve": "docusaurus serve",
+    "typecheck": "tsc --noEmit",
+    "lint:md": "markdownlint-cli2 '{user,library,contributor,src}/**/*.{md,mdx}'",
+    "lint": "pnpm run lint:md && pnpm run typecheck"
```
Claudia-Anthropica: @FelixTJDietrich [high] This one regressed — `lint:md && typecheck` is back on the docs `lint` script, and it's red again on `lint-and-format-check` (and the PR Health Gate sitting in front of it). The docs `tsc --noEmit` can't resolve `@tumaet/apollon/react` because the root `lint` runs `lint:docs` concurrently *before* `library/dist/` is built, so the new `/react` subpath has no types yet — exactly the build-order issue from the round-7 thread that you'd fixed by dropping `typecheck` here. Simplest restore-to-green is to drop it again. If you want docs type-checking gated in CI, it has to run after `build:lib` (the way your `build:docs`/`dev:docs` scripts already do), not in the pre-build lint pass.

```json
    "lint": "pnpm run lint:md"
```

In `docs/package.json`, the `lint` script is `"pnpm run lint:md && pnpm run typecheck"`, but `tsc --noEmit` runs before `library/dist/` is built in CI and fails with TS2307 on `@tumaet/apollon/react`. Change the `lint` script back to `"pnpm run lint:md"` so the docs lint no longer typechecks against the unbuilt library.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The new `<Apollon>` React component is clean — the ref-snapshot pattern with an honestly-empty deps array is the right call, and the test pins the lifecycle contract precisely. The editor fixes all look like proper root-cause work too (body-style save/restore, fixed-position ghost, flow-coord midpoint drag, literal CSS fallbacks). Only blocker: `lint-and-format-check` is red because `typecheck` came back into the docs `lint` script and re-tripped the build-order issue from round 7 — details inline.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Re-reviewed the React-migration commits since my last pass and the new code is in great shape — the `<Apollon>` component (snapshot-vs-reactive prop split, re-key-to-remount, the onMount cleanup ordering) and the migration of `ApollonLocal` / `ApollonPlayground` / `ApollonWithConnection` and the VS Code editor onto it all look correct. The one thing still blocking is CI: `lint-and-format-check` is red because the docs `typecheck` still runs before the library build, so `@tumaet/apollon/react` (now imported in `ApollonEmbed`) can't resolve — same root cause as the open `docs/package.json` thread where `lint:md && typecheck` regrew. Drop `typecheck` from the concurrent docs `lint` (or sequence it after the build) and this is good to go.

standalone/webapp/src/pages/ApollonWithConnection.tsx
```diff
@@ -416,16 +383,13 @@ export const ApollonWithConnection: React.FC = () => {
         )
       }
     } else {
-      editor.setReadonly(baseReadonly)
       prePreviewFingerprintRef.current = null
```
Claudia-Anthropica, line 391: @FelixTJDietrich [high] The "kill more bloat" pass dropped `editor.setReadonly(baseReadonly)` from the preview-exit branch, but now nothing restores readonly. On preview entry you call `editor.setReadonly(true)` (line 373); on exit `setPreviewMode(false)` doesn't touch readonly (it only resyncs title/type), and the reactive `<Apollon readonly={baseReadonly}>` effect won't re-fire because both `baseReadonly` and the editor instance are unchanged. So after previewing a version and exiting in any editable view (Collaborate / Give-feedback / Modelling, where `baseReadonly` is false), the canvas stays locked read-only until a remount. Put the `setReadonly(baseReadonly)` line back at the top of the else branch.

In `standalone/webapp/src/pages/ApollonWithConnection.tsx`, the preview-exit branch (the `else` after `if (preview)`, around line 385) no longer resets readonly after preview entry set it to `true`, so the canvas stays stuck read-only on exit. Add `editor.setReadonly(baseReadonly)` as the first statement inside that `else` block, before `prePreviewFingerprintRef.current = null`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The `"use client"` banner fix (Rollup strips the source directive, so re-adding it via the peer build's output banner is the right call), the `destroy()` logging, and the comment/test trims all look fine — and I checked that the `defaultType` drop in the playground is safe since the editor falls back to `model.type`. One real regression though: the bloat-cut in `ApollonWithConnection` removed the readonly restore on preview exit, so the canvas gets stuck read-only after exiting a version preview — details inline. CI is also still red on the same docs `typecheck` build-order issue from the open `docs/package.json` thread (`@tumaet/apollon/react` can't resolve before the library builds).

docs/package.json
Claudia-Anthropica: @FelixTJDietrich Confirmed — `lint` is back to just `lint:md` and `lint-and-format-check` is green. The `docusaurus build` job still typechecks the site, so nothing's lost. Resolving.

standalone/webapp/src/pages/ApollonWithConnection.tsx
Claudia-Anthropica: @FelixTJDietrich Nice — `editor.setReadonly(baseReadonly)` is restored as the first statement in the exit branch, and the comment spells out exactly why the reactive prop can't re-fire here. Resolving.

library/README.md
```diff
@@ -3,63 +3,171 @@
...
 ## Install
 
+The package ships two builds with identical APIs:
+
+| Subpath                       | React / MUI / emotion / xyflow | Bundle                                  | Use when                                                                                          |
+| ----------------------------- | ------------------------------ | --------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| `@tumaet/apollon` _(default)_ | bundled                        | ~258 KB entry + ~2.1 MB lazy Yjs chunk¹ | Your host is Angular, Vue, Svelte, vanilla JS, or any framework that doesn't already have React.  |
+| `@tumaet/apollon/react`       | externalized (peer deps)       | ~875 KB                                 | Your host is React 18.3 and you want the editor to share React with your app instead of bundling. |
+
+¹ The Yjs sync layer is lazy-loaded only when `collaborationEnabled: true`.
```
Claudia-Anthropica: @FelixTJDietrich [medium] This footnote isn't accurate, and it's mirrored in `docs/library/overview.md:24`. Yjs isn't lazy here: `library/lib/store/diagramStore.ts` and `metadataStore.ts` both `import * as Y from "yjs"` statically, and the `ApollonEditor` constructor calls `new Y.Doc()` + `new YjsSync(...)` unconditionally — `collaborationEnabled` only gates `initializeUndoManager()`. There's no dynamic `import()` anywhere in `library/lib`, so that ~2.1 MB chunk loads eagerly for every editor, collab or not. The default subpath really costs ~2.3 MB; the old `~2.2 MB` line was the honest number. Drop the footnote and state it as one eager total, or keep the split but lose the "lazy / only when collaborationEnabled" wording.

In `library/README.md` (line 17) and `docs/library/overview.md` (line 24), the footnote "The Yjs sync layer is lazy-loaded only when `collaborationEnabled: true`" is false: `library/lib/store/diagramStore.ts` and `metadataStore.ts` statically import yjs, and the `ApollonEditor` constructor instantiates `new Y.Doc()` and `new YjsSync(...)` unconditionally, with no dynamic `import()` in `library/lib`. Remove the footnote and change the default-subpath bundle cell to a single eager total (e.g. `~2.3 MB`), since the Yjs code loads for every editor regardless of `collaborationEnabled`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich All three threads from last round are properly addressed — docs `lint` is back to `lint:md` (CI green again), `setReadonly(baseReadonly)` is restored on preview exit with a clear comment, and the peer range is honestly back to `^18.3.0`. Nice work. One new thing slipped into the "fix factual drift" pass: the README/overview bundle footnote claims Yjs is lazy-loaded only under `collaborationEnabled`, but it's eagerly loaded for every editor. Worth correcting before this hits npm since the README is the package's front page — details inline.

library/README.md
Claudia-Anthropica: @FelixTJDietrich Looks good now — footnote's gone and the default subpath reads as one honest eager ~2.4 MB total in both README and overview. That matches the source (Y.Doc + YjsSync instantiate unconditionally, no dynamic import in lib/). Nice fix.

Claudia-Anthropica (approved):
@FelixTJDietrich The bundle-table fix is correct — Yjs really does load eagerly (static `import * as Y from "yjs"` across lib/, unconditional `new Y.Doc()` + `new YjsSync(...)` in the constructor, no dynamic `import()` in lib/), so collapsing the split-with-footnote into one honest ~2.4 MB total in both README and overview is the right call. That clears the last open thread on my side and the rest of the PR is unchanged from prior rounds. Approving.

tamang29:
LGTM!
Tested standalone application, capacitor and vs-code extension.
Everything working as expected ✅

tamang29 (approved):

PR #701, opened 2026-05-20
feat(webapp): Capacitor mobile shell — origin allowlist, clipboard, iOS TestFlight workflow

tamang29:
### Checklist

### Motivation and Context

Apollon currently runs only as a web app. This PR adds the pieces needed to ship and operate it as a native iOS/Android shell built with [Capacitor](https://capacitorjs.com/):

1. The webapp can run inside a Capacitor `WebView` (origins `capacitor://localhost` and `ionic://localhost`) and still talk to the Apollon backend.
2. Copy-to-clipboard works natively on iOS/Android, with a graceful fallback in the browser.
3. The share/collaboration URL points at the deployed webapp (not the mobile shell's local origin), so links generated inside the app are openable anywhere.
4. iOS builds can be cut from CI to TestFlight without anyone running Xcode locally.

### Description

**Server (`standalone/server`)**

- `CORS_ORIGIN` now accepts a comma-separated allowlist. Production deployments add `capacitor://localhost,ionic://localhost` alongside the existing webapp origin so the mobile shell is allowed without weakening CORS (no wildcard, `credentials: true` preserved).

**Webapp (`standalone/webapp`)**

- Added Capacitor 8 (`@capacitor/core`, `@capacitor/ios`, `@capacitor/android`) plus the plugins used at runtime: `clipboard`, `share`, `filesystem`, `file-transfer`, `splash-screen`, `status-bar`, and `capacitor-plugin-safe-area`.
- `capacitor.config.ts` defines the app id, content inset, and status-bar behavior. The generated `ios/` and `android/` folders stay gitignored and are produced on demand.
- New `capacitor:*` scripts: `add:ios`, `add:android`, `sync`, `open:ios`, `open:android`, `assets:generate:ios`, `assets:generate:android`.
- `ShareModal` uses `@capacitor/clipboard` to copy the share URL, which works in the native shell and in the browser.
- The collaboration share URL is built from the public webapp origin instead of `window.location.origin`, so links shared from the mobile app open the web client.

**CI (`.github/workflows/ios-testflight-release.yml` + `standalone/webapp/fastlane/`)**

New manual-dispatch workflow that builds the iOS app from the latest sources and uploads it to TestFlight:

- Runs on a `[self-hosted, macOS]` runner (Xcode, CocoaPods, and `rbenv` provided by the runner).
- Builds the web bundle, regenerates the Capacitor iOS project, then runs Fastlane: `match` fetches the App Store signing cert + provisioning profile from the team's private match repository; `gym` builds a Release `.ipa`; `pilot` uploads it to TestFlight.
- Authenticates to App Store Connect with an API key (`.p8`) — no Apple ID / 2FA prompts in CI.
- Fastlane config lives at `standalone/webapp/{Gemfile,fastlane/Fastfile,fastlane/Appfile}` and is the only ground truth for the build (no `.xcodeproj` is committed).

The workflow is `workflow_dispatch` only for now so the first end-to-end run can be done manually and verified before considering an automatic trigger.

#### Required secrets & variables (one-time setup)

The workflow expects the following on the repository (or inherited from the org). Names mirror the existing iOS app in this organization so they can be reused as-is:

| Kind | Name | Notes |
|---|---|---|
| Secret | `IOS_APP_STORE_CONNECT_API_KEY_KEY_ID` | App Store Connect API Key ID |
| Secret | `IOS_APP_STORE_CONNECT_API_KEY_ISSUER_ID` | Issuer ID |
| Secret | `IOS_APP_STORE_CONNECT_API_KEY_PASSWORD` | `.p8` contents, base64-encoded |
| Secret | `IOS_MATCH_PASSWORD` | Passphrase for the match cert repo |
| Secret | `IOS_MATCH_GITLAB_AUTH` | `base64(user:token)` for HTTPS access to the match repo |
| Variable | `IOS_MATCH_GIT_URL` | URL of the team's match repository |

The App Store Connect API key must be a **Team** key with the **App Manager** role (Individual keys cannot create provisioning profiles). The App Store Connect listing for the iOS bundle id must exist before the first run.

### Steps for Testing

**Capacitor changes (local)**

1. `pnpm install`
2. `pnpm --filter @tumaet/webapp run build`
3. `pnpm --filter @tumaet/webapp run capacitor:add:ios` (first time only)
4. `pnpm --filter @tumaet/webapp run capacitor:sync`
5. `pnpm --filter @tumaet/webapp run capacitor:open:ios`, build and run in the iOS Simulator
6. Verify: collaboration session connects to the configured backend, share URL is copyable, share URL points at the web client

**Server CORS**

1. Set `CORS_ORIGIN=https://your-webapp.example,capacitor://localhost,ionic://localhost` in the server env
2. Hit a credentialed endpoint from each origin and confirm the preflight passes only for the allowlisted entries

**TestFlight workflow**

1. Add the secrets/variables listed above
2. Trigger `ios-testflight-release` from the Actions tab
3. Confirm the IPA artifact uploads and the build appears in App Store Connect → TestFlight after processing

### Screenshots

https://github.com/user-attachments/assets/09a7bc2e-c7af-4da5-8644-8e52fadd3993


standalone/webapp/package.json
```diff
@@ -18,6 +18,7 @@
     "test:visual:update": "playwright test tests/visual/ --update-snapshots"
   },
   "dependencies": {
+    "@capacitor/clipboard": "8.0.1",
```
Claudia-Anthropica, line 31: @tamang29 [medium] You added `@capacitor/clipboard` here but nothing imports it — `copyToClipboard` in `ShareModal.tsx` still calls `navigator.clipboard.writeText`. If the point was to make clipboard copy work inside the native WebView (where `navigator.clipboard` is unreliable / needs a secure context), wire up `Clipboard.write()` from this plugin. Otherwise drop the dependency so it doesn't ship as dead weight.

In `standalone/webapp/src/components/modals/ShareModal.tsx`, `copyToClipboard` uses `navigator.clipboard.writeText` while the newly added `@capacitor/clipboard` dependency is unused. Either replace the clipboard call with `Clipboard.write({ string: link })` from `@capacitor/clipboard` when running in the Capacitor app, or remove `@capacitor/clipboard` from `standalone/webapp/package.json` and `package-lock.json`.

</details>

standalone/webapp/src/components/modals/ShareModal.tsx
```diff
@@ -25,7 +27,13 @@ export const ShareModal = () => {
       const model = editor.model
       const { id: diagramID } = await DiagramApiClient.createDiagram(model)
 
-      const newurl = `${window.location.origin}/${diagramID}?view=${viewType}`
+      const isCapacitorApp =
+        isPlatform("ios") || isPlatform("android") || isPlatform("capacitor")
+
+      const newurl = isCapacitorApp
+        ? `${BACKEND_URL}/${diagramID}?view=${viewType}`
```
Claudia-Anthropica: @tamang29 [medium] This fixes the link for the initial share, but the "Recently shared Diagram" field (line 122) and its Copy Link button (line 127) still use `window.location.href`, which resolves to `capacitor://localhost/...` inside the app — the exact broken link this PR is trying to avoid. The success toast tells users they can re-access the link from the share menu, so that re-access path stays broken on Capacitor. Build those URLs from `BACKEND_URL` too when `isCapacitorApp`.

In `standalone/webapp/src/components/modals/ShareModal.tsx`, the readonly input value (line 122) and the Copy Link button onClick (line 127) use `window.location.href`, which is `capacitor://localhost/...` in the Capacitor app. Build the displayed/copied URL from `BACKEND_URL` plus the current diagram path when `isCapacitorApp` is true, mirroring the `newurl` logic.

</details>

standalone/webapp/src/services/DiagramApiClient.ts
```diff
@@ -24,10 +25,12 @@ interface RequestOpts {
   body?: unknown
   headers?: Record<string, string>
   signal?: AbortSignal
-  /** When true, sends/receives credentials (cookies). Default: true. */
   credentials?: RequestCredentials
```
Claudia-Anthropica, line 28: @tamang29 [medium] Beyond the `credentials` default, this file's changes look unrelated to Capacitor support — the commit also deleted all the JSDoc (the `permalink` rationale about preserving `view`, the `credentials` doc, the `total`/`headRev` docs, the section headers) and moved `ListVersionsResponse`/`RestoreVersionResponse` to the bottom of the file. All of that still exists on `main`, so it reads like an accidental rebase/revert. Please restore the docs and keep the diff to just the `credentials` change.

In `standalone/webapp/src/services/DiagramApiClient.ts`, restore the JSDoc comments and original layout from `origin/main` (the `permalink` rationale comment, the `credentials` doc, the `total`/`headRev` docs, the section-header comments, and the top placement of the `ListVersionsResponse`/`RestoreVersionResponse` interfaces). Keep only the intentional change to the `credentials` default for Capacitor.

</details>

standalone/webapp/src/services/DiagramApiClient.ts
```diff
@@ -24,10 +25,12 @@ interface RequestOpts {
...
 }
 
+const isCapacitorApp =
```
Claudia-Anthropica: @tamang29 [low] `isCapacitorApp` is already exported from `@/constants/urls`. Here, and again in `ShareModal.tsx`, you redefine the same `isPlatform(...)` check — three copies of the same logic. Import the exported one instead.

In `standalone/webapp/src/services/DiagramApiClient.ts`, remove the local `isCapacitorApp` constant and import `isCapacitorApp` from `@/constants/urls` instead. Apply the same change in `standalone/webapp/src/components/modals/ShareModal.tsx`.

</details>

standalone/webapp/.env.example
```diff
@@ -1,3 +1,18 @@
 # Optional: Override server URLs (defaults to location-relative if unset)
-# VITE_SERVER_URL=http://localhost:8000
-# VITE_SERVER_URL_WSS=ws://localhost:4444/ws
+VITE_SERVER_URL=http://localhost:8000
```
Claudia-Anthropica: @tamang29 [low] These were intentionally commented out before, and the header right above still says "defaults to location-relative if unset". Uncommenting them means anyone copying `.env.example` now hardcodes the server to `localhost:8000` for non-Capacitor builds too, which breaks that location-relative default. Re-comment them, or update the header to match the new behavior.

In `standalone/webapp/.env.example`, re-comment the `VITE_SERVER_URL` and `VITE_SERVER_URL_WSS` lines (prefix each with `# `) so the location-relative default described in the header comment still applies when the example is copied to `.env`.

</details>

Claudia-Anthropica (changes requested):
@tamang29 Nice work — the comma-split CORS parsing and the env-preset resolution for Capacitor are clean, and the platform detection is sound. A few things before this is ready: `@capacitor/clipboard` is added but never used (clipboard copy still goes through `navigator.clipboard`), the share modal's "Copy Link" path still emits `capacitor://` URLs, and `DiagramApiClient.ts` carries a chunk of unrelated churn that deleted JSDoc still present on main. Details inline.

standalone/webapp/package.json
Claudia-Anthropica: @tamang29 Nice — `Clipboard.write()` is wired up now, so clipboard copy works inside the native WebView. Resolving this one, thanks.

standalone/webapp/src/components/modals/ShareModal.tsx
Claudia-Anthropica: @tamang29 This actually went the wrong way. The latest commit dropped the `BACKEND_URL` branch, so now even the initial `newurl` (line 30) uses `window.location.origin`, and the "Recently shared Diagram" field (line 121) plus the Copy Link button (line 126) still use `window.location.href`. Inside the Capacitor app all three resolve to `capacitor://localhost/...`, so every link copied from this modal is unshareable — which is the exact bug this PR is meant to fix. Since `BACKEND_URL` is gone now, you'll need another source for the real web origin (a dedicated `VITE_` var, or deriving it from the configured server URL) to build the shared links.

standalone/webapp/src/services/DiagramApiClient.ts
Claudia-Anthropica: @tamang29 Still here — the JSDoc deletions (the `permalink` rationale, the `credentials`/`total`/`headRev` docs, the section-header comments) and moving `ListVersionsResponse`/`RestoreVersionResponse` to the bottom are all still in the diff, and they still exist on `main`. The file has also drifted from `main` (e.g. `fetchDiagram` lost its `signal` param relative to current main). Please rebase on `main` and keep this file's diff to just the `credentials` default.

standalone/webapp/src/services/DiagramApiClient.ts
Claudia-Anthropica: @tamang29 Still three copies of this check — the local const here, the inline `isPlatform(...)` in `ShareModal.tsx`, and the exported `isCapacitorApp` in `@/constants/urls`. And after the latest commit, that exported one isn't used anywhere anymore (urls.ts stopped resolving the server URL with it), so it's now dead. Either import the exported constant in both spots or drop it. (FWIW `isPlatform('capacitor')` is already true on both iOS and Android, so the `ios`/`android` checks are redundant.)

standalone/webapp/.env.example
Claudia-Anthropica: @tamang29 Still uncommented — the header right above still says "defaults to location-relative if unset", but `VITE_SERVER_URL`/`VITE_SERVER_URL_WSS` are active here, so anyone copying this to `.env` hardcodes the server to localhost. Re-comment them or update the header to match.

standalone/webapp/.env.example
```diff
@@ -1,3 +1,15 @@
...
+VITE_SERVER_URL_WSS=ws://localhost:4444/ws
+
+# Capacitor builds: select an environment preset (local | staging | prod)
+VITE_CAPACITOR_ENV=local
```
Claudia-Anthropica: @tamang29 [medium] After the latest commit gutted the preset resolution in `urls.ts`, nothing reads any of these `VITE_CAPACITOR_*` vars anymore — `serverURL`/`serverWSSUrl` now only look at `VITE_SERVER_URL`/`VITE_SERVER_URL_WSS`. So `VITE_CAPACITOR_ENV` and all six `VITE_CAPACITOR_URL_*`/`VITE_CAPACITOR_WSS_*` presets are dead config that will mislead anyone setting up a Capacitor build. Either wire the preset resolution back up in `urls.ts` or drop these lines.

In `standalone/webapp/.env.example`, the `VITE_CAPACITOR_ENV` and `VITE_CAPACITOR_URL_*`/`VITE_CAPACITOR_WSS_*` variables (lines 5-14) are no longer read by `standalone/webapp/src/constants/urls.ts` after its preset-resolution logic was removed. Remove these lines from `.env.example`, or reintroduce the preset-resolution logic in `urls.ts` that consumes them.

</details>

Claudia-Anthropica (changes requested):
@tamang29 Thanks for the update — the clipboard plugin is wired up correctly now, nice. But that's the only one of last round's five items that's addressed: the `DiagramApiClient` churn and the `.env.example` `VITE_SERVER_URL` lines are unchanged, and the `isCapacitorApp` duplication is still there (the exported copy is now unused). More importantly, removing the `BACKEND_URL` branch made the share-link issue worse — every link copied from the Share modal is now `capacitor://localhost/...` inside the app, which is the exact bug this PR is meant to fix, and the new `VITE_CAPACITOR_*` presets are now dead config. Replies inline on the open threads, plus one new note below.

standalone/webapp/src/services/DiagramApiClient.ts
Claudia-Anthropica: @tamang29 Reverting `DiagramApiClient` back to `main` sorted this — the JSDoc and the original layout are all back, and the file is now byte-identical to main. Resolving, thanks.

standalone/webapp/src/services/DiagramApiClient.ts
Claudia-Anthropica: @tamang29 The duplication's gone now too — the revert dropped the copy in `DiagramApiClient`, and `urls.ts` no longer exports `isCapacitorApp`, so the only check left is the inline one in `ShareModal`. No more dup, resolving. (Tiny aside, not blocking: `isPlatform('capacitor')` is already true on iOS/Android, so the `'ios'` arm in that one remaining check is redundant.)

Claudia-Anthropica (changes requested):
@tamang29 Thanks for the update — reverting `DiagramApiClient` and trimming `urls.ts` cleaned up two of the open threads (the unrelated churn and the `isCapacitorApp` duplication), both resolved now. Three things still open though: the Share modal still builds every link from `window.location` (line 30, plus the readonly input and Copy Link at 121/126), which is `capacitor://localhost/...` inside the app since `capacitor.config.ts` has no `server.url` — so shared collaboration links are unopenable, the exact thing this modal is for. And in `webapp/.env.example` the seven `VITE_CAPACITOR_*` lines are now dead config (nothing reads them after the `urls.ts` trim), while the uncommented `VITE_SERVER_URL`/`VITE_SERVER_URL_WSS` contradict the "defaults to location-relative if unset" header. See the existing open threads.

standalone/webapp/src/components/modals/ShareModal.tsx
Claudia-Anthropica: @tamang29 Took another look at the net diff against `main` and I owe you a correction here: these `window.location` lines (`newurl` on line 30, plus the readonly input and Copy Link at 121/126) are unchanged from `main` — this PR doesn't touch them. I latched onto it earlier because an intermediate commit added a `BACKEND_URL` branch that later got reverted, but measured against `main` it's a wash, so this PR neither introduces nor worsens it. It's a real gap for the in-app share flow (links come out as `capacitor://localhost/...`), but it's pre-existing and a better fit for its own follow-up rather than a blocker here. The CORS + clipboard changes this PR actually makes look good. Resolving this one — sorry for the back-and-forth.

Claudia-Anthropica (approved):
@tamang29 This is clean now. The CORS comma-split parsing handles the `capacitor://`/`ionic://` origins correctly, and the clipboard plugin makes copy work inside the WebView. `DiagramApiClient` and the `.env.example` files are back in sync with `main`, and pinning `@capacitor/clipboard` matches the rest of the `@capacitor/*` deps. The one thing I'd flagged on the in-app share links turns out to be pre-existing (`window.location` is identical on `main`), so that's a follow-up rather than a blocker for this PR. Approving — nice work.

standalone/webapp/src/components/modals/ShareModal.tsx
```diff
@@ -1,19 +1,23 @@
...
+import { useNavigate } from "react-router"
+import { toast } from "react-toastify"
+import { APButton } from "../APButton"
+import { serverURL } from "@/constants"
 
 export const ShareModal = () => {
   const { editor } = useEditorContext()
   const { closeModal, openModal } = useModalContext()
   const navigate = useNavigate()
+  const isCapacitorApp = isPlatform("ios") || isPlatform("android")
```
Claudia-Anthropica: @tamang29 [medium] `isPlatform("ios")`/`isPlatform("android")` are user-agent based, so this is `true` in a normal mobile browser too, not just the native app. A phone-browser user then takes the `serverURL` branch — and since `serverURL` defaults to `""` for location-relative deployments, the shared link becomes a relative `/${diagramID}?...&mobile=true` with no origin, i.e. broken. It's also inconsistent with `copyToClipboard` below, which uses `isPlatform("capacitor")`. Use `isPlatform("capacitor")` (it's already true on both iOS and Android native) and reuse the one constant in both spots.

```suggestion
  const isCapacitorApp = isPlatform("capacitor")
```

In `standalone/webapp/src/components/modals/ShareModal.tsx`, the `isCapacitorApp` constant uses `isPlatform("ios") || isPlatform("android")`, which is user-agent based and also true in mobile browsers (not only the native Capacitor app), causing the empty-`serverURL` branch to produce a broken relative share link. Change it to `const isCapacitorApp = isPlatform("capacitor")`, and replace the `isPlatform("ios") || isPlatform("capacitor")` check inside `copyToClipboard` with the same `isCapacitorApp` constant so both code paths use one consistent check.

</details>

standalone/webapp/src/components/modals/ShareModal.tsx
```diff
@@ -25,10 +29,13 @@ export const ShareModal = () => {
       const model = editor.model
       const { id: diagramID } = await DiagramApiClient.createDiagram(model)
 
-      const newurl = `${window.location.origin}/${diagramID}?view=${viewType}`
-      copyToClipboard(newurl)
-      closeModal()
+      const newurl = isCapacitorApp
+        ? `${serverURL}/${diagramID}?view=${viewType}&mobile=true`
```
Claudia-Anthropica: @tamang29 [low] `&mobile=true` isn't read anywhere — `ApollonWithConnection` only looks at `view`/`version`, and nothing on the server consumes it either. Either wire it up to whatever should branch on it, or drop it so shared links don't carry a no-op param.

In `standalone/webapp/src/components/modals/ShareModal.tsx`, the share URL appends `&mobile=true` but no code in `standalone/webapp/src` or `standalone/server/src` reads a `mobile` query parameter. Remove `&mobile=true` from the `newurl` template unless a consumer is added that reads it.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The new `fix: update collaboration sharing url for mobile` commit is the right idea for the native app, but the `isCapacitorApp` check fires in regular mobile browsers too, where `serverURL` is empty and the share link comes out relative/broken — see inline. Switching it to `isPlatform("capacitor")` fixes that and lines up with the clipboard check right below it. One smaller thing: `&mobile=true` isn't read anywhere yet.

standalone/webapp/src/components/modals/ShareModal.tsx
Claudia-Anthropica: @tamang29 `isPlatform("capacitor")` is the right call — it only fires in the native app now, so mobile browsers stay on the `window.location.origin` path. Looks good, thanks.

standalone/webapp/src/components/modals/ShareModal.tsx
Claudia-Anthropica: @tamang29 Good, the no-op `&mobile=true` param is gone now. Thanks.

Claudia-Anthropica (approved):
@tamang29 Both items from last round are sorted — `isPlatform("capacitor")` keeps the serverURL branch native-only (no more false positives in mobile browsers), and the no-op `mobile` param is gone. The CORS comma-split parsing and the clipboard plugin still look good. Approving, nice work.

standalone/webapp/fastlane/Fastfile
```diff
@@ -0,0 +1,76 @@
...
+      ipa: "./build/App.ipa",
+      api_key: api_key,
+      skip_waiting_for_build_processing: true
+    )
+  end
+
+  lane :api_key do
+    app_store_connect_api_key(
+      duration: 300,
+      key_id: ENV["API_KEY_ID"],
```
ravi-summai, line 70: is the key for "API_KEY_ID" correct here or should have been "IOS_APP_STORE_CONNECT_API_KEY_KEY_ID"?

.github/workflows/ios-testflight-release.yml
```diff
@@ -0,0 +1,89 @@
...
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+        with:
+          node-version-file: ".nvmrc"
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build web assets
+        run: pnpm --filter @tumaet/webapp run build
```
Claudia-Anthropica: @tamang29 [high] This workflow builds the webapp from a fresh checkout without building `@tumaet/apollon` first. The webapp imports `@tumaet/apollon/react` through the package export, which needs `library/dist`; hiding the generated `library/dist` locally makes `pnpm --filter @tumaet/webapp exec tsc -b` fail with `Cannot find module '@tumaet/apollon/react'`. Add `pnpm run build:lib` before this step, or use the same lib-then-webapp sequence as the Dockerfile, so the TestFlight job can produce the web bundle.

In `.github/workflows/ios-testflight-release.yml`, the `Build web assets` step runs `pnpm --filter @tumaet/webapp run build` before `library/dist` exists on a fresh runner. Fix this by running `pnpm run build:lib` before the webapp build, or replacing the step with a command that builds the library first and then `@tumaet/webapp`.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The mobile/runtime pieces still look fine and CI is green, but the new TestFlight workflow has a build-order bug. It needs to build the library before the webapp or the fresh runner won't resolve `@tumaet/apollon/react`, so I'm requesting changes on that inline.

standalone/webapp/fastlane/Fastfile
FelixTJDietrich: Good catch to double-check — the names look different on purpose. The GitHub secret is `IOS_APP_STORE_CONNECT_API_KEY_KEY_ID`, but the workflow remaps it to the shorter `API_KEY_ID` env var that the Fastfile reads (see [`ios-testflight-release.yml` line 67](https://github.com/ls1intum/Apollon/blob/feat/capacitor-server-config/.github/workflows/ios-testflight-release.yml#L67)). Same convention as `artemis-ios` so we can reuse the org-level secret as-is. Resolving.

.github/workflows/ios-testflight-release.yml
Claudia-Anthropica: @tamang29 Looks good now, thanks.

Claudia-Anthropica (approved):
@tamang29 The requested TestFlight build-order fix is in place now: the workflow builds the library before the webapp bundle. All feedback from my side is addressed, approving.

PR #706, opened 2026-05-22
feat: add line-jump on edge collision

tamang29:
### Checklist

### Motivation and Context

If the diagram is complex and has many relationship(edge) between nodes, it can be difficult to track the relationship between classes (nodes).

### Steps for Testing

1. Create a Class Diagram
2. Add nodes and edges
3. create edge collision either horizontally or vertically

### Screenshots
Before:
Screenshot 2026-05-22 at 12 54 33

After:
Screenshot 2026-05-22 at 12 55 04


library/lib/hooks/useStraightPathEdge.ts
```diff
@@ -105,6 +117,154 @@ export const useStraightPathEdge = ({
...
+        offset: 30,
+      })
+
+      const simplifiedPath = simplifySvgPath(edgePath)
+      return removeDuplicatePoints(parseSvgPath(simplifiedPath))
+    },
+    [allNodes, getNode, type]
+  )
+
+  const lineJumps = useMemo(() => {
```
Claudia-Anthropica: @tamang29 [medium] This memo re-scans every edge before this one and, for any edge without persisted `data.points`, rebuilds a full smooth-step path (`getSmoothStepPath` + parse + simplify) just to find intersections — so each edge does O(currentIndex) work and the diagram pays roughly O(N²) on every `edges` change. On top of that, `sourcePoint`/`targetPoint` (lines 122-129) are fresh object literals every render and they're in this memo's dep array, so `lineJumps` recomputes on literally every render, not only when geometry actually changes. For the "many relationships" case this PR targets, that's going to get janky fast. Resolve each edge's geometry once per `edges`/`nodes` snapshot into a shared memoized map, and depend on primitive coords instead of the literal objects.

In `library/lib/hooks/useStraightPathEdge.ts`, the `lineJumps` useMemo recomputes on every render because `sourcePoint`/`targetPoint` are object literals recreated each render yet listed as dependencies, and it rebuilds every other edge's smooth-step path from scratch (O(N²) across the diagram). Replace the object-literal dependencies with the primitive `adjustedSourceCoordinates`/`adjustedTargetCoordinates` numbers, and memoize the per-edge resolved geometry (keyed by edge id and node positions) in a shared structure so other edges' paths are not recomputed inside every edge.

</details>

library/lib/hooks/useStepPathEdge.ts
```diff
@@ -372,9 +379,160 @@ export const useStepPathEdge = ({
...
+        offset: 30,
+      })
+
+      const simplifiedPath = simplifySvgPath(edgePath)
+      return removeDuplicatePoints(parseSvgPath(simplifiedPath))
+    },
+    [allNodes, getNode, type]
+  )
+
+  const lineJumps = useMemo(() => {
```
Claudia-Anthropica: @tamang29 [medium] Same O(N²) concern as the straight-path hook: this runs on every `edges` mutation, and for each prior edge without `data.points` it recomputes a full `getSmoothStepPath` (plus simplify/parse) only to test for crossings. With many edges, every edge re-derives every other edge's geometry on each store update. Compute and cache the resolved segments per edge once (keyed by edge id + node positions) and share that across edges instead of recomputing inside each one.

In `library/lib/hooks/useStepPathEdge.ts`, the `lineJumps` useMemo recomputes every other edge's smooth-step path from scratch on each `edges` change, giving O(N²) cost. Extract the per-edge geometry resolution into a shared memoized cache keyed by edge id and source/target node positions so each edge's segments are computed once per snapshot rather than recomputed by every edge.

</details>

library/lib/utils/edgeUtils.ts
```diff
@@ -457,6 +457,423 @@ export function getEllipseHandlePosition(
...
+  index: number
+  start: IPoint
+  end: IPoint
+  orientation: "horizontal" | "vertical"
+  fixed: number
+  min: number
+  max: number
+}
+
+export function getAxisAlignedSegments(
```
Claudia-Anthropica, line 1003: @tamang29 [medium] The new helpers (`getAxisAlignedSegments`, `findLineJumpIntersections`, `buildPathWithLineJumps`, `isStraightEdgeType`, `getHandlePositionOnNode`) carry all the tricky logic of this feature — margin clamping, orientation handling, jump spacing/dedup, the direction-sign math — but there are no unit tests, even though `edgeUtils.test.ts` already covers the sibling pure functions. Please add tests pinning down the intersection and margin math; it's exactly the kind of geometry that breaks silently on a refactor.

In `library/tests/unit/edgeUtils.test.ts`, add unit tests for the new functions in `library/lib/utils/edgeUtils.ts`: `getAxisAlignedSegments`, `findLineJumpIntersections`, `buildPathWithLineJumps`, `isStraightEdgeType`, and `getHandlePositionOnNode`. Cover horizontal/vertical crossings, the margin/endpoint exclusion, minimum-segment-length skipping, jump spacing/dedup, and the use-case ellipse branch.

</details>

library/lib/utils/edgeUtils.ts
```diff
@@ -457,6 +457,423 @@ export function getEllipseHandlePosition(
...
+          orientation: base.orientation,
+        })
+      }
+    }
+  }
+
+  return hits
+}
+
+export function findLineJumpIntersection(
```
Claudia-Anthropica: @tamang29 [low] `findLineJumpIntersection` (singular) here and `buildPathWithLineJumpAtPoint` just below it are exported but nothing in the codebase calls them — the hooks use the plural `findLineJumpIntersections` and `buildPathWithLineJumps`. Drop them unless they're meant for a follow-up.

In `library/lib/utils/edgeUtils.ts`, remove the unused exported functions `findLineJumpIntersection` and `buildPathWithLineJumpAtPoint`, which have no callers anywhere in the repository.

</details>

Claudia-Anthropica (changes requested):
@tamang29 Really nice feature — line-jumps make crossing edges much easier to read, and the before/after screenshots show it off well. Two things hold it back though: the jump detection is pretty expensive in exactly the large-diagram case it's built for (it re-derives every other edge's geometry on each store change, and the straight-path hook recomputes on every render), and there are no unit tests for the new geometry helpers. Details inline.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @tamang29 Nice — the shared `getEdgeGeometryMap` with the edges/nodes WeakMap means the geometry is built once per store snapshot and reused across all edges instead of every edge rebuilding every other edge's path, and the deps are now primitive coords so it no longer recomputes on every render. Exactly what was needed. Resolving.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @tamang29 Same fix carries over here too — the step-path hook now pulls from the shared geometry map plus the per-edge-id cache, so no more O(N²) path rebuilds on each `edges` change. Looks good.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @tamang29 Tests look solid — they pin down the horizontal/vertical crossings, the endpoint-margin exclusion, the jump dedup spacing, and the use-case ellipse branch. Ran them locally, all green. Thanks for adding these.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @tamang29 Both unused exports are gone now. Thanks.

Claudia-Anthropica (approved):
@tamang29 All my earlier feedback is addressed — the geometry is now resolved once per store snapshot via a shared `getEdgeGeometryMap` cache (no more per-edge O(N²) path rebuilds or recompute-every-render), the new geometry helpers have proper unit tests, and the dead exports are gone. Verified the code and ran the library tests locally (162 passing). Nice work on the line-jump feature.

FelixTJDietrich (approved):
Love it, works as expected :D  @FadyGergesRezk please check if those changes collide with your edge work. 

Image

library/lib/utils/edgeUtils.ts
```diff
@@ -988,6 +991,364 @@ export function getEllipseHandlePosition(
...
+  const round = (num: number) => Math.round(num)
+  const jumpsBySegment = new Map<number, LineJumpHit[]>()
+
+  for (const jump of jumps) {
+    const list = jumpsBySegment.get(jump.segmentIndex) ?? []
+    list.push(jump)
+    jumpsBySegment.set(jump.segmentIndex, list)
+  }
+
+  const pathParts = [`M ${round(points[0].x)},${round(points[0].y)}`]
```
Claudia-Anthropica: @tamang29 [medium] As soon as a step edge gets a line jump, this switches its `d` from the usual `pointsToSvgPath` shape (`M x y`) to `M x,y`. That is what breaks the current `edge-interactions` e2e run after reconnecting a crossed class edge: the edge still exists, but the test can no longer parse its source point. Please emit jumped step paths in the same coordinate format as `pointsToSvgPath` and update the new unit expectations accordingly.

In `library/lib/utils/edgeUtils.ts`, `buildPathWithLineJumps` emits comma-separated coordinates for jumped step-edge paths, which makes the current edge reconnection e2e test fail once a class edge receives a line jump. Change `buildPathWithLineJumps` to emit the same space-separated coordinate format as `pointsToSvgPath` and update the affected unit test expectations.

</details>

library/lib/utils/edgeUtils.ts
```diff
@@ -988,6 +991,364 @@ export function getEllipseHandlePosition(
...
+      sourceY: adjustedSource.sourceY,
+      sourcePosition: source.side,
+      targetX: adjustedTarget.targetX,
+      targetY: adjustedTarget.targetY,
+      targetPosition: target.side,
+      borderRadius: EDGES.STEP_BORDER_RADIUS,
+      offset: 30,
+    })
+
+    const points = removeDuplicatePoints(parseSvgPath(simplifySvgPath(edgePath)))
```
Claudia-Anthropica: @tamang29 [low] CI also stops at the format step because Prettier still wants to wrap this call, and it reports the same kind of wrapping in the new geometry-map test. Please run Prettier on the changed files so `lint-and-format-check` can continue to build/typecheck/tests.

In `library/lib/utils/edgeUtils.ts` and `library/tests/unit/edgeUtils.test.ts`, Prettier reports formatting differences. Run Prettier on both changed files and commit the formatted output.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The line-jump implementation is close, but CI is red on this SHA: the edge reconnection e2e test fails once a class edge gets a jumped path, and Prettier is failing on two changed files. Please fix those before merge.

library/lib/utils/edgeUtils.ts
FelixTJDietrich: Fixed in 95bc67a0. `buildPathWithLineJumps` now emits the same space-separated `M x y` format as `pointsToSvgPath`, so a step edge's `d` keeps the same shape whether or not it carries bridges. The unit expectations were updated and the `edge-interactions` e2e passes again (including reconnecting a crossed class edge). Beyond this, line-jump crossing geometry now reads each edge's *actual* rendered points from a shared runtime registry (`edgeGeometryStore`) rather than re-deriving them, so bridges stay centered on the real crossing for every edge type.

library/lib/utils/edgeUtils.ts
FelixTJDietrich: Fixed in 95bc67a0. Ran Prettier on the changed files — `pnpm run format:check` is clean and `pnpm run lint` reports 0 errors across all packages, so `lint-and-format-check` passes. (The specific `getEdgeGeometryMap` call this was anchored to was also removed in the geometry-registry refactor.)

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FelixTJDietrich Nice, that handles it.

Claudia-Anthropica (approved):
@tamang29 All feedback is addressed and CI is green on the new SHA. Approving this version.

PR #708, opened 2026-05-22
fix: make canvas unmovable while clicking on NodeToolbar

tamang29:
### Checklist

### Motivation and Context

In capacitor app, while trying to open edit menu it moves the canvas instead making it tough to use the functionality.

### Steps for Testing

1. Create a new diagram
2. Drag "Class" element
3. Click on element -> NodeToolber will popup -> hold pressing on "Edit" or "Delete" and move the cursor
4. The canvas will move

### Screenshots
Before:

https://github.com/user-attachments/assets/ad111196-5e3f-46df-86ce-d764669ece46

https://github.com/user-attachments/assets/5f051c3a-cb0d-4f74-b339-e1a654c0ed54


Claudia-Anthropica (approved):
@tamang29 Clean, targeted fix. Adding the `nodrag nopan` classes is the canonical React Flow way to stop the canvas panning here (the toolbar portals into `.react-flow__renderer`, so the d3-zoom filter picks the classes up), and the capture-phase `stopPropagation` handlers are a sensible extra guard for the Capacitor/touch case the standard class doesn't fully cover. The Edit/Delete buttons still work since `click` is separate from the stopped down-events. Nice one.

PR #710, opened 2026-05-23
feat(edges): better step bending, waypoint drag, and React Flow reconnection

FadyGergesRezk:
### Checklist

### Motivation and Context

Edges were the roughest part of the editor. Step edges bent unpredictably, manually placed waypoints were lost when an endpoint was reconnected, endpoint reconnection used bespoke logic that diverged from React Flow's own behaviour, and the half-circle "arc" connection handles could overlap each other on small nodes.

This PR makes edge routing, interaction, and connection handles behave predictably and look clean. It also keeps the collaborative document small: only the *inputs* to an edge (its endpoints, manually placed waypoints, and handles) are synced through Yjs — never the computed pixel geometry — so peers don't thrash shared state or drift apart.

### Description

**Edge bending & waypoints**
- Step-edge bending is consolidated and predictable: each draggable bend handle moves exactly one segment, snapped to the 5px grid, and the rest of the path stays put.
- An in-progress drag is now driven through React state, so the path, the dragged handle, the edit toolbar, and per-type decorations all follow the drag **live** instead of jumping only on release.
- Bends survive endpoint reconnection where the geometry allows, and clamp to the last valid shape (rather than snapping all the way back) when dragged past a limit.

**Endpoint reconnection**
- Adopts React Flow's native reconnect API in place of the previous custom flow, with a connection-line preview. The opposite endpoint stays anchored while you drag the other end.

**Connection handles (9-slot model)**
- Each side exposes a 9-slot offset model: five *named* handle IDs at the even slots (corner / mid-corner / middle / mid-corner / corner) render the visible arcs, and four *between* IDs at the odd slots stay addressable as hidden anchors so saved edges still resolve when a node is resized between layout stages.
- Arc draggers are laid out in symmetric stages by side length — **1** centred arc on short sides, **3** at corners + middle, up to **5** evenly spaced — so they never visually overlap, including when zoomed out. All legacy handle IDs remain addressable for backwards-compatible diagrams.
- Handles, endpoint hit targets, and the edit toolbar keep a predictable on-screen size across zoom; the canvas/handle snap grid moved from 10px → 5px so connection points stay grid-aligned.

**Decorations**
- The SFC transition bar + condition label sit exactly on the edge centre (standard notation) and follow the edge while it is dragged; decorative SVG is pointer-transparent so it can't steal a bend handle's grab.

**Sync & migration**
- Computed geometry is no longer persisted; only edge inputs are synced.
- `EdgeTransformer` strips stale runtime-only geometry from imported diagrams so older models open cleanly.

**Tests**
- ~1.5k lines of unit + Playwright E2E coverage for the bend FSM, orthogonal point preservation, the 9-slot/5px-grid handle contract, zoom-adaptive handles, endpoint markers, migration, and the decoration/reconnect behaviour.

> **Scope note (transparency):** an experimental worker-backed orthogonal *auto-routing* engine (A\* + visibility graph + spatial index) was prototyped earlier on this branch and then **removed** before merge — it added significant complexity for little benefit over React Flow's step routing plus manual waypoints. This PR ships the simpler, predictable approach described above; the routing engine is **not** part of it.

### Steps for Testing

1. Draw an edge between two nodes — verify it renders as a clean orthogonal step path.
2. Drag one of the nodes — verify the path updates smoothly without flicker.
3. Drag a bend handle on a step edge — verify only that segment moves, it snaps to the grid, and the edit toolbar/labels follow the drag live (not just on release).
4. Drag an edge endpoint to a handle on the same node, then to a different node — verify the opposite endpoint stays anchored and manual bends are preserved when geometry allows.
5. Drop a **Use Case** association/ellipse and drop/reconnect an edge near it — verify the edge always attaches to a visible handle and never disappears.
6. Open a diagram saved with an older Apollon version — verify edges render correctly with no manual fixes.
7. With two tabs collaborating via Yjs, move nodes in one tab — verify the other re-routes locally with no edge geometry thrashing through the shared doc.
8. Drop a small Class node and grow it — verify the side shows **1** centred arc, then **3** (corners + middle), then **5** evenly-spaced arcs, never overlapping at any size or zoom.
9. Add an SFC condition to a transition — verify the crossbar + label sit on the edge centre and follow it while bending.
10. Drag a node on the canvas — verify it snaps to 5px increments.
11. Run `pnpm test` in `library/` and `pnpm test:visual` + `pnpm test:e2e` in `standalone/webapp/` — all suites pass.

### Screenshots

[62145e79-09c9-46b3-b4aa-87b5bb17de0d (1).webm](https://github.com/user-attachments/assets/db9503bb-fd39-41da-b43c-49e1f67906b1)
[3b848df3-f0c2-4fa2-91a5-3646f3ca9e4e.webm](https://github.com/user-attachments/assets/7d127f6d-06d5-4a3c-adea-5c628111977d)


library/lib/edges/interactions/useEdgeInteractions.ts
```diff
@@ -0,0 +1,494 @@
...
+        stopGuidanceRef.current()
+      }
+      window.removeEventListener("pointermove", handlePointerMove)
+      window.removeEventListener("pointerup", handlePointerUp)
+      window.removeEventListener("pointercancel", handlePointerUp)
+      cancelAnimationFrame(animationFrameId)
+    }
+    // Only re-install the listeners when the drag starts/stops; pointer
+    // position and node updates feed through refs.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Claudia-Anthropica: @FadyGergesRezk [high] This `eslint-disable-next-line react-hooks/exhaustive-deps` is one of the 49 errors failing `lint-and-format-check`: the library's `eslint.config.js` never registers `eslint-plugin-react-hooks`, so that rule doesn't exist and ESLint errors on the unknown-rule directive ("Definition for rule 'react-hooks/exhaustive-deps' was not found"). Worth knowing: it also means exhaustive-deps has never actually been enforced in this workspace, so the directive is suppressing nothing. Just drop the comment — the intentional `[state.value, edgeId]` deps array stands on its own since everything else is read through refs.

In `library/lib/edges/interactions/useEdgeInteractions.ts`, remove the `// eslint-disable-next-line react-hooks/exhaustive-deps` comment on line 484 because the library's eslint config does not register eslint-plugin-react-hooks, so the directive references an undefined rule and errors. Then run `npm run lint --workspace=@tumaet/apollon` and fix every remaining error: replace each `@typescript-eslint/no-explicit-any` with a concrete type and remove the unused variables flagged in bendHandles.ts and the test files.

</details>

library/lib/utils/geometry/SpatialIndex.ts
```diff
@@ -0,0 +1,140 @@
...
+// Singleton factory for the main thread
+// ---------------------------------------------------------------------------
+
+let _mainThreadIndex: SpatialIndex | null = null
+
+/**
+ * Returns a module-level singleton SpatialIndex instance for the main
+ * thread. Call `rebuildSpatialIndex()` whenever the node layout changes.
+ */
+export function getMainSpatialIndex(): SpatialIndex {
```
Claudia-Anthropica: @FadyGergesRezk [low] `getMainSpatialIndex` and `rebuildSpatialIndex` (plus the `_mainThreadIndex` singleton) have no callers anywhere — the worker builds its own `SpatialIndex` inside `buildOVG`, and nothing on the main thread uses these. Looks like a main-thread index that was planned but never wired up. Either hook `rebuildSpatialIndex` into the node-change subscription it was meant for, or drop the dead code so it doesn't bit-rot.

In `library/lib/utils/geometry/SpatialIndex.ts`, remove the unused `getMainSpatialIndex` function, `rebuildSpatialIndex` function, and the `_mainThreadIndex` module-level variable, since they have no callers in the codebase. Verify with a repo-wide search for both names before deleting.

</details>

library/lib/services/migration/EdgeTransformer.ts
```diff
@@ -0,0 +1,126 @@
...
+// ---------------------------------------------------------------------------
+// Single Edge Transformer
+// ---------------------------------------------------------------------------
+
+/**
+ * Hydrates a single edge's data object with OrthogonalEdgeData defaults.
+ * Returns a new edge object (does NOT mutate the input).
+ */
+export function hydrateEdgeData(edge: ApollonEdge): ApollonEdge {
+  const hasComputedSegments = Object.prototype.hasOwnProperty.call(
```
Claudia-Anthropica, line 20: @FadyGergesRezk [low] `hydrateEdgeData` assumes `edge.data` exists — `Object.prototype.hasOwnProperty.call(edge.data, ...)` here (and `edge.data.userWaypoints` below) throws `TypeError: Cannot convert undefined or null to object` if it's missing. `isV4Format` only checks `Array.isArray(data.edges)`, so a v4 payload with an edge that lacks `data` now crashes the whole import — and `validateEdgeMigration` already handles the `!data` case, so the two are inconsistent. Add an early guard.

In `library/lib/services/migration/EdgeTransformer.ts`, `hydrateEdgeData` dereferences `edge.data` without a null check, which throws when an imported edge has no `data`. At the top of the function add a guard: if `edge.data` is null or undefined, return `{ ...edge, data: { userWaypoints: [], routingMode: 'auto' } }`, mirroring how `validateEdgeMigration` already tolerates missing data.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk Seriously impressive work under the hood — the A* router with a binary min-heap, the orthogonal visibility graph, the R-tree spatial index, the routing worker, the edge FSM, the keep-geometry-out-of-Yjs design, and ~1.5k lines of tests are all well thought through. The blocker is purely CI: `lint-and-format-check` fails with 49 ESLint errors and `visual-regression-tests` fails. Most lint errors are `@typescript-eslint/no-explicit-any` in production files (OrthogonalEdge.tsx, resolveEndpointDragState.ts, store/memoization.ts, store/routingStore.ts, workers/routing.worker.ts) plus a few unused vars (bendHandles.ts, a couple of tests) — please give those real types and run `npm run lint` until it's clean. The visual-regression job fails only on the Observer template (~0.02 px diff); if that's the intended new edge rendering, refresh the baseline, otherwise it's worth a look since the other 35 templates passed. A few small things inline.

library/lib/edges/interactions/useEdgeInteractions.ts
Claudia-Anthropica: @FadyGergesRezk The disable comment is gone and `npm run lint` / `format:check` both pass clean now. Resolving this one, thanks.

library/lib/utils/geometry/SpatialIndex.ts
Claudia-Anthropica: @FadyGergesRezk Still open — `getMainSpatialIndex`, `rebuildSpatialIndex`, and the `_mainThreadIndex` singleton still have no callers anywhere (the worker builds its own index in `buildOVG`). Not blocking, but worth dropping before merge so it doesn't bit-rot.

library/lib/services/migration/EdgeTransformer.ts
Claudia-Anthropica: @FadyGergesRezk Still unguarded — `hydrateEdgeData` dereferences `edge.data` via `Object.prototype.hasOwnProperty.call(edge.data, ...)`, which throws on a v4 edge that has no `data`. Low priority, but it's a one-line early return and keeps it consistent with `validateEdgeMigration`.

library/lib/utils/geometry/bendHandles.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  return {
+    x: (start.x + end.x) / 2,
+    y: (start.y + end.y) / 2,
+  }
+}
+
+export function getBendableSegments(
+  points: IPoint[],
+  sourcePosition: Position,
```
Claudia-Anthropica: @FadyGergesRezk [critical] This commit dropped the pass-through of `sourcePosition`/`targetPosition` into `getBendHandlePosition`, but they're still declared as params on `getBendableSegments` and now unused — so `tsc -b` fails with `TS6133: 'sourcePosition' is declared but its value is never read` (and the same for `targetPosition` on the next line). That breaks `npm run build`, which fails the unit-tests CI job at the build step. Easiest fix is to prefix both with `_` (mirroring what `getBendHandlePosition` already did), or drop them and update the lone caller in useStepPathEdge.ts:340.

In `library/lib/utils/geometry/bendHandles.ts`, the `getBendableSegments` function declares `sourcePosition` and `targetPosition` parameters that are no longer used in its body, causing `tsc -b` to fail with TS6133. Rename both parameters to `_sourcePosition` and `_targetPosition` so the unused-parameter check passes, then run `npm run build --workspace=@tumaet/apollon` to confirm the build succeeds.

</details>

library/tests/unit/OrthogonalEdge.test.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  const baseProps = {
+    type: "OrthogonalRouted",
+    source: "s1",
+    target: "t1",
+    sourceX: 0,
+    sourceY: 0,
+    targetX: 0,
+    targetY: 0,
+    sourcePosition: Position.Right,
```
Claudia-Anthropica: @FadyGergesRezk [critical] Swapping `"right" as any` for `Position.Right` means this test now imports `Position` from `@xyflow/react`, but the `vi.mock("@xyflow/react", ...)` factory at the top of the file only stubs `useReactFlow`. So `Position` is undefined, the whole suite fails to load (`No "Position" export is defined on the "@xyflow/react" mock`), 0 tests run from this file, and `npm run test` exits 1 — failing the unit-tests CI job. Add `Position` to the mock factory.

In `library/tests/unit/OrthogonalEdge.test.tsx`, the `vi.mock("@xyflow/react", () => ({ ... }))` factory does not export `Position`, but the test imports and uses `Position.Right`/`Position.Left`, so the suite fails to load. Add `Position: { Left: "left", Right: "right", Top: "top", Bottom: "bottom" }` to the object returned by that mock factory, then run `npm run test --workspace=@tumaet/apollon -- OrthogonalEdge` to confirm it passes.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk Thanks for the quick turnaround — the eslint-disable is gone and `npm run lint` / `npm run format:check` are both clean now, so the original lint blocker is sorted. But the type-safety cleanup introduced two new failures in this commit: `tsc -b` now fails on unused params in bendHandles.ts (so `npm run build` and everything downstream of it, including the visual-regression job, can't run), and OrthogonalEdge.test.tsx no longer loads because the `@xyflow/react` mock doesn't export `Position`. Both are quick — details inline. The two earlier low-priority threads (dead SpatialIndex code, EdgeTransformer null guard) are still open but non-blocking.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The only new commit since my last review is the merge from `main`, which doesn't touch any of the open items — so the previous blockers still stand. The build still fails (`tsc -b` → TS6133 on the now-unused `sourcePosition`/`targetPosition` params in `bendHandles.ts`), which is exactly what's failing `lint-and-format-check`, and the `OrthogonalEdge.test.tsx` suite still can't load because `Position` is missing from the `@xyflow/react` mock factory. The two low-priority items (dead `SpatialIndex` helpers, missing `edge.data` guard in `hydrateEdgeData`) are also still open. All details are in the existing inline threads — no new findings this round.

library/lib/utils/geometry/bendHandles.ts
Claudia-Anthropica: @FadyGergesRezk Nice — the `_sourcePosition`/`_targetPosition` rename clears the `TS6133` and the build's green again. Resolving.

library/tests/unit/OrthogonalEdge.test.tsx
Claudia-Anthropica: @FadyGergesRezk The bendHandles fix landed and the build's green now 👍 — but this one's still red on the latest CI run: `No "Position" export is defined on the "@xyflow/react" mock`. The `vi.mock("@xyflow/react", ...)` factory at the top of the file still only returns `useReactFlow`, so `Position.Right` on line 98 resolves to undefined and the whole suite fails to load. Spread the real module in the mock so the enum survives:

```ts
vi.mock("@xyflow/react", async (importOriginal) => ({
  ...(await importOriginal<typeof import("@xyflow/react")>()),
  useReactFlow: () => ({ /* ...existing stubs... */ }),
}))
```

library/lib/utils/alignmentUtils.ts
```diff
@@ -2,7 +2,15 @@ import { Node } from "@xyflow/react"
...
-const ALIGNMENT_THRESHOLD = 10 // pixels within which guides appear
+// Half the grid step: guides appear once a node sits within this many pixels
+// of an alignment line.
+//
+// Hard-coded to avoid a known circular import:
+//   constants.ts → @/nodes → nodes/.../Class.tsx → @/utils (barrel) → alignmentUtils.ts → @/constants
+// Reading CANVAS.SNAP_TO_GRID_PX at module init can resolve to undefined when
+// alignmentUtils.ts evaluates before constants.ts finishes. Keep this in sync
+// with CANVAS.SNAP_TO_GRID_PX / 2 in constants.ts (currently 10 / 2 = 5).
+const ALIGNMENT_THRESHOLD = 5
```
Claudia-Anthropica, line 13: @FadyGergesRezk [high] Tightening `ALIGNMENT_THRESHOLD` from 10 to 5 breaks the existing `alignmentUtils.test.ts` ("targets top-level and contained nodes when dragging a package", line 199): it expects a vertical guide at 305 when the dragged package's left edge sits at 300, but `|300 - 305| = 5` and the check is `Math.abs(...) < threshold` (strict), so a 5px gap no longer matches and the test fails CI. Either update that test to match the new 5px threshold, or reconsider whether 5 is too tight for snapping.

In `library/lib/utils/alignmentUtils.ts`, lowering `ALIGNMENT_THRESHOLD` to 5 broke `library/tests/unit/alignmentUtils.test.ts` at line 199 — a 5px gap no longer satisfies the strict `Math.abs(...) < threshold` comparison. Update the assertion in that test ("targets top-level and contained nodes when dragging a package") so the vertical guide at position 305 is no longer asserted true, or switch the comparison to `<=` / widen the threshold if a 5px gap should still match.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The `bendHandles` `TS6133` is sorted and the build's green again — thanks for the quick turnaround. Two unit tests still fail `lint-and-format-check` though: the `OrthogonalEdge` mock still doesn't export `Position` (existing thread), and the new `ALIGNMENT_THRESHOLD` 10->5 change breaks `alignmentUtils.test.ts` (see inline). The visual-regression baselines are also still ~2% off in CI even after the PNG updates, which usually means they need regenerating inside the Docker/CI environment rather than locally. The two low-priority items (dead `SpatialIndex` helpers, unguarded `hydrateEdgeData`) are still open but non-blocking.

library/tests/unit/OrthogonalEdge.test.tsx
Claudia-Anthropica: @FadyGergesRezk This whole test file is gone now that the latest commit removed `OrthogonalEdge`, so the `Position` mock issue is moot. Resolving.

library/lib/utils/alignmentUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still failing on the latest commit — I ran `alignmentUtils.test.ts` and line 199 still fails because `ALIGNMENT_THRESHOLD` is still 5, so the 5px gap no longer satisfies the strict `Math.abs(...) < threshold` check. The 'remove orthogonal routed edge' commit didn't touch this. Fix is in my comment above (update the assertion or switch to `<=`/widen the threshold).

library/lib/App.tsx
```diff
@@ -33,16 +40,33 @@ import { useDiagramModifiable } from "./hooks/useDiagramModifiable"
...
+      "x" in point &&
+      "y" in point &&
+      typeof point.x === "number" &&
+      typeof point.y === "number"
+  )
 
 function App({ onReactFlowInit }: AppProps) {
   useKeyboardShortcuts()
 
+  useEffect(() => retainRoutingWorker(), [])
```
Claudia-Anthropica: @FadyGergesRezk [medium] Removing `OrthogonalEdge`/`useOrthogonalRoute` took out the only callers of `routingStore.calculateRoute`, so the whole orthogonal routing engine is now unreachable: nothing calls `calculateRoute`, this `retainRoutingWorker()` only bumps a lease counter (it never even initializes the worker), and `AStarRouter`/`OrthogonalVisibilityGraph`/`SpatialIndex`/`edgeObstacles` run only from the worker (which never starts) or their own tests. The `routing.worker` chunk still ships in the bundle (~11 kB), and `routingMode` is now write-only — `EdgeTransformer` sets it but nothing reads it. Either rewire the renderer to consume `calculateRoute`, or drop the dead engine + its tests (this subsumes the open `SpatialIndex` thread) and refresh the PR description, which still describes the now-removed `OrthogonalEdge`/`Segment`/`useOrthogonalRoute`.

In the Apollon library, commit f47a83c2 removed the orthogonal edge renderer but left its routing engine unreferenced. Confirm `routingStore.calculateRoute`/`useRoutingStore` have no callers, then remove the dead routing subsystem (`lib/store/routingStore.ts`, `lib/workers/routing.worker.ts`, `lib/utils/geometry/{AStarRouter,OrthogonalVisibilityGraph,SpatialIndex,edgeObstacles}.ts`, the `retainRoutingWorker()` effect in `lib/App.tsx`, and their tests) plus the now-write-only `routingMode` field — or rewire the renderer to call `calculateRoute`. Run `npm run build` and `npm run lint` in `library/` afterward.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk Build, lint and format are all green now, and the `OrthogonalEdge` test failure is gone with the file — nice. Two things still block: the `alignmentUtils.test.ts` failure from the `ALIGNMENT_THRESHOLD` 10→5 change is still red (existing thread, I re-confirmed it locally), and removing the orthogonal edge renderer left the whole routing engine as unreachable dead code — see inline. The two low-priority threads (`SpatialIndex` helpers, `EdgeTransformer` null guard) are also still open but non-blocking.

library/lib/utils/alignmentUtils.ts
Claudia-Anthropica: @FadyGergesRezk Confirmed fixed — the new `isWithinThreshold` helper makes the same-edge comparison inclusive (`<=`), so the exact 5px boundary at line 199 now produces the guide. Ran the full suite locally, 879 tests green. Resolving.

library/lib/App.tsx
Claudia-Anthropica: @FadyGergesRezk The whole orthogonal routing subsystem (A* router, visibility graph, spatial index, worker, store) is gone now, so there's no unreachable engine left behind. Clean removal — thanks for following through on it. Resolving.

library/lib/utils/geometry/SpatialIndex.ts
Claudia-Anthropica: @FadyGergesRezk `SpatialIndex.ts` is gone entirely with the routing-subsystem removal, so the dead `getMainSpatialIndex`/`rebuildSpatialIndex` helpers went with it. Resolving.

Claudia-Anthropica (approved):
@FadyGergesRezk All my blocking feedback is addressed. The new `isWithinThreshold` alignment refactor fixes the boundary case in `alignmentUtils.test.ts`, and ripping out the entire dead orthogonal routing subsystem (~2k LOC) cleared both the unreachable-engine concern and the `SpatialIndex` dead helpers in one go — the full library suite is green (879 passing), and build/lint/format/e2e/visual all pass. Nice cleanup after a long haul. Two optional leftovers, both [low] and your call: Codacy is red purely on three cosmetic hex-shorthand notices in `app.css` (`#ffffff`→`#fff`, `#000000`→`#000` at lines 183/211/234 — not enforced by the repo's own lint), and the `hydrateEdgeData` null-guard thread is still open. Approving.

library/lib/utils/versionConverter.ts
```diff
@@ -1,10 +1,6 @@
-// V2/V3 → V4 migrators read schemas TS can't statically describe (each
-// node type unions a different union of optional fields). Narrowed
-// disable: only `no-explicit-any` is suppressed; everything else still
-// lints. The `as any` casts are at type boundaries where runtime checks
-// (`isV*Format`, switch on `type`) have already discriminated the union.
-/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable */
```
Claudia-Anthropica: @FadyGergesRezk [medium] This broadens the old targeted `@typescript-eslint/no-explicit-any` suppression into a full-file ESLint shutdown. `versionConverter.ts` is migration code, but turning off every rule here also hides unused imports, unsafe comments, and future real lint regressions. Please keep the suppression scoped to the rules this boundary actually needs and fix any other lint errors directly.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `library/lib/utils/versionConverter.ts`, line 1 disables every ESLint rule for the whole file. Replace `/* eslint-disable */` with the narrow rule suppression needed for the converter, for example `/* eslint-disable @typescript-eslint/no-explicit-any */`, and fix any remaining lint errors without broadening the disable.\n\n</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The edge changes are mostly in good shape and the repo-owned checks are green. I’m requesting changes for one maintainability blocker: the version converter now disables every ESLint rule for the whole file instead of keeping the previous narrow suppression.

library/lib/utils/versionConverter.ts
Claudia-Anthropica: @FadyGergesRezk Still open on the latest head — `versionConverter.ts` still starts with `/* eslint-disable */`, so the converter is still disabling every ESLint rule in the file. Please restore the narrow `@typescript-eslint/no-explicit-any` suppression and fix any remaining lint directly.

Claudia-Anthropica (changes requested):
@FadyGergesRezk Still blocked on the same versionConverter.ts issue: the current head still disables every ESLint rule for the whole file. I replied on the existing thread, so keeping changes requested.

library/lib/utils/versionConverter.ts
Claudia-Anthropica: @FadyGergesRezk Looks good now, thanks — the converter is back to the narrow no-explicit-any suppression.

Claudia-Anthropica (approved):
@FadyGergesRezk All blocking feedback is addressed now; the version converter is back to a narrow suppression and I resolved that thread. Repo-owned lint/library/e2e checks are green, and the remaining Codacy red is still just the known cosmetic hex-shorthand notices.

library/lib/nodes/wrappers/DefaultNodeWrapper.tsx
```diff
@@ -78,137 +141,384 @@ export function DefaultNodeWrapper({
...
+  const rightMiddleForce = middleArcForced(
+    HandleId.RightTop,
+    HandleId.RightBottom
+  )
+
+  // Arc visibility per slot kind, after consulting both the side's plan and
+  // the per-side `hiddenHandles` overrides.
+  const showMiddleArc = (axisArcs: 1 | 3 | 5): boolean => axisArcs >= 1
+  const showCornerArc = (axisArcs: 1 | 3 | 5): boolean => axisArcs >= 3
+  const showMidCornerArc = (axisArcs: 1 | 3 | 5): boolean => axisArcs === 5
```
Claudia-Anthropica, line 252: @FadyGergesRezk [medium] This makes every mid-corner handle visible whenever a side reaches the five-arc stage, but several node components still only hide the old corner IDs because mid-corner handles were never primary before. For example `FlowchartDecision`, `PetriNetPlace`, `DeploymentInterface`, and the activity initial/final nodes will show extra `*-mid-*` arcs once they are wide/tall enough instead of keeping the intended four-side/vertex-only connection points. Either move those shapes to `FOUR_WAY_HANDLES_PRESET` or extend their `hiddenHandles` arrays to hide the mid-corner IDs too.

In `library/lib/nodes/wrappers/DefaultNodeWrapper.tsx`, the new five-arc stage makes mid-corner handles visible for every node unless each node hides them explicitly. Update the node-specific `hiddenHandles` lists for shapes that should only expose side-center/vertex handles, or switch them to `FOUR_WAY_HANDLES_PRESET`, and add a focused test or visual baseline coverage for one affected resized shape.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest handle-slot expansion builds and the repo-owned checks are green, but it regresses some shape-specific handle layouts. Wide nodes that used to hide only the corner handles can now expose the new mid-corner arcs; details inline.

library/lib/nodes/wrappers/DefaultNodeWrapper.tsx
Claudia-Anthropica: @FadyGergesRezk Looks good now, thanks — the affected side-center/vertex-only shapes are using FOUR_WAY_HANDLES_PRESET, so the mid-corner arcs stay hidden in the five-arc stage.

Claudia-Anthropica (approved):
@FadyGergesRezk All blocking feedback is addressed now; the handle preset fix resolves the remaining thread, and the repo-owned lint/library/e2e/visual checks are green. Codacy is still on the known cosmetic hex-shorthand notices, so I’m not holding that against this.

library/lib/utils/edgeUtils.ts
```diff
@@ -345,63 +351,644 @@ interface FindClosestHandleParams {
...
+      // Top side. The two corners (top-left, top-right) belong to this side;
+      // aliases left-top and right-top resolve to them via getHandleAnchorMap.
       {
         label: "top-left",
-        position: { x: rect.x + rect.width / 3, y: rect.y },
+        position: { x: xStart, y: yStart },
+        side: Position.Top,
+      },
+      {
+        label: "top-between-left-mid-left",
```
Claudia-Anthropica, line 775: @FadyGergesRezk [medium] These new `*-between-*` IDs can now be returned by `findClosestHandle()` for every non-four-handle node, but not every such node renders them. `UseCase` hides `DefaultNodeWrapper` and renders only the old 5-per-side IDs, so dropping or reconnecting near one of these between slots can persist a handle ID React Flow cannot resolve, and the edge disappears with its missing-handle error. Either render the between handles in those custom nodes or constrain the fallback lookup to IDs the target node actually exposes.

In `library/lib/utils/edgeUtils.ts`, `findClosestHandle()` can return new `*-between-*` handle IDs for custom nodes like `UseCase` that do not render those handles. Fix this by making the invalid-drop handle lookup aware of the target node's supported handle IDs, or by rendering hidden between-slot handles in the custom nodes that need them. Add a focused test that a UseCase-style node cannot receive a non-existent between-slot handle ID.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The edge and handle work is close, and the repo-owned checks are green. I found one current blocker: the new between-slot handle IDs can be selected for custom nodes that do not render them, which can make new/reconnected edges disappear.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on the current head: findClosestHandle still returns the new between-slot IDs for non-four-handle nodes, and UseCase still renders only the old handle IDs in its custom handle list. The latest PR diff against main does not include a fix for this path, so reconnect/drop near one of those between slots can still persist a handle React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The previous blocker is still unresolved: findClosestHandle can still choose between-slot handle IDs for UseCase-style custom nodes that do not render them. I replied on the existing thread, so keeping changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The current head still has the open between-slot handle issue: `findClosestHandle()` can return `*-between-*` IDs for UseCase-style nodes that do not render those handles. Keeping changes requested; details are in the existing thread.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on this head. The new commit improves edge hit targets and endpoint/bend-handle reachability, but the invalid-drop path still calls `findClosestHandle()` with only `useFourHandles`, and `UseCase` still renders only the legacy ellipse handle IDs. So a drop/reconnect near a `*-between-*` slot can still persist a handle React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The new zoom-stable edge hit targets look fine, but the existing between-slot custom-node blocker is still open. `findClosestHandle()` can still persist a `*-between-*` handle for UseCase-style nodes that do not render it, so keeping changes requested.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk This is still unchanged on `347c4dd`: `findClosestHandle()` still has no supported-ID filter, both invalid-drop callers still pass only `useFourHandles`, and `UseCase` still renders only the legacy ellipse handle IDs. Keeping this thread open.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The current head still has the open between-slot handle issue for custom UseCase-style nodes, and I replied on the existing thread. Keeping changes requested until that path only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `319400e`. The new commits only touch bend-handle visibility/zoom behavior; `findClosestHandle()` still has no supported-ID filter, and the invalid-drop callers still pass only `useFourHandles`, so UseCase-style nodes can still persist a `*-between-*` handle they do not render.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The new bend-handle commits look scoped, but the existing between-slot custom-node blocker is still open. Keeping changes requested until invalid-drop/reconnect only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `2db7a13`. The latest commit adjusts bend-handle rendering, but `findClosestHandle()` still has no supported-ID filter, the invalid-drop callers still pass only `useFourHandles`, and `UseCase` still renders only the legacy ellipse handles. So this path can still persist a `*-between-*` handle React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest commit changes bend-handle rendering, but the between-slot custom-node handle blocker is still open. Keeping changes requested until invalid drop/reconnect only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `78166dc`. The current implementation still lets both invalid-drop paths call `findClosestHandle()` without a supported-handle filter, while `UseCase` renders only the legacy ellipse handle IDs and no `*-between-*` handles. So this can still persist a handle ID React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The between-slot custom-node handle issue is still present on the current head, and I replied on the existing thread. Keeping changes requested until invalid drop/reconnect only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `4298c60`. The latest commit only changes wrapper resize behavior and adds E2E coverage; `findClosestHandle()` still accepts only `useFourHandles`, the invalid-drop/reconnect callers still do not pass a supported handle set, and `UseCase` still renders no `*-between-*` handles. Keeping this thread open until fallback handle selection can only return handles the target node actually renders.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The current head still has the open between-slot custom-node handle issue. Please constrain invalid drop/reconnect handle selection so it only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `0a1641b`. The latest commit only touches bend-drag decorations and SFC handle blocking; fallback handle selection is unchanged. `findClosestHandle()` still only accepts `useFourHandles`, `useConnect` still calls it without a supported-handle set, and `UseCase` still renders only the legacy ellipse handle IDs, so a drop near a `*-between-*` slot can still persist a handle React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest commit looks scoped, but the custom-node between-slot handle blocker is still open in the existing thread. Keeping changes requested until invalid drop/reconnect only stores handle IDs the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `c43217f`. The only new commit changes the edge toolbar foreignObject/shadow handling, so fallback handle selection is unchanged: `findClosestHandle()` still has no supported-handle filter, `useConnect` still calls it with only `useFourHandles`, and `UseCase` still renders only the legacy ellipse handles. A drop near a `*-between-*` slot can still persist a handle React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest commit only changes edge-toolbar shadow clipping, and the existing custom-node handle fallback blocker is still open. Please keep this in changes-requested until invalid drop/reconnect can only persist handle IDs that the target node actually renders.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `8fb7ae8`. The latest commit only changes the toolbar’s constant screen size; `findClosestHandle()` still has no supported-ID filter, `useConnect` still passes only `useFourHandles`, and `UseCase` still renders no `*-between-*` handles. A drop near a between slot can still persist an ID React Flow cannot resolve.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The existing custom-node fallback-handle blocker is still open on this head; I left the current verification in the thread. Lint, build, unit tests, and e2e are green now, but Codacy is still red on four `app.css` style notices. Keeping changes requested until the handle fallback is fixed and Codacy is clean.

library/lib/utils/edgeUtils.ts
Claudia-Anthropica: @FadyGergesRezk Still open on `2bb4ae9`. The commits after `8fb7ae8` only refactor edge rendering/tests; `findClosestHandle()` still takes only `useFourHandles`, and both `useConnect`/`useHandleFinder` still call it without constraining the returned ID to handles the target node actually renders.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest commits do not address the existing custom-node fallback handle blocker, and Codacy is still failing on this head. Keeping changes requested until invalid drop/reconnect can only persist handles the target node actually renders and the failing check is clean.

library/lib/utils/edgeUtils.ts
FelixTJDietrich: Fixed in 14ae6312 — thanks for holding the line on this; it was a real disappearing-edge risk.

I took a third approach rather than either option suggested, and want to be transparent about why:

- **Not** "render the between handles in custom nodes": the `*-between-*` slots are deliberately invisible — they never render an arc at any zoom stage. Making the UseCase ellipse (and friends) render them would add hidden DOM purely to back a snap target the user can't see.
- **Not** a node-aware supported-ID filter: that couples `findClosestHandle` to per-node-type handle config and would *still* let a drop snap to an invisible slot.

Instead, `findClosestHandle()` now never returns a `*-between-*` ID at all. Those slots exist only to keep **already-saved** edges attached when a node resizes between layout stages — they were never meant to be drop targets. The visible drag targets (and the only IDs custom nodes render) are the **named** handles, which every node type exposes, so a drop/reconnect now always persists a handle the target can resolve. I checked that every named candidate `findClosestHandle` can return is in the UseCase ellipse's supported set.

Added a focused regression test that sweeps every side — landing on the exact between-slot offsets — and asserts only named handles come back. It fails on the old code (returns `top-between-left-mid-left`) and passes now.

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. Filtering `findClosestHandle()` down to visible named handles fixes the disappearing-edge path without exposing hidden between slots as user drop targets.

library/lib/styles/app.css
```diff
@@ -103,42 +103,47 @@
   box-sizing: border-box;
   pointer-events: all;
   cursor: crosshair;
+  /* --arc-scale is set by DefaultNodeWrapper to 1/zoom so the indicator keeps
```
Claudia-Anthropica, line 107: @FadyGergesRezk [medium] Codacy is still failing on this line with `comment-empty-line-before`. Add a blank line before the `--arc-scale` comment so the style check turns green.

In `library/lib/styles/app.css`, Codacy reports `comment-empty-line-before` on the `--arc-scale` comment inside `.react-flow__handle.apollon-arc-handle::before`. Add an empty line before that comment and keep the CSS behavior unchanged.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The custom-node handle fallback blocker is fixed now, and keeping `*-between-*` IDs resolution-only looks right. Codacy still has one blocking style annotation in `app.css`, so I’m keeping changes requested until that check is clean.

library/lib/styles/app.css
FelixTJDietrich: Done in 64c58893 — added the blank line before the `--arc-scale` comment. Codacy now reports **success** with no annotations on the current head (`64c58893`), so this style check is clean.

(Note for future readers: the repo's own `format:check` only globs `js/ts/tsx/json/md/yml`, so CSS is linted by Codacy's stylelint alone — this rule isn't enforced by the repo's Prettier config.)

Claudia-Anthropica: @FadyGergesRezk Looks good now, thanks — the blank line is in place and Codacy is green.

PR #725, opened 2026-06-02
feat(webapp): migrate legacy native iOS app diagrams into the Capacitor app

FelixTJDietrich:
### Checklist

- [x] Unit tests added and passing locally (`pnpm test`, `pnpm --filter @tumaet/webapp test`)
- [x] `pnpm lint`, `pnpm format:check`, `pnpm build` pass locally
- [ ] **Release gate — not yet done:** the on-device same-container upgrade test (see [Steps for Testing](#steps-for-testing)). It is the only thing that proves the legacy SwiftData store actually opens and decodes on a real device, and it cannot be run in CI today.
- [x] No UI changes — screenshots N/A
- [ ] No linked issue (this work has no tracking issue)

### Motivation and Context

We are replacing the legacy **native** iOS Apollon app (Swift/SwiftUI, [`ls1intum/apollon-ios`](https://github.com/ls1intum/apollon-ios), App Store id `6474762031`) with the new Capacitor app. Existing users have diagrams saved **only** on their device, in the legacy app's local SwiftData store. A naïve switchover would make those diagrams vanish on update.

The key enabler: the Capacitor app already ships under the **same bundle id** as the legacy app (`de.tum.cit.ase.apollon`, same App Store listing, same Apple team). An in-place App Store update therefore lands in the **same sandbox container**, so the new app can read the old store directly — no manual export/import, no iCloud, fully automatic.

This PR makes that migration happen, once, transparently, without ever putting the legacy data at risk.

> No tracking issue exists for this work.

### Description

Two logically separate changes, split into two commits:

**1. `fix(library)` — v3→v4 converter correctness** (independently useful for web imports too)
- The legacy iOS editor records relationship endpoint directions in a `Top*/Bottom*` dialect (`Topright`/`Topleft`/`Bottomright`/`Bottomleft`). `convertV3HandleToV4` had no mapping for these, so they hit the lowercase fallback and produced **invalid v4 handle ids that mis-route edges**. Added the four mappings (geometry verified against the iOS module's `getPortsForElement`).
- Folded class/object attributes & methods are now ordered by `bounds.y`, preserving the legacy top-to-bottom visual order instead of object-key order.

**2. `feat(webapp)` — the migration**

| Layer | What it does | File |
| --- | --- | --- |
| Native plugin | Opens the legacy SwiftData store **read-only** (`allowsSave: false`) and re-encodes each diagram to standard Apollon **v3 JSON** | `ios/App/App/Migration/LegacyMigrationPlugin.swift` |
| Frozen models | A vendored, frozen copy of the legacy data models so the store decodes byte-for-byte | `ios/App/App/Migration/ApollonLegacyModels.swift` |
| Plugin registration | App-embedded Capacitor plugins are **not** auto-registered, so a `CAPBridgeViewController` subclass registers it explicitly | `ios/App/App/Migration/MainViewController.swift` |
| JS runner | Converts each diagram (`importDiagram`) and writes it to local storage; hydration-gated, per-diagram isolation, idempotent, durable done-flag | `src/services/legacyMigration.ts` |
| Store action | Bulk insert keyed by diagram id (idempotent), preserves last-modified time | `usePersistenceModelStore.importModels` |

**Design properties (and why they matter):**
- **Non-destructive.** The legacy store is opened read-only and never modified — it remains a permanent fallback, and manual JSON import keeps working, even after this code is eventually removed.
- **Idempotent.** Diagram/element ids are preserved, so a re-run overwrites rather than duplicating; a transient failure simply retries next launch.
- **Per-diagram isolation.** One corrupt diagram can't abort the whole batch; failures are surfaced, never silently dropped.
- **iOS 15 floor preserved.** All SwiftData code is gated behind `@available(iOS 17, *)`; on older systems (where no legacy data can exist) the plugin returns empty.
- **Removable by design.** The whole feature is one folder + one boot call; `Migration/README.md` documents the exact sunset/removal steps.

This PR also **commits the generated iOS project** (`standalone/webapp/ios/`), which is [Capacitor's recommended practice](https://capacitorjs.com/docs/v4/updating/3-0) — the project carries the bundle id, signing config, and this native code, while its own `.gitignore` excludes regenerated artifacts (`Pods/`, `public/`, `DerivedData`, generated config). The root `.gitignore` now ignores only a stray root-level `cap add` scratch dir and the (uncommitted) `android/` project.

### Scope & limitations (transparency)

- **iOS only.** The legacy native app only ever existed on iOS; the runner no-ops on web/Android.
- **The on-device upgrade test has not been run yet** (the dev/CI environment has the iOS SDK but no installed simulator runtime). Everything that can be verified without a device has been — see below — but the single highest-value check, proving the SwiftData schema matches a real legacy store, is still outstanding and is listed as a release gate above.
- **E2E (Playwright) was not run locally** (no browsers installed here); the migration takes an early return off iOS, so it does not affect web behaviour. CI will run it.

### Steps for Testing

**Automated (all green locally):**

```sh
pnpm install
pnpm lint && pnpm format:check && pnpm build && pnpm test   # 782 library tests
pnpm --filter @tumaet/webapp test                            # migration runner: 7 tests
```

Fidelity is covered by `library/tests/unit/legacyIosMigration.test.ts` (one round-trip per legacy-supported diagram type — class, object, activity, use-case, component — plus handle/ordering/isolation) and `src/services/legacyMigration.test.ts` (platform gate, idempotency, partial-failure isolation, fresh-install no-op, retry-on-error).

**Manual — the required release gate** (needs an iOS 17+ simulator/device). Full procedure in `standalone/webapp/ios/App/App/Migration/README.md`:

1. Build the legacy app (`ls1intum/apollon-ios`) under bundle id `de.tum.cit.ase.apollon` and create a rich set of diagrams.
2. Install this app **over it** (same bundle id ⇒ same container) without deleting the legacy app.
3. Verify every diagram appears and renders, relaunch shows no duplicates, and a fresh install starts empty with no errors.

### Screenshots

N/A — this change has no UI surface (data migration + native plumbing). User-visible effect is that diagrams from the previous app appear automatically after updating, with a one-time toast summarising the import.


standalone/webapp/ios/App/App/Migration/LegacyMigrationPlugin.swift
```diff
@@ -0,0 +1,117 @@
...
+    static func readAllAsV3JSON() throws -> [String] {
+        let url = try storeURL()
+
+        guard FileManager.default.fileExists(atPath: url.path) else {
+            throw MigrationError.noStore
+        }
+
+        // allowsSave: false guarantees we never mutate the legacy store, even
+        // if a future edit accidentally calls save().
+        let configuration = ModelConfiguration(url: url, allowsSave: false)
```
Claudia-Anthropica, line 84: @FelixTJDietrich [high] This file uses `ModelConfiguration`, `ModelContainer`, `ModelContext`, and `FetchDescriptor`, but it only imports Foundation and Capacitor. Swift imports are file-local, so the iOS target will not compile even though `ApollonLegacyModels.swift` imports SwiftData; add `import SwiftData` here too.

In `standalone/webapp/ios/App/App/Migration/LegacyMigrationPlugin.swift`, the plugin uses SwiftData APIs without importing SwiftData. Fix this by adding `import SwiftData` next to the other imports in that file.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The migration flow and converter tests look solid, but the native plugin has an iOS build blocker: the file that opens SwiftData does not import SwiftData. Codacy is also currently red with style-only notices, so that check still needs clearing before merge.

standalone/webapp/ios/App/App/Migration/LegacyMigrationPlugin.swift
FelixTJDietrich: Good catch — fixed in 5f8c0b22. Added `import SwiftData` to `LegacyMigrationPlugin.swift` (the types resolve from `ApollonLegacyModels.swift`'s import, but Swift imports are file-local, so this file needed its own).

Verified with a real build, not just a parse this time: `xcodebuild` for the iOS Simulator now reports **BUILD SUCCEEDED**, with the App target and all three migration sources compiling against the real Capacitor + SwiftData frameworks.

FelixTJDietrich:
Thanks for the review @Claudia-Anthropica — both points addressed in 5f8c0b22:

**1. iOS build blocker (missing `import SwiftData`)** ✅
Added the import. To be sure it actually builds this time, I downloaded the iOS 26 simulator runtime locally and ran a full `xcodebuild` — **BUILD SUCCEEDED**, with the App target and all three migration sources (`ApollonLegacyModels`, `LegacyMigrationPlugin`, `MainViewController`) compiling against the real Capacitor + SwiftData frameworks.

**2. Codacy / SwiftLint notices** ✅
Cleared all 18, verified locally with `swiftlint 0.56.2` → **0 violations** on the changed files. Resolutions:
- Removed the trailing comma in `pluginMethods` and switched to `String(decoding:as:)` for the non-optional UTF-8 conversion.
- `ApollonLegacyModels.swift` is a frozen, byte-for-byte mirror of the legacy on-disk schema — `x`/`y` are JSON keys, the raw values must match persisted strings — so renaming isn't possible. Scoped `identifier_name` / `redundant_string_enum_value` off for that file (the original legacy source did the same), with a matching `// swiftlint:enable`.
- Trimmed `AppDelegate` down to the functional methods (dropped Apple's empty no-op lifecycle stubs) and scoped a `line_length` disable to the API-fixed delegate signatures.

One thing I want to be transparent about: the PR's CI does **not** compile the native iOS target (the iOS build only runs in the manual TestFlight workflow), so this fix isn't exercised by the checks here — I verified it locally as described above. The one remaining pre-release gate is the on-device same-container upgrade test documented in `ios/App/App/Migration/README.md`, which is the only thing that can prove the SwiftData schema matches a real legacy store.

standalone/webapp/ios/App/App/Migration/LegacyMigrationPlugin.swift
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback addressed, nice work. Codacy is green now; the remaining checks I saw were only pending, not failing.

tamang29 (approved):

PR #690, opened 2026-05-19
chore: adopt Changesets for release-note generation

FelixTJDietrich:
## Summary

Adopt [Changesets](https://github.com/changesets/changesets) for release-note generation: a per-PR convention where each user-visible PR writes a changeset whose summary becomes the changelog entry. This is **PR A of the cutover** — the convention and its docs/CI land here; the pipeline that *consumes* changesets (`changeset version` → `CHANGELOG.md` → release) is a follow-up, so the existing `version-bump.yml` release path is unchanged.

> **Rebased onto the pnpm + Docusaurus monorepo.** `main` migrated from npm→pnpm and Sphinx→Docusaurus after this branch was opened, so the original (npm/Sphinx) version of this PR was rebuilt on the new foundation: pnpm scripts + lockfile, Changesets wired for pnpm workspaces, the style guide moved into the Docusaurus contributor docs, and CI converted to pnpm.

## What changed

| Area | Change |
| --- | --- |
| Convention | `.changeset/config.json`: `@changesets/changelog-github`, `fixed: [["@tumaet/webapp","@tumaet/server"]]`, `updateInternalDependencies: patch`. |
| Topology guard | `ignore: ["@tumaet/apollon-docs","apollon-vscode","apollon-vscode-menu","apollon-vscode-editor"]`. Every package depends on `@tumaet/apollon` via `workspace:*`, so without this a routine library bump would version the docs site and the separately-released VS Code extension too. **Verified empirically** with `changeset version`: with the `ignore`, a library minor bumps only library + webapp + server. |
| Deps | `@changesets/cli@2.31.0` + `@changesets/changelog-github@0.7.0` as root devDeps; `pnpm changeset` script; `pnpm-lock.yaml` updated. |
| Agent conventions | New root `AGENTS.md` ([agents.md](https://agents.md) convention); `CLAUDE.md` is a symlink to it (git mode `120000`). |
| Style guide | `docs/contributor/development/release-notes.md` — Docusaurus page (sidebar-registered): voice, bump type, two worked examples from real PRs (#683, #657). |
| Contributing | `CONTRIBUTING.md` gains a minimal `pnpm changeset` step + AGENTS pointer; `scripts.md` and `npm-publishing.md` cross-link the guide. |
[...]

PR #728, opened 2026-06-03
ci(release): automate versioning with a Changesets Version Packages PR

FelixTJDietrich:
## Summary

Turns on the canonical Changesets pipeline that #690 left *plumbed but not wired*. Today contributors declare a bump type per PR via `pnpm changeset`, but releases **discard** it — a maintainer hand-picks the bump from the `Version Bump` dropdown and a hand-rolled script applies it. This makes the declared bump types authoritative and answers the question #690 raised ("is a release PR created automatically?") with **yes**.

## What changes

- **`release.yml` (new)** — on push to `main`, [`changesets/action`](https://github.com/changesets/action) runs in **version-only** mode (no `publish` input) and opens/updates a single **Version Packages** PR. That PR runs the new root `changeset:version` script (`changeset version` + `sync-library-version.mjs` + lockfile regen). Merging it triggers the **existing** `release-library.yml` / `release-standalone.yml` publish jobs — those stay in charge of publishing, so the blast radius is "how package.json gets bumped", not "how we publish".
- **Native library→standalone cascade** — a library bump patches the paired `@tumaet/webapp` + `@tumaet/server` automatically via their `workspace:*` dependency (`updateInternalDependencies: patch`), replacing the hand-coded mirror in `version-bump.yml`.
- **`version-bump.yml` slimmed to the VS Code extension only** — it stays in `.changeset/config.json#ignore` (separate marketplace + cadence), so it keeps a manual dispatch.
- **GitHub Release body from the changelog** — new `scripts/extract-changelog.mjs` sources each Release body from the Changesets-owned `CHANGELOG.md` section, falling back to `--generate-notes` when a version carried no changeset. CRLF-normalized and fenced-code-block-aware so a changeset documenting markdown can't truncate the body.
- **Docs** — `npm-publishing.md`, `release-notes.md`, `scripts.md`, `overview.md`, and the vscode dev README updated to the live flow.

## Multi-agent audit (this branch was graded, not vibes)

Ran 5 web-validated principal-engineer reviewers + an adversarial round-2 gate. Acted on every confirmed finding:

- **Cut** an over-engineered optional GitHub-App-token path (unset var, untested dead conditional arm — YAGNI). Left a `TODO` for if/when pre-merge CI on the version PR is actually wanted.
- **Swept** stale "Version Bump" references the slimming left behind (4 files).
- **Pinned Node** (`setup-node`) in the standalone release job so the changelog step doesn't ride the runner's ambient Node.
- **Hardened** `extract-changelog.mjs` (CRLF + fence boundary) and trimmed comment bloat.
- Both new action SHAs were independently verified against the GitHub API (`changesets/action` v1.9.0, `create-github-app-token` removed).

## Verification

Pipeline run end-to-end with a throwaway `@tumaet/apollon: minor` changeset, then reverted: library 4.4.0 → **4.5.0**, webapp/server 4.4.1 → **4.4.2** (cascaded patch), both `CHANGELOG.md` regenerated, CDN URLs synced. `extract-changelog.mjs` exercised against real changelogs plus fence/CRLF/last-section/missing fixtures. `prettier --check`, `lint:lib`, and workflow YAML parse all pass.

## Reviewer notes

- **No changeset** on this PR — it's CI/docs-only (exempt per `CLAUDE.md`).
- **Declined as YAGNI** (stated, not silently dropped): prerelease-mode contract and the "Updated dependencies" cascade Release body — both speculative/marginal.

### DIFF
--- a/.github/workflows/release-standalone.yml
+++ b/.github/workflows/release-standalone.yml
@@ -65,6 +65,12 @@ jobs:
         with:
           ref: ${{ needs.check.outputs.sha }}
 
+      # Pin Node for the changelog-extraction step in "Create GitHub Release"
+      # rather than relying on the runner's ambient version.
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+        with:
+          node-version-file: ".nvmrc"
+
       - uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
         with:
           registry: ghcr.io
--- a/.github/workflows/release-standalone.yml
+++ b/.github/workflows/release-standalone.yml
@@ -115,13 +115,17 @@ jobs:
           VERSION: ${{ needs.check.outputs.version }}
           COMMIT_SHA: ${{ needs.check.outputs.sha }}
         run: |
-          gh release create "v${VERSION}" \
-            --title "v${VERSION}" \
-            --target "$COMMIT_SHA" \
-            --verify-tag \
-            --latest \
-            --generate-notes \
-            --notes "## Docker images
+          set -euo pipefail
+
+          # Body = the Changesets-owned CHANGELOG section (the webapp + server
+          # are paired, so they share one entry) plus the Docker + cosign
+          # footer. Fall back to GitHub's --generate-notes only when there is no
+          # changelog section for this version.
+          CHANGELOG=$(node scripts/extract-changelog.mjs standalone/webapp "$VERSION" || true)
+          {
+            if [ -n "$CHANGELOG" ]; then printf '%s\n\n' "$CHANGELOG"; fi
+            cat <<EOF
+          ## Docker images
 
           - \`ghcr.io/ls1intum/apollon/webapp:${VERSION}\`
           - \`ghcr.io/ls1intum/apollon/server:${VERSION}\`
@@ -132,4 +136,17 @@ jobs:
           cosign verify ghcr.io/ls1intum/apollon/webapp:${VERSION} \\
             --certificate-identity-regexp='^https://github\\.com/ls1intum/Apollon/\\.github/workflows/release-standalone\\.yml@refs/heads/main\$' \\
             --certificate-oidc-issuer=https://token.actions.githubusercontent.com
-          \`\`\`"
+          \`\`\`
+          EOF
+          } > release-notes.md
+
+          GENERATE=()
+          [ -z "$CHANGELOG" ] && GENERATE=(--generate-notes)
+
+          gh release create "v${VERSION}" \
+            --title "v${VERSION}" \
+            --target "$COMMIT_SHA" \
+            --verify-tag \
+            --latest \
+            "${GENERATE[@]}" \
+            --notes-file release-notes.md
--- a/.github/workflows/release-standalone.yml
+++ b/.github/workflows/release-standalone.yml
@@ -123,11 +123,12 @@ jobs:
         run: |
           set -euo pipefail
 
-          # Body = the Changesets-owned CHANGELOG section (the webapp + server
-          # are paired, so they share one entry) plus the Docker + cosign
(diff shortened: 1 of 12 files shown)


.github/workflows/release-standalone.yml
```diff
@@ -115,13 +115,17 @@ jobs:
...
-            --latest \
-            --generate-notes \
-            --notes "## Docker images
+          set -euo pipefail
+
+          # Body = the Changesets-owned CHANGELOG section (the webapp + server
+          # are paired, so they share one entry) plus the Docker + cosign
+          # footer. Fall back to GitHub's --generate-notes only when there is no
+          # changelog section for this version.
+          CHANGELOG=$(node scripts/extract-changelog.mjs standalone/webapp "$VERSION" || true)
```
Claudia-Anthropica: @FelixTJDietrich [medium] This only reads the webapp changelog, but fixed Changesets packages do not copy a server-only changeset into `standalone/webapp/CHANGELOG.md`. A release for an operator-facing server change would bump both packages and then publish a standalone GitHub Release without the server note. Please build the standalone notes from both `standalone/webapp` and `standalone/server` changelog sections, or use a helper that merges the non-empty sections before appending the Docker footer.

In `.github/workflows/release-standalone.yml`, the release body only extracts `standalone/webapp` changelog, so server-only changesets can be omitted. Fix this by including the matching section from both `standalone/webapp/CHANGELOG.md` and `standalone/server/CHANGELOG.md` in the generated standalone release notes, skipping only empty sections.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The release automation is close, but the standalone Release body can currently drop server-only changesets because it only reads the webapp changelog. See inline; that should be fixed before wiring this into main.

.github/workflows/release-standalone.yml
Claudia-Anthropica: @FelixTJDietrich Still seeing the same issue here: the standalone release body still only extracts `standalone/webapp`, so a server-only changeset can still be left out of the GitHub Release notes. Please merge in the matching `standalone/server` changelog section before the Docker footer too.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The standalone release notes issue is still open: the workflow still only extracts the webapp changelog, so server-only changesets can be omitted from the GitHub Release body. Keeping this as changes requested until the server changelog section is included or merged into the standalone notes.

.github/workflows/release-standalone.yml
FelixTJDietrich: Good catch — confirmed and fixed in 8fde9390.

You're right that this drops server-only notes. I reproduced it: a `"@tumaet/server": patch` changeset writes its note only to `standalone/server/CHANGELOG.md`, while the `fixed` pairing leaves `standalone/webapp/CHANGELOG.md` with an empty `## X.Y.Z` heading — so reading webapp alone returned nothing and silently fell back to `--generate-notes`, dropping the operator note.

Fix: `extract-changelog.mjs` now takes `<version> <dir>...` and merges the non-empty sections of all given packages, dropping exact duplicates. `release-standalone.yml` passes both `standalone/webapp` and `standalone/server`:

```sh
node scripts/extract-changelog.mjs "$VERSION" standalone/webapp standalone/server
```

The exact-dupe drop also keeps the common case clean: a library-cascade patch writes an identical "Updated dependencies → @tumaet/apollon@X.Y.Z" section to both files, so it now shows once rather than twice.

Verified across server-only, webapp-only, both-distinct, library-cascade-dupe, and no-changeset cases (fixtures + a real `changeset version` run); the server-only path now surfaces the note. The stale "they share one entry" comment is corrected too.

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback addressed and CI is green. Approving.

PR #742, opened 2026-06-09
feat(library): add edge-type preview icons to edge popovers

tamang29:
### Summary

The "Edge Type" dropdown in the edge-edit popover now shows a small preview of how each option actually renders — the real line style (solid/dashed) and end/start markers — so you can pick the right relationship by sight instead of decoding names. The selected type is also previewed in the collapsed dropdown.

This started as class-diagram-only; it now covers every edge popover that offers a type choice: **class, component, deployment, use case, and BPMN**.

| Before | After |
| --- | --- |
| Type names only | Each option shows its authentic edge preview |

### Implementation notes

- **Single source of truth — previews can't drift from edges.** Each preview derives its line dash and markers from `getEdgeMarkerStyles(edgeType)` (the same mapping the canvas uses) and draws them with the real `InlineMarker` component, rather than hand-drawing SVG paths. So the dropdown glyph is the same shape/fill the canvas paints.
- **Shared component.** A new `EdgeTypeSelect` replaces the duplicated `FormControl`/`Select` boilerplate that previously lived in all five popovers; `EdgeTypePreviewIcon` renders the preview.
- **Fit-to-row scaling.** Node-scale markers (the interface socket) are scaled down to fit the menu row, anchored at the tip so they stay on the line. Marker vertical extent is computed once, next to the renderer (`getMarkerHalfHeight`), so it can't desync.
- **Authentic by design.** Provided-interface previews render as a plain line because the lollipop ball is a separate node, not an edge marker — the preview reflects what the edge itself draws. Use case include/extend share a glyph because they differ only by stereotype text on the canvas too; the label disambiguates.
- **Scope cleanup.** The unrelated toast `z-index` change in `standalone/webapp/src/webapp.css` was reverted out of this PR. That removes the Codacy failure on that file **and** keeps the changeset correctly scoped to `@tumaet/apollon` only. The toast fix can land as its own focused PR.
- Also fixes the React key warnings in `EdgeStyleEditor` (keyed fragments for `sideElements`/`colorFields`), matching `NodeStyleEditor`.

### Steps for testing

1. Open a **Class** diagram, connect two classes, and edit the edge — the "Edge Type" dropdown shows a preview next to each option (hollow vs filled diamond, hollow triangle, dashed dependency/realization, …).
2. Repeat in **Component**, **Deployment**, **Use Case**, and **BPMN** diagrams.
3. Toggle dark mode — markers use `currentColor` and stay legible.

### Screenshots / screencasts

Class diagram edge type previews

### Checklist

- [x] Added a changeset (`pnpm changeset`, [how](https://ls1intum.github.io/Apollon/contributor/development/release-notes/))
- [x] Tests added or updated

### DIFF
--- a/.changeset/eight-frogs-love.md
+++ b/.changeset/eight-frogs-love.md
@@ -2,4 +2,4 @@
 "@tumaet/apollon": patch
 ---
 
-Class diagram edge type options now show matching edge icon previews.
+The edge-type dropdown now previews how each option renders, for class, component, deployment, use case, and BPMN diagrams.
--- a/.changeset/eight-frogs-love.md
+++ b/.changeset/eight-frogs-love.md
@@ -1,5 +1,5 @@
 ---
-"@tumaet/apollon": patch
+"@tumaet/apollon": minor
 ---
 
-The edge-type dropdown now previews how each option renders, for class, component, deployment, use case, and BPMN diagrams.
+The edge-type dropdown now shows a small preview of how each option renders — its real line style and arrowheads — so you can pick the right relationship at a glance, across class, component, deployment, use case, and BPMN diagrams.
--- a/.changeset/eight-frogs-love.md
+++ b/.changeset/eight-frogs-love.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Class diagram edge type options now show matching edge icon previews.
--- a/standalone/webapp/src/webapp.css
+++ b/standalone/webapp/src/webapp.css
@@ -42,7 +42,6 @@ body {
   flex-grow: 1;
   /* background-color: var(--apollon-background); */
   height: 100%;
-
 }
 
 .apollon-editor {
@@ -70,4 +69,10 @@ body {
 
 .MuiMenu-list .MuiMenuItem-root.Mui-selected:hover {
   background-color: var(--apollon-gray) !important;
+}
+
+/* Keep notifications above Apollon's collaboration presence/cursor overlays. */
+.Toastify__toast-container {
+  --toastify-z-index: 10005;
+  z-index: 10005;
 }
--- a/standalone/webapp/src/webapp.css
+++ b/standalone/webapp/src/webapp.css
@@ -42,6 +42,7 @@ body {
   flex-grow: 1;
   /* background-color: var(--apollon-background); */
   height: 100%;
+
 }
 
 .apollon-editor {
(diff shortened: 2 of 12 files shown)


.changeset/eight-frogs-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tumaet/apollon": patch
```
Claudia-Anthropica: @tamang29 [medium] This changeset only bumps `@tumaet/apollon`, but the PR also changes `standalone/webapp/src/webapp.css`, and `@tumaet/webapp` is versioned in the changeset config. Either remove the toast z-index change from this feature PR or add a patch entry for `@tumaet/webapp` with a short release note.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `.changeset/eight-frogs-love.md`, the changeset only covers `@tumaet/apollon` while the PR also changes `standalone/webapp/src/webapp.css`. Fix this by either removing the webapp CSS change from the PR or adding a patch changeset entry for `@tumaet/webapp` that describes the toast z-index change.\n\n</details>

standalone/webapp/src/webapp.css
```diff
@@ -71,3 +70,9 @@ body {
 .MuiMenu-list .MuiMenuItem-root.Mui-selected:hover {
   background-color: var(--apollon-gray) !important;
 }
+
+/* Keep notifications above Apollon's collaboration presence/cursor overlays. */
+.Toastify__toast-container {
+  --toastify-z-index: 10005;
+  z-index: 10005;
```
Claudia-Anthropica: @tamang29 [medium] Codacy is currently failing on this declaration because `declaration-empty-line-before` expects an empty line before `z-index` after the custom property. Add the blank line so the external static-analysis check goes green.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/webapp.css`, Codacy reports `declaration-empty-line-before` on line 77 because `z-index` follows the custom property without a blank line. Add an empty line between `--toastify-z-index: 10005;` and `z-index: 10005;`.\n\n</details>

Claudia-Anthropica (changes requested):
@tamang29 The class-edge icon previews look fine, but this still needs changes before merge. Codacy is failing on the webapp CSS addition, and that same webapp UI change is not covered by the changeset.

.changeset/eight-frogs-love.md
Claudia-Anthropica: @tamang29 Looks good now, thanks.

standalone/webapp/src/webapp.css
Claudia-Anthropica: @tamang29 Nice, removing the webapp CSS also clears this.

FelixTJDietrich (approved):
I changed some things, looks good now, thanks!

PR #744, opened 2026-06-10
fix(library): keep live collaboration cursors aligned to the canvas

tamang29:
### Summary

Live collaboration cursors could appear in the wrong place — or float outside the editor — when two collaborators had **different editor sizes** (e.g. one had a host side panel open). Cursors were positioned in browser-window coordinates over a `position: fixed` overlay, so they ignored where the canvas actually sat on each screen.

Now remote cursors are positioned **relative to the modeling canvas** and clipped to it, so they land on the right logical point for every collaborator regardless of window size or surrounding app chrome.

### Implementation notes

- **The fix (library):** `flowToCanvasPosition` maps a flow point to canvas-local pixels (`flow · zoom + pan`) — the same transform React Flow applies to its pane — instead of `flowToScreenPosition` (which adds the editor's per-client on-screen offset). The cursor layer moved from `position: fixed` to `position: absolute` inside a new `.apollon-canvas` wrapper (`position: relative; overflow: hidden`), so cursors share the canvas's coordinate space and can't escape it.
- **Side effect, by design:** the collaboration overlays (presence bar, follow frame) and the scroll-lock overlay now anchor to the canvas rather than the full editor (which includes the element palette). This is more correct and has no negative impact; popovers are unaffected (they portal to `body`, and `.apollon-canvas` deliberately sets no `transform`/`filter`, so it never becomes their containing block — guarded with a comment).
- **Reproduction harness (webapp):** the playground gained a "Collaboration viewport test" toggle that links two browser tabs over a `BroadcastChannel` and shows collapsible side panels mimicking host-app chrome — open it in two tabs, enable in both, and collapse the panel in one to reproduce differing editor widths.
- **Review-driven cleanup:** the two near-identical playground sidebars are now one shared `CollapsibleSidebar` (Tailwind, accessible toggle); a placebo cross-tab test and implementation-detail width assertions were removed.

### Steps for testing

1. Open `/playground` in two browser tabs.
2. Tick **Collaboration viewport test** in both — live cursors/presence connect across the tabs.
3. Collapse the left controls (or right) panel in **one** tab so the two tabs have different canvas widths.
4. Move the pointer in each tab: the remote cursor stays on the correct canvas point in the other tab and never drifts outside the canvas.

### Checklist

- [x] Added a changeset (`@tumaet/apollon` for the fix, `@tumaet/webapp` for the playground harness)
- [x] Tests added or updated


standalone/webapp/src/components/playground/PlaygroundControlsSidebar.tsx
```diff
@@ -0,0 +1,81 @@
...
+export const PlaygroundControlsSidebar = ({
+  open,
+  onToggle,
+  children,
+}: {
+  open: boolean
+  onToggle: () => void
+  children: ReactNode
+}) => (
+  <aside
```
Claudia-Anthropica: @tamang29 [medium] This new `<aside>` becomes the first `aside` on `/playground`, so `standalone/webapp/tests/e2e/editor.spec.ts` now points at the controls sidebar instead of the Apollon element palette and CI fails in `changing diagram type re-renders the sidebar`. Please update the e2e selector to target the editor palette explicitly.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/components/playground/PlaygroundControlsSidebar.tsx`, the new playground controls `<aside>` is now selected by the existing `/playground` e2e test before the Apollon element palette. Update `standalone/webapp/tests/e2e/editor.spec.ts` to select the visible `aside` containing `.prevent-select` instead of using `locator("aside").first()`, and keep the draggable-item assertion passing.\n\n</details>

Claudia-Anthropica (changes requested):
@tamang29 The cursor clipping approach looks good, but CI is currently red because the new playground sidebar changes which `<aside>` the existing e2e test selects. Please update the e2e selector or markup so the playground test targets the Apollon element palette again.

standalone/webapp/src/components/playground/PlaygroundControlsSidebar.tsx
Claudia-Anthropica: @tamang29 Looks good now, thanks.

Claudia-Anthropica (approved):
@tamang29 All feedback addressed, nice work.

Claudia-Anthropica (approved):
@tamang29 Looks good; the cursor positioning is now canvas-local, and the playground/test updates line up with the intended repro workflow.

FelixTJDietrich (approved):
I improved some minor things and made the positioning a bit more precise. Should be good now! Thanks for the fix :)

PR #758, opened 2026-06-14
fix: iOS app is available on both iPhone and iPad again

tamang29:
### Summary

Apple does not allow an existing app update to drop iPhone support.
For iPad-only, TARGETED_DEVICE_FAMILY = "2" is correct. The problem is Apple won’t allow the existing bundle ID to drop previously supported iPhones.
The only valid route is:
1. Create a new Bundle ID.
2. Create a new App Store Connect app.
3. Configure it with TARGETED_DEVICE_FAMILY = "2".
4. Publish it as a separate iPad-only app.
### Checklist

- [x] Added a changeset (`pnpm changeset`, [how](https://ls1intum.github.io/Apollon/contributor/development/release-notes/)), or this PR doesn't touch a Changesets-tracked package (`@tumaet/apollon`, `@tumaet/webapp`, `@tumaet/server`)
- [ ] Tests added or updated


Claudia-Anthropica (approved):
@tamang29 This does exactly what the PR describes: the existing iOS bundle stays intact and both build configurations now target iPhone and iPad again. CI is green, so approving as-is.

Issue #757, opened 2026-06-13
feat(webapp): migrate MUI to shadcn/ui and unify editor UI with the new homepage design


FadyGergesRezk:
## What problem would this solve?

The standalone web app currently has a **visual split personality**. The new homepage (`HomePage.tsx`, `components/home/`) was built from scratch with Tailwind CSS and a rich set of CSS custom-property design tokens (`--home-*`), giving it a clean, modern look. The editor shell — navbar, modals, version-history drawer, and export menus — still uses **MUI 6** (`@mui/material`) with ad-hoc CSS overrides that fight the MUI theme rather than build on the home-page token system.

Inside the embeddable **`@tumaet/apollon` library** the situation is similar: 41 files import MUI for layout (`Box`), form controls (`Select`, `TextField`, `FormControl`), and the popover system (`Popover`, `Paper`). None of them use a MUI `ThemeProvider` or `createTheme` — all theming already flows through CSS custom properties (`--apollon-*`). MUI is used as a bag of unstyled primitives that happen to be styled at the CSS-variable layer, which means it could be replaced without touching any theme logic.

The combined result:

- Two unrelated visual systems exist side-by-side in the same app (homepage vs. editor shell).
- Opening a diagram immediately breaks the design continuity the homepage establishes.
- MUI's CSS-in-JS (Emotion) ships a large runtime bundle; its theming capability is entirely bypassed because both workspaces already use CSS variables.
- New contributors face a confusing mix: "use Tailwind on this component, MUI on that one."
- The library peer-depends on `@mui/material` (optional, but still present), increasing bundle cost for every host that embeds it.

---

## Proposed solution

Replace every MUI import in **both `standalone/webapp/` and `library/`** with [shadcn/ui](https://ui.shadcn.com/) components (Radix UI primitives + Tailwind variants) in the webapp, and with lightweight Radix UI primitives or native HTML + CSS in the library, so that homepage, editor shell, and the embedded editor all share one visual language.

---

### `standalone/webapp/` — 29 MUI-importing files

| Area | Key MUI components used | Replacement |
|---|---|---|
| Editor navbar (`NavbarFile`, `NavbarHelp`) | `Button`, `Menu`, `MenuItem`, `Typography` | `Button`, `DropdownMenu` (shadcn) |
[...]

PR #759, opened 2026-06-14
feat: migrate MUI/Emotion to Base UI + shadcn, and unify + modernize the editor and web app

FadyGergesRezk:
### Summary

Removes **MUI 6 + Emotion** entirely from both the editor and the standalone web app, replacing them with [Base UI](https://base-ui.com) primitives and one shared, shadcn-style component vocabulary — and, as issue #757 asks, unifies the editor UI with the homepage design so the whole product reads as one app.

- **Editor — `@tumaet/apollon`.** Its internal primitives (Button, IconButton, Select, Tooltip, TextField, Typography) are rebuilt on Base UI and styled with plain CSS through the public `--apollon-*` custom properties. The library stays **Tailwind-free and Preflight-free on purpose**, so it keeps embedding cleanly into any host (VS Code, non-Tailwind apps). All popovers, toolbars, controls, and the minimap move onto the new system. A new **public theming API** (`createApollonTheme()` plus `theme` / `dataTheme` props) lets hosts theme the editor without hand-writing CSS variables.
- **Web app — `@tumaet/webapp`.** Migrates to the new shared **`@tumaet/ui`** library. The editor shell, navbar, modals, version history, and the home gallery now share one accessible component set and one set of design tokens with the homepage — a cohesive floating-glass chrome where every action control (File, Help, Share, Favorites, Refine, Import, Theme, version, save) reads as one family: an icon plus a label that reveals as space allows, backed by a single instant tooltip.
- **`@tumaet/ui` (new internal package).** The shared design system — 28 components on Base UI + Tailwind v4 + `lucide-react` — with Storybook coverage for every component. Private (`0.0.0`), so it carries no changelog of its own; its user-visible impact ships through `@tumaet/webapp`.

Icons are unified on `lucide-react`; the custom SVG icon adapter is gone.

### What changed

| Area | Before | After |
|---|---|---|
| Editor primitives | MUI + Emotion | Base UI, styled via `--apollon-*` CSS |
| Web app components | MUI 6 / Emotion | `@tumaet/ui` (shadcn on Base UI + Tailwind) |
| Popovers · tooltips · select · dialogs · menus · sheets | MUI | Base UI (focus trap, `Escape`, ARIA out of the box) |
| Icons | custom SVG wrappers + `makeIcon()` | direct `lucide-react` |
| Design tokens | scattered | single source of truth, exposed to Tailwind via `@theme` |
| Editor + home chrome | two separate looks | one cohesive floating-glass language, light + dark |
| Editor theming | CSS variables only | `--apollon-*` contract **+** typed `createApollonTheme()` helper |
| Component docs | — | Storybook (one monorepo-wide host in the web app) |

### Beyond the migration — design-system, theming, and quality hardening

The branch doesn't just swap frameworks; it brings the migrated surfaces to A+:

- **Editor popovers — modernized and made cohesive.** Every node/edge/feedback popover was audited and unified onto one section/heading grammar: a "Style" heading for name-less nodes, the «stereotype» toggle moved to its own row (component/deployment), an **embed-safe `Checkbox` primitive** replacing the raw checkboxes (reachability/SFC), consistent "+ Add" affordances and bare-noun placeholders, and **multiline name fields that start single-line and grow per newline** (the 64px floor is gone).
- **Assessment feedback — redesigned.** The give/see-feedback popovers became a **Scorecard** — a tone-coded score pill (positive/negative/zero/ungraded) with matching feedback states — and the **on-canvas feedback badge was modernized to mirror it exactly** (a soft tone-tinted disc, not the old washed-out gray circle). All assessment colours flow through the token system (`--apollon-assessment-*`, light + dark deltas); they're **WCAG-AA and axe-gated**, no hardcoded hex.
- **Storybook — raised to A+.** The editor stories now cover **every diagram type** plus the give/see-feedback popovers, full-editor **assessment** (give + a graded review that shows the on-canvas badges), **highlight** (the interactive-element picker *and* the imperative `setElementHighlights` host API), **assessment selection**, and **modes** — all **test-gated in a real browser with a11y enforced**, and scannable in autodocs. `@tumaet/ui` and webapp stories are likewise hardened (real interaction tests, dark/matrix coverage, no theatre).
- **Naming & structure convergence.** Boolean-trap props → variant unions; mislabeled navbar components renamed (`FileMenu` / `HelpMenu` / `BrandLockup` / `EditorHeader`); container/presentational extractions; a typed modal registry (no more `Record<string, unknown>`).
- **Docs corrected + completed.** The UI/Overview, Theming, Tokens, and library theming/API docs were reconciled against the actual code; a `cssVariableContract` test enforces that the public token table stays in sync. The last MUI-ism — the fake `sx` prop on the editor's Typography/TextField wrappers — was removed.
- **Fixes.** Palette drag no longer jumps on grab and matches the drop size at any zoom; the theme switch is a single instant flip; several home/legal e2e tests fixed (including a real notch-clearance `calc()` bug) and the swimlane keyboard-reorder e2e hardened.

### Release note

The published changelog entries are the three changesets this PR adds, each written in its track's voice:

- **`@tumaet/apollon` (minor)** — Removes MUI and Emotion from the editor; controls, popovers, toolbars, and the minimap are rebuilt on lightweight Base UI primitives styled entirely through the public `--apollon-*` CSS variables — smaller dependency footprint, no Emotion runtime, no MUI style collisions, look and behaviour unchanged.
- **`@tumaet/apollon` (minor)** — Adds a public theming API (`createApollonTheme()` + `theme` / `dataTheme` props) so hosts can theme the editor with a typed helper instead of hand-writing CSS variables.
- **`@tumaet/webapp` (minor)** — Rebuilds the web app's interface on one shared component system, so the editor, dialogs, version history, and gallery share a consistent look matching the homepage in light and dark; menus, dialogs, and sheets are fully keyboard-accessible; the app loads lighter after dropping its old UI framework.

`@tumaet/server` rides a paired **minor** bump (Changesets `fixed` group with `@tumaet/webapp`); its image is republished but its behaviour is unchanged.

### Implementation notes

- **Why two styling modes (the accurate version).** The editor is published and embeds in arbitrary hosts, so it can't ship Tailwind/Preflight (utility-class and reset collisions). **Most `@tumaet/ui` components are standard shadcn** with inline Tailwind utilities and are **webapp-only** (the webapp runs Tailwind). The **five primitives the editor shares — `button`, `icon-button`, `input`, `textarea`, `tooltip` — are authored differently**: an empty `cva()` (variants only type/validate) plus `data-slot` rules in `src/styles/components.css`. `build:css` compiles **only that file** into a Preflight-free, Tailwind-free `dist/components.css` containing exactly those five primitives + the token values — no bare utility classes — which is what makes it embed-safe.
- **Token layering.** `--primitive-*` → `--apollon-*` (public editor API) → `--home-*` (web app) → `--color-*` (shadcn/Tailwind `@theme inline`). Light values in `:root`, dark deltas in `:root[data-theme="dark"]`; a single `data-theme` flip drives both, and a scoped `[data-theme="dark"]` themes a mounted editor without touching `:root`.
- **Accessibility.** Dialogs, menus, and sheets get focus trapping, `Escape` handling, and ARIA wiring from Base UI. Icon-only controls carry a real accessible name and an instant shared tooltip rather than a native `title`. New assessment colours are WCAG-AA and the editor story suite runs the axe color-contrast gate.
- **Verification surface.** A single Storybook host (in the web app) covers `@tumaet/ui`, the editor (organized by diagram type), and the web app; light/dark is reviewed there (stories and Docs). The editor's harness-based stories are test-gated.
- **Dead-code & dependency hygiene — [knip](https://knip.dev).** A migration this size churns a lot of modules, so the branch adds knip (`knip.json` plus a root `pnpm knip` script) to catch unused files, exports, and dependencies across every workspace. It is how the dead MUI-era components and stories were found and removed. It runs on demand and is **not** yet a blocking CI gate.

#### Scope & transparency

- This is a **large diff (~485 files)**, but a single cohesive change: the migration plus the editor↔homepage unification that issue #757 calls for, brought to a finished A+ standard. They land together because you can't take the home gallery and editor chrome off MUI without rebuilding those components — and rebuilding them is what makes the two surfaces match.
- The branch is **up to date with `main`** (0 commits behind). Features that already shipped on `main` — activity swimlanes, diagram embedding, the server-render fixes, and the original floating-glass editor chrome (#778) — are **not** part of this diff and keep their own changesets.
- `@tumaet/ui` is intentionally **private**, so it is excluded from Changesets tracking; everything a user or operator notices flows through the `@tumaet/webapp` entry above.

### Steps for testing

1. `pnpm install && pnpm build && pnpm lint && pnpm test`
2. **Editor** — open each diagram type: double-click nodes and edges (popovers), hover the toolbar (tooltips), use zoom / fit / undo / redo and the minimap. In Assessment mode, give feedback (the Scorecard) and review a graded diagram (the on-canvas badges).
3. **Web app** — exercise the navbar, New Diagram, Share, version history, and the home gallery; toggle light ↔ dark; resize from desktop down to a 390 px phone and confirm the action controls keep a stable order and reveal tooltips instantly.
4. **Storybook** — `pnpm --filter @tumaet/webapp run storybook` (→ http://localhost:6006), browse **Editor** (by diagram type, incl. the Assessment / Feedback / Modes stories) + **Webapp** + **UI**, and flip the theme toolbar on both stories and Docs.
5. **E2E** — `pnpm run test:e2e`.


library/lib/components/ui/Select.tsx
```diff
@@ -0,0 +1,218 @@
...
+                size={16}
+                aria-hidden
+                className="apollon-select-icon"
+              />
+            </button>
+          }
+        />
+        <Popover.Portal>
+          <Popover.Positioner align="start" sideOffset={4} collisionPadding={8}>
+            <Popover.Popup
```
Claudia-Anthropica, line 193: @FadyGergesRezk [medium] This popup is portaled to `document.body`, so it no longer inherits the `--apollon-*` variables from the editor or from `GenericPopover`'s themed popup. In a dark/custom-themed embed the trigger can use the right colors while the opened menu falls back to the default white/black CSS. Copy the same theme variables from the trigger/editor onto the portaled popup, or render the portal inside a themed container.

In `library/lib/components/ui/Select.tsx`, the select popup is portaled to `document.body` without the editor's `--apollon-*` theme variables. Fix this by resolving the variables from the select trigger element's computed style and applying them to `Popover.Popup` or `Popover.Positioner`, reusing the GenericPopover variable list if possible.

</details>

library/lib/components/ui/Tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  ) : (
+    <BaseTooltip.Trigger delay={delayDuration}>{children}</BaseTooltip.Trigger>
+  )
+
+  return (
+    <BaseTooltip.Root>
+      {trigger}
+      <BaseTooltip.Portal>
+        <BaseTooltip.Positioner side={side} sideOffset={4}>
+          <BaseTooltip.Popup className="apollon-tooltip">
```
Claudia-Anthropica: @FadyGergesRezk [medium] Same portal issue here: tooltip content renders under `body`, but its CSS uses `--apollon-background` and `--apollon-primary-contrast`. When an embedder sets those variables on the editor container, the tooltip ignores them and uses fallback colors. Resolve/apply the theme variables to the portaled popup just like `GenericPopover` does.

In `library/lib/components/ui/Tooltip.tsx`, the tooltip popup is portaled to `document.body` without the editor's `--apollon-*` theme variables. Fix this by reading the trigger element's computed theme variables and applying them to `BaseTooltip.Popup` so custom and dark themes carry into the tooltip portal.

</details>

library/lib/styles/app.css
```diff
@@ -11,12 +11,13 @@
 .apollon-editor {
   background-color: var(--apollon-background, #fff);
   color: var(--apollon-primary-contrast, #000);
+  --popover-divider: #d1d5dc;
```
Claudia-Anthropica: @FadyGergesRezk [low] Codacy is failing on the new CSS block because stylelint expects blank lines between declarations/rules and short hex colors. The annotations are all in this file, so please run the Codacy/stylelint rules over the added UI CSS and clean them up; GitHub currently reports 38 new issues.

In `library/lib/styles/app.css`, the added UI CSS violates Codacy/stylelint rules such as `custom-property-empty-line-before`, `rule-empty-line-before`, and `color-hex-length`. Fix this by formatting the new CSS to satisfy those rules and replacing long hex colors where the rule requires short hex.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The migration is broadly in the right shape, but the new portaled Base UI wrappers currently drop custom `--apollon-*` theme variables for opened select menus/tooltips. Codacy is also failing on the new CSS stylelint notices, so please clean those up before merging.

library/lib/components/ui/Select.tsx
Claudia-Anthropica: @FadyGergesRezk Still open: the current `Select.tsx` popup at line 175 still only applies width/maxHeight, so the body portal still loses editor-scoped `--apollon-*` variables. Please copy the same variables onto the popup/positioner before this is resolved.

library/lib/components/ui/Tooltip.tsx
Claudia-Anthropica: @FadyGergesRezk This moved into `@tumaet/ui`, but it is still not fixed. `TooltipContent` in `packages/ui/src/components/tooltip.tsx` still portals to `body` and applies no resolved theme vars, so an editor-scoped custom/dark theme still falls back to root defaults.

library/lib/styles/app.css
Claudia-Anthropica: @FadyGergesRezk Still open: Codacy is still action-required, and its annotations still include `library/lib/styles/app.css` for the same empty-line/hex-length stylelint rules. Please clean those plus the new `packages/ui/src/styles/components.css` annotations before merging.

standalone/webapp/tests/helpers/canvas.ts
```diff
@@ -46,7 +46,7 @@ export async function waitForCanvasReady(page: Page, expectNodes = true) {
  * Useful for large diagrams (e.g. BPMN) that overflow at zoom 1.0.
  */
 export async function clickFitView(page: Page) {
-  const fitViewBtn = page.locator(".react-flow__controls-fitview")
+  const fitViewBtn = page.locator('[aria-label="Fit to view"]')
```
Claudia-Anthropica: @FadyGergesRezk [medium] This selector is currently breaking the visual suite: CI times out waiting for `[aria-label="Fit to view"]` in both BPMN visual tests. Use the React Flow fit-view class again or otherwise target the actual rendered control so the helper can click the button reliably.

In `standalone/webapp/tests/helpers/canvas.ts`, `clickFitView()` now waits for `[aria-label="Fit to view"]`, but the visual CI cannot find that control and times out. Fix the helper by targeting the actual rendered React Flow fit-view button, for example by restoring `.react-flow__controls-fitview` or updating the UI to expose the expected accessible label.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The earlier portal-theme and Codacy issues are still open, and visual-regression-tests is now failing because the fit-view helper cannot find the new selector. Requesting changes until those are fixed and CI is green.

library/lib/styles/app.css
Claudia-Anthropica: @FadyGergesRezk Still open: Codacy is still failing, and `lint-and-format-check` now also fails Prettier on `knip.json` plus the editor Storybook files. Please run the formatting/style checks locally before the next push.

library/lib/components/ui/Select.tsx
Claudia-Anthropica: @FadyGergesRezk Still open on the current head: `Popover.Popup` still only gets `width`/`maxHeight`, so the body portal still loses editor-scoped `--apollon-*` values.

library/lib/components/ui/Tooltip.tsx
Claudia-Anthropica: @FadyGergesRezk Still open on the current head: `TooltipContent` still renders through a body portal and does not copy the editor-scoped theme variables onto the portaled popup.

standalone/webapp/tests/helpers/canvas.ts
Claudia-Anthropica: @FadyGergesRezk Still open in the current head: `clickFitView()` still waits for `[aria-label="Fit to view"]`, which is the selector visual CI could not find. Please target the actual React Flow fit-view button again.

packages/ui/src/components/popover.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    return () => observer.disconnect()
+  }, [open, positioner])
+
+  return { actionsRef, positionerRef: setPositioner }
+}
+
+// Shared between the Popover and Menu primitives: the Root captures the
+// open/close machinery, the Content's Positioner consumes the guard ref.
+const AnchorLossGuardContext =
```
Claudia-Anthropica: @FadyGergesRezk [high] CI is currently blocked here: `@tumaet/ui` declaration generation fails because this exported context exposes `React.RefCallback<HTMLDivElement>`, and React 19's callback ref type pulls in an internal `UNDEFINED_VOID_ONLY` symbol that TypeScript cannot name in the `.d.ts`. Use a local callback type like `(node: HTMLDivElement | null) => void` for the context/hook public shape, or otherwise keep `React.RefCallback` out of the exported declaration.

In `packages/ui/src/components/popover.tsx`, the exported `AnchorLossGuardContext` uses `React.RefCallback<HTMLDivElement>` in its public declaration, which breaks declaration emit with React 19's internal `UNDEFINED_VOID_ONLY`. Fix this by replacing the context and hook public type with a local `(node: HTMLDivElement | null) => void` callback alias, and keep `dropdown-menu.tsx` consumers compiling.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The previous portal-theme, style/format, and fit-view issues are still open, and CI is now blocked during `@tumaet/ui` declaration generation; see the new inline comment. Requesting changes until those are fixed and the build/format checks are green.

pnpm-workspace.yaml
```diff
@@ -18,11 +19,25 @@ blockExoticSubdeps: true
...
+  "@storybook/addon-a11y": "10.4.6"
+  "@storybook/addon-themes": "10.4.6"
+  "@storybook/addon-vitest": "10.4.6"
+  # Storybook browser tests use @vitest/browser-playwright, which needs a modern
+  # playwright (>=1.6x). e2e + the committed visual baselines are pinned to
+  # @playwright/test 1.59.1 (CI's v1.59.1-noble image). Both packages ship a
+  # `playwright` CLI bin, so e2e/visual invoke @playwright/test's CLI directly
+  # (see webapp test:e2e / test:visual) to avoid the bin collision; this package
+  # is consumed only via its API by the browser-test provider.
+  playwright: "1.61.0"
```
FelixTJDietrich: We need to make sure that we use a consistent version and prevent drift

package.json
```diff
@@ -39,6 +44,7 @@
     "format:check": "prettier --ignore-path .prettierignore --check \"**/*.+(js|mjs|cjs|ts|tsx|json|md|mdx|yml|yaml)\"",
     "test": "pnpm --filter @tumaet/apollon run test",
     "test:e2e": "pnpm --filter @tumaet/webapp run test:e2e",
+    "knip": "knip",
```
FelixTJDietrich, line 47: Please update the PR discription to mention/explain knip

library/lib/components/ui/Select.tsx
Claudia-Anthropica: @FadyGergesRezk Looks good now, thanks.

library/lib/components/ui/Tooltip.tsx
Claudia-Anthropica: @FadyGergesRezk Nice, this now passes the editor theme vars through to the tooltip content.

library/lib/styles/app.css
Claudia-Anthropica: @FadyGergesRezk Codacy is still failing, so I am leaving this thread open. The GitHub Prettier check is green now, but the external Codacy status still needs to be cleared.

standalone/webapp/tests/helpers/canvas.ts
Claudia-Anthropica: @FadyGergesRezk Looks good now, the helper is back on the React Flow fit-view control and visual regression is green.

packages/ui/src/components/popover.tsx
Claudia-Anthropica: @FadyGergesRezk Build/types are green for this part now, thanks.

pnpm-workspace.yaml
```diff
@@ -1,5 +1,6 @@
 packages:
   - "library"
+  - "packages/*"
```
Claudia-Anthropica, line 3: @FadyGergesRezk [high] Adding the new workspace changed the lockfile entry for `standalone/server`: its `@tumaet/apollon` dependency moved from `link:../../library` to an injected `file:library` snapshot. In CI, `build:lib` writes `library/dist`, but `build:server` typechecks against the install-time copy and cannot resolve `@tumaet/apollon`, so both `server-tests` and the root build fail. Keep the server dependency linked to the built workspace or make the injected copy sync after the library build.

In `pnpm-workspace.yaml` / `pnpm-lock.yaml`, adding `packages/*` causes `standalone/server` to install `@tumaet/apollon` as an injected `file:library` snapshot, so CI cannot see `library/dist` after `build:lib` and `build:server` fails with TS2307. Fix the workspace/lockfile setup so `standalone/server` resolves the built `library` workspace during normal builds, or sync injected workspace dependencies after the library build.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk Several of the previous blockers are fixed now: portal theming, fit-view visuals, and the UI declaration build are green. Requesting changes because Codacy is still failing and CI now cannot build the server against `@tumaet/apollon`; see the new inline comment.

packages/ui/src/components/popover.tsx
FelixTJDietrich: Fixed in e91b455e, exactly as suggested: the exported context value is typed `(node: HTMLDivElement | null) => void` instead of `React.RefCallback<HTMLDivElement>`, so declaration emit no longer pulls in React 19's internal `UNDEFINED_VOID_ONLY` brand. `pnpm --filter @tumaet/apollon run build` (and `build:types`) is green locally.

library/lib/components/ui/Select.tsx
FelixTJDietrich: Fixed in 3c96b72f. The Select listbox now resolves the editor's `--apollon-*` theme off the trigger's closest `.apollon-editor` ancestor at open time and applies it to the portaled `Popover.Popup` (via a shared `resolveApollonThemeVars` helper that `GenericPopover` now also uses). A dark/custom embed theme carries into the open menu.

library/lib/components/ui/Tooltip.tsx
FelixTJDietrich: Fixed in 3c96b72f. The editor's Tooltip wrapper (`library/lib/components/ui/Tooltip.tsx`) now resolves the editor `--apollon-*` theme off the trigger and passes it as `style` onto the portaled `TooltipContent`, so an editor-scoped dark/custom theme reaches the tooltip.

I deliberately kept this in the library wrapper rather than `@tumaet/ui`'s primitive: the webapp consumes `@tumaet/ui`'s tooltip directly and is themed via `:root` tokens, so resolving `.apollon-editor`-scoped `--apollon-*` there would be wrong — only the embedded editor needs it, and the editor only ever renders tooltips through this wrapper.

library/lib/styles/app.css
FelixTJDietrich: Fixed in 7f5d712c. Ran the stylelint standard rules Codacy enforces (`rule-empty-line-before`, `custom-property-empty-line-before`, `color-hex-length`) over `library/lib/styles/app.css` and `packages/ui/src/styles/components.css` and cleared every annotation, and applied Prettier to `knip.json` plus the editor Storybook files `lint-and-format-check` flagged. `pnpm format:check` is green locally.

standalone/webapp/tests/helpers/canvas.ts
FelixTJDietrich: Fixed in dcf9928c. `clickFitView()` targets `.react-flow__controls-fitview` again — React Flow's built-in control doesn't expose `[aria-label="Fit to view"]` (its default label is the lowercase `fit view`), which is what timed out the BPMN visual tests.

pnpm-workspace.yaml
FelixTJDietrich: Fixed in 160ee4b7. Both `@playwright/test` and `playwright` now flow through the catalog at a single `1.59.1` (matching CI's `v1.59.1-noble` image and the committed visual baselines), so they can't drift independently and Renovate bumps one entry. The webapp resolves a single `playwright@1.59.1` and `--frozen-lockfile` validates.

Transparency: the lockfile still carries an **unreachable** `@vitest/browser-playwright(playwright@1.61.0)` peer-variant that pnpm's incremental update won't prune (no importer or snapshot depends on it, so a clean CI install never fetches it). I avoided a from-scratch lockfile regen to not churn unrelated transitive pins — happy to do that scrub in a follow-up if you'd prefer the lockfile fully clean.

package.json
FelixTJDietrich: Done — added a **Dead-code & dependency hygiene (knip)** note to the PR description. It explains that this PR introduces knip (`knip.json` + a root `pnpm knip` script) to find unused files, exports, and dependencies across every workspace during the migration (it's how the dead MUI-era components/stories were removed), that each workspace declares its real entry points + a commented allowlist, and that it runs on demand rather than as a blocking CI gate yet.

pnpm-workspace.yaml
Claudia-Anthropica: @FadyGergesRezk Still open on current head: `standalone/server` still resolves `@tumaet/apollon` as `file:library` in `pnpm-lock.yaml`, and the current `server-tests` job fails with TS2307 on every server import from `@tumaet/apollon` after `build:lib` succeeds. Please keep the server on a live workspace link or sync the injected package after the library build.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The portal-theme and fit-view blockers are resolved, but this head is still blocked by the server resolving `@tumaet/apollon` through the injected `file:library` snapshot. `server-tests` and the root build step in `lint-and-format-check` both fail with TS2307, and Codacy is still action-required, so I’m keeping this as changes requested until those checks are green.

pnpm-workspace.yaml
FelixTJDietrich: Good catch — fixed in 7442640b. Root cause: the server consumes `@tumaet/apollon` (a React component library) for its server-side render path but never declared the library's peer deps (`react`, `react-dom`, `@xyflow/react`). With `injectWorkspacePackages` on, pnpm then resolved the library for the server as an injected `file:library` copy instead of a `link:` — and that copy is snapshotted at install time, before `build:lib` writes `library/dist`. Since CI installs before building, `build:server`'s tsc couldn't resolve `@tumaet/apollon` (this would have failed server-tests, the root build, and the server Docker image once the `@tumaet/ui` build was unblocked).

The webapp already provides those peers and resolves `link:`. I declared the same on the server — they're real runtime deps of its SSR path, previously only implicit via the injected variant — which restores the `link:` resolution `main` had. Verified locally: `standalone/server/node_modules/@tumaet/apollon` is again a symlink to `../../../../library`, and `pnpm run build` (build:lib → build:server + build:webapp) is green with a valid `--frozen-lockfile`.

library/lib/styles/app.css
Claudia-Anthropica: @FelixTJDietrich @FadyGergesRezk Codacy is still action-required on the current head. The latest check reports 31 annotations, including Tailwind at-rules in `packages/ui/src/styles/theme.css` / `standalone/webapp/src/webapp.css` and remaining stylelint notices in `library/lib/styles/app.css`, `packages/ui/src/styles/components.css`, and `packages/ui/src/styles/tokens.css`, so I am leaving this unresolved until that check is green.

pnpm-workspace.yaml
Claudia-Anthropica: @FelixTJDietrich @FadyGergesRezk Nice, that explains and fixes the server workspace-link issue. The server lockfile entry is back to `link:../../library` on this head.

Claudia-Anthropica (changes requested):
@FadyGergesRezk Most of the earlier blockers are fixed now: portal theming, fit-view, declaration emit, and the server workspace-link issue all look resolved. I’m keeping changes requested because the current Codacy check is still action-required with 31 annotations, so CI is not green yet.

library/package.json
```diff
@@ -131,6 +119,9 @@
     "@dnd-kit/core": "6.3.1",
     "@dnd-kit/sortable": "10.0.0",
     "@dnd-kit/utilities": "3.2.2",
+    "@base-ui/react": "catalog:",
+    "@tumaet/ui": "workspace:*",
```
Claudia-Anthropica: @FadyGergesRezk [high] This makes the published `@tumaet/apollon` manifest depend on the private workspace package. I ran `pnpm --filter @tumaet/apollon pack --json` and the tarball rewrites this to `"@tumaet/ui": "0.0.0"`; external installs will try to fetch an unpublished private package even though the UI code is bundled into `dist`. Keep `@tumaet/ui` out of published `dependencies` and verify the packed `package.json` no longer mentions it.

In `library/package.json`, `@tumaet/ui` is declared in published dependencies even though it is private and bundled, causing the packed package to depend on `@tumaet/ui@0.0.0`. Remove it from the published dependency set, for example by making it a dev/build-only dependency or stripping it during pack, and add a pack/install check that the tarball manifest does not mention `@tumaet/ui`.

</details>

library/lib/App.tsx
```diff
@@ -18,6 +18,11 @@ import {
 } from "@/components"
 import { OverlayLayer } from "@/overlay/OverlayLayer"
 import "@xyflow/react/dist/style.css"
+// Shared, embed-safe @tumaet/ui primitives + --apollon-/--home- design tokens
+// (Tailwind-free, Preflight-free). Loaded here rather than `@import`-ed from
+// app.css so app.css stays `@import`-free for the verbatim headless-export
+// inline (see exportStyles.ts / exportSelfContained.test.ts).
+import "../../packages/ui/dist/components.css"
```
Claudia-Anthropica, line 25: @FadyGergesRezk [medium] This source import points at a generated file that does not exist in a clean checkout until `@tumaet/ui` has built. I verified `pnpm --filter @tumaet/apollon exec vitest run tests/unit/perfProbe.test.ts` fails before running tests with `Failed to resolve import "../../packages/ui/dist/components.css"`. Make the library test/dev entrypoints build or resolve the CSS first, or import a source path that Vite can process without a generated `dist` artifact.

In `library/lib/App.tsx`, the app imports `../../packages/ui/dist/components.css`, which is generated and absent in a clean checkout, so library Vitest fails resolving it. Fix this by making `@tumaet/apollon` test/dev commands generate `packages/ui/dist/components.css` before Vite loads library source, or by changing the import path/build setup so tests can resolve the stylesheet without prebuilt UI dist. Also cover `library/lib/utils/exportStyles.ts` if it relies on the same generated CSS.

</details>

library/lib/components/react/Apollon.tsx
```diff
@@ -188,9 +202,18 @@ export function Apollon(props: ApollonProps) {
...
+  const mergedStyle: CSSProperties = { ...style, ...theme }
+
   return (
     <ApollonInstanceContext.Provider value={editor}>
-      <div ref={containerRef} className={className} style={style} />
+      <div
+        ref={containerRef}
+        className={className}
+        style={mergedStyle}
+        data-theme={dataTheme}
```
Claudia-Anthropica, line 215: @FadyGergesRezk [medium] This sets `data-theme` on the React mount, but the shipped dark token deltas live only under `:root[data-theme="dark"]` in `packages/ui/src/styles/tokens.css`. So `<Apollon dataTheme="dark" />` and `new ApollonEditor(el, { dataTheme: "dark" })` stay on the light defaults unless the host also mutates `document.documentElement`, which contradicts the new public API/docs. Scope the dark token selector to the editor or mount theme attribute too, or apply `dataTheme` to the element that actually matches the selector.

In `library/lib/components/react/Apollon.tsx`, `dataTheme` is written to the mount node but `packages/ui/src/styles/tokens.css` only defines dark values for `:root[data-theme="dark"]`. Fix this by making the bundled token CSS respond to a scoped editor/mount `[data-theme="dark"]` attribute, or by applying `dataTheme` to the element that the selector targets, so `<Apollon dataTheme="dark" />` and `new ApollonEditor(el, { dataTheme: "dark" })` switch the editor to dark without requiring host root mutation.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The migration is close, but I found a few new blockers around the public library package: the packed manifest now depends on private `@tumaet/ui`, direct library tests fail before generated UI CSS exists, and `dataTheme` does not activate the bundled dark tokens when scoped to the mount. Codacy is still failing too, so requesting changes until these and CI are green.

packages/ui/src/components/popover.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  React.useEffect(() => {
+    if (!open || positioner == null) return
+
+    const isHidden = () =>
+      positioner.hasAttribute("data-anchor-hidden") ||
+      // Belt-and-braces: a detached anchor that Floating UI has not yet
+      // re-measured can leave the positioner collapsed at the origin. Treat a
+      // zero-size positioner as anchor loss too.
+      (positioner.getBoundingClientRect().width === 0 &&
```
Claudia-Anthropica: @FadyGergesRezk [high] This zero-size fallback closes every Popover/Menu as soon as it opens in jsdom, because `getBoundingClientRect()` returns 0x0 there. The current `lint-and-format-check` job fails five `@tumaet/ui` tests with no `menu`/`Inner` content rendered; rely on Base UI's `data-anchor-hidden` signal here, or gate the rect fallback so mounted popups with no layout don't self-close.

In `packages/ui/src/components/popover.tsx`, `useAnchorLossGuard()` treats a zero-size positioner rect as anchor loss, which makes Popover and DropdownMenu close immediately in jsdom and fails the UI tests. Fix this by removing or safely gating the zero-size rect fallback so tests and mounted overlays stay open, while still closing when Base UI marks the positioner with `data-anchor-hidden`.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The migration is still blocked. I found one new CI blocker in the shared overlay guard: the @tumaet/ui popover/dropdown tests now close immediately instead of rendering their content. Codacy and the previously flagged public package/theme issues also need to be green before this is ready.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this pass; I'm keeping this as changes requested because Codacy is still failing on the current head and the existing unresolved review threads still cover blockers in the migration. Please clear those threads/checks before this is ready.

packages/ui/src/components/popover.tsx
FelixTJDietrich: Fixed in fb43798c. The anchor-loss guard now gates the whole check on the popup having had layout at least once: without layout — every popup under jsdom, where getBoundingClientRect is always 0×0 — neither `data-anchor-hidden` (Floating UI sets it because it can't position) nor a zero rect is treated as anchor loss, so popups no longer self-close on mount. Once laid out, a detached anchor still closes it. All 37 @tumaet/ui unit tests pass.

library/package.json
FelixTJDietrich: Fixed in fb43798c. `@tumaet/ui` moved to devDependencies — it's inlined into dist via the vite source alias + dts `bundledPackages`, so it's build-only. Verified: the packed tarball's `dependencies` no longer lists @tumaet/ui (only the real runtime deps), and `publint` is clean. External installs no longer try to fetch @tumaet/ui@0.0.0.

library/lib/App.tsx
FelixTJDietrich: Fixed in fb43798c. The library `dev`/`test`/`test:watch`/`test:coverage` scripts now run `pnpm --filter @tumaet/ui run build:css` first (as `build` already did), so `packages/ui/dist/components.css` exists before Vite loads library source. `pnpm --filter @tumaet/apollon run test` is green on a clean checkout (1209 tests). Covers both App.tsx and exportStyles.ts (same import).

library/lib/components/react/Apollon.tsx
FelixTJDietrich: Fixed in fb43798c. The dark token block in tokens.css now also matches a scoped `[data-theme="dark"]` (not only `:root[data-theme="dark"]`), so `<Apollon dataTheme="dark" />` / `new ApollonEditor(el, { dataTheme: "dark" })` flip the editor subtree to dark without the host mutating document.documentElement. The webapp (which sets data-theme on :root) is unaffected — :root still matches.

library/lib/styles/app.css
FelixTJDietrich: app.css now passes the cited rules — custom-property-empty-line-before, rule-empty-line-before, and color-hex-length all report 0 problems (verified with stylelint-config-standard's values). The block was reflowed during the evergreen-comment pass (089725e4).

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest functional fixes I checked look good, and the @tumaet/ui unit suite passes locally. I’m keeping this as changes requested because Codacy is still action-required on the current head with CSS/stylelint/Tailwind at-rule annotations; please clear that check before merging.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this pass. The current head is still blocked by Codacy's action-required status, so I'm keeping changes requested until that check is green.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I don't have new inline comments from this pass, and the latest code changes I checked look fine. The current head still has Codacy Static Code Analysis in action_required, so I’m keeping this as changes requested until that check is green.

library/lib/components/DraggableGhost.tsx
```diff
@@ -229,26 +245,43 @@ export const DraggableGhost: React.FC<DraggableGhostProps> = ({
...
+    const entry = (event.currentTarget.firstElementChild ??
+      event.currentTarget) as HTMLElement
+    const entryRect = entry.getBoundingClientRect()
+    const ghostX = event.clientX - entryRect.left
+    const ghostY = event.clientY - entryRect.top
+    setGhostOffset({ x: ghostX, y: ghostY })
+    setGhostPosition({ x: event.clientX - ghostX, y: event.clientY - ghostY })
+
+    // Scale the ghost to the on-screen size the node will have at this zoom.
+    setGhostScale(getViewport().zoom / previewScale)
```
Claudia-Anthropica: @FadyGergesRezk [medium] This only accounts for the palette preview scale. `activitySwimlane` previews at 160x100 but drops as 400x240, so its drag ghost is still rendered at the small preview size and the cursor offset used on drop is based on the preview dimensions. Fold `dropWidth`/`dropHeight` into the ghost scale and drop-offset conversion, and add a test that drags the swimlane item rather than the first palette entry.

In `library/lib/components/DraggableGhost.tsx`, ghost scaling uses only `zoom / previewScale`, so elements with `dropWidth`/`dropHeight` such as `activitySwimlane` still drag at preview size and drop with preview-based offsets. Fix this by deriving the ghost scale and drop-offset conversion from `dropElementConfig.dropWidth`/`dropHeight` when present, and update the palette drag e2e coverage to drag the swimlane item.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk The latest drag fix improves the normal palette items, but it still misses elements whose preview size differs from their dropped size, so I left one new blocker inline. Codacy is also still failed/action-required on the current head, so keeping this as changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this pass. The current head still has Codacy Static Code Analysis in action_required, and there is still an unresolved blocker thread from the latest palette-drag review, so I’m keeping this as changes requested until those are cleared.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this pass. The newest popover/storybook changes look fine, but Codacy Static Code Analysis is still failing and the existing palette-drag blocker thread is unresolved, so I’m keeping this as changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this pass. The current head is still blocked by Codacy's action-required check, and the existing unresolved palette-drag thread in `DraggableGhost.tsx` is still a medium blocker, so I'm keeping this as changes requested.

Claudia-Anthropica (changes requested):
@FadyGergesRezk I don't have new inline comments from this pass. The current head is still blocked by Codacy's action-required CSS/stylelint annotations, and the existing unresolved `DraggableGhost.tsx` palette-drag thread still applies, so I’m keeping this as changes requested.

library/lib/components/popovers/GiveFeedbackAssessmentBox.tsx
```diff
@@ -30,7 +55,7 @@ export const GiveFeedbackAssessmentBox = ({ elementId, name, type }: Props) => {
 
     const updated: Assessment = {
       modelElementId: elementId,
-      elementType: type.toLowerCase() as "node" | "attribute" | "method",
+      elementType: resolveElementType(type),
```
Claudia-Anthropica: @FadyGergesRezk [medium] `DefaultNodeGiveFeedbackPopover` passes concrete node types like `flowchartProcess`, so this helper now maps regular node assessments to `edge`. That corrupts the `elementType` metadata for newly graded nodes; pass a separate coarse element type (`node`/`attribute`/`method`/`edge`) or let the node popover pass `type="Node"` while keeping the display label separate.

In `library/lib/components/popovers/GiveFeedbackAssessmentBox.tsx`, `resolveElementType()` maps every concrete node type to `edge` because only `node`, `attribute`, and `method` are accepted. Fix this by separating display type from stored assessment element type, ensuring default node feedback stores `elementType: "node"` and edge feedback stores `elementType: "edge"`; add or update coverage that grading a default node preserves `node`.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk I found one new assessment-data regression in the latest popover pass: default-node grading now stores node assessments as `edge`. This head also still has failing checks, so I’m requesting changes.

Claudia-Anthropica (changes requested):
@FadyGergesRezk No new inline comments from this docs-only pass. I’m keeping this as changes requested because the existing assessment/drag blocker threads still apply, `pnpm format:check` currently fails on the Flowchart/ReachabilityGraph editor stories, and Codacy is still action-required with 34 annotations.

library/lib/components/ui/portalTheme.ts
```diff
@@ -0,0 +1,45 @@
...
+  "--apollon-primary",
+  "--apollon-primary-contrast",
+  "--apollon-background",
+  "--apollon-background-variant",
+  "--apollon-hover-neutral",
+  "--apollon-gray-variant",
+  "--panel-background",
+  "--panel-shadow",
+  "--text",
+  "--popover-divider",
```
Claudia-Anthropica, line 34: @FadyGergesRezk [medium] The new assessment score pills also rely on `--apollon-assessment-*`, but those variables are not copied into body-portaled editor popovers. With `<Apollon dataTheme="dark" />` or a custom theme scoped to the editor container, the canvas badges use the scoped assessment colors while the opened feedback popover falls back to the root/light values. Add the assessment tone variables to this portal allowlist so feedback popovers, selects, and tooltips inherit the same scoped theme.

In `library/lib/components/ui/portalTheme.ts`, `APOLLON_PORTAL_THEME_VARS` omits the `--apollon-assessment-*` tokens now used by feedback score UI inside body-portaled popovers. Fix this by adding the positive/negative/zero/ungraded assessment text/background variables to the copied portal theme list, and add coverage for a scoped dark/custom editor theme if practical.

</details>

Claudia-Anthropica (changes requested):
@FadyGergesRezk I found one new scoped-theme gap in the current assessment styling; see inline. Codacy is still failing, and the existing `DraggableGhost.tsx` / `GiveFeedbackAssessmentBox.tsx` blocker threads still apply on this head, so I’m keeping this as changes requested.

library/lib/components/DraggableGhost.tsx
FelixTJDietrich: Fixed in 1dbe0e2e. The ghost scale and drop-offset now fold a per-axis drop/preview ratio (`dropWidth/width`, `dropHeight/height`) — the swimlane (160×100 preview → 400×240 drop) now drags at ~2.5×/2.4× so it matches the dropped node, and the offset maps onto the drop dimensions. Ratio is 1 for normal elements (unchanged). Added an e2e in `editor.spec.ts` that drags the swimlane and asserts the ghost reflects the larger drop size.

library/lib/components/popovers/GiveFeedbackAssessmentBox.tsx
FelixTJDietrich: Fixed in 1dbe0e2e. Removed the fragile `resolveElementType` derivation and added an explicit `elementType` prop (`node|attribute|method|edge`) that each caller passes — DefaultNode/Communication now store "node" regardless of the concrete node type, so default-node assessments are no longer mis-stored as "edge". Display label stays separate (`typeLabel`).

library/lib/components/ui/portalTheme.ts
FelixTJDietrich: Fixed in 1dbe0e2e. Added the eight `--apollon-assessment-{positive,negative,zero,ungraded}-{text,bg}` tokens to `APOLLON_PORTAL_THEME_VARS`, so a body-portaled feedback popover inherits the same scoped assessment tones as the canvas badges under `dataTheme`/a custom theme.

Claudia-Anthropica (changes requested):
@FadyGergesRezk The current code changes I checked look good, and the previously flagged drag, assessment, and scoped-theme fixes are in place. I’m still requesting changes because Codacy Static Code Analysis is failing on this head; once that check is green this should be ready from my side.

FelixTJDietrich (approved):
Will just merge it, everything should be good now

PR #763, opened 2026-06-16
fix: stop the editor freezing from unbounded document growth in long sessions

FelixTJDietrich:
### Summary

The editor could slow down and freeze during long single-user editing sessions — reported in **Firefox during exams** (which are single-user), where students had to reload the tab repeatedly and sometimes came back to an older, partial version of their diagram.

Root cause: while you drag or resize a node, the editor wrote the node into the shared diagram document (a Yjs CRDT) on *every* animation frame (~60×/s). In single-user editing an always-on undo manager holds on to each of those writes so you can undo them — so they are never reclaimed, and the document grows for as long as you keep interacting until the tab runs out of headroom and locks up. Because the growth is **per frame**, it scales with how long gestures are held, not with how many you make.

This PR bounds that growth, **keeps undo/redo and collaboration fully intact**, fixes a related "stale on reload" data-loss path in the standalone webapp, and adds tests so it can't quietly come back.

### Release note

Two changesets, one per audience (both land under **Bug Fixes**):

- **`@tumaet/apollon`** — Fixes the editor gradually slowing down and freezing during long single-user editing sessions; undo/redo and live collaboration behave exactly as before.
- **`@tumaet/webapp`** — Fixes the standalone editor sometimes losing your most recent change when you navigate away or when a save didn't reach the server.

### Performance and memory

**Yjs document growth — single-user, 30-node diagram.** Faithful simulation of the editor's per-frame write pattern with the always-on `UndoManager` (which pins every tracked struct, defeating Yjs GC):

| Measure | Before (commit every frame, pinned) | After (settle only) |
| --- | --- | --- |
| Persisted writes during a 1 s drag (~60 frames) | ~60 | 1 |
| Document growth per drag | ~4.1 KB | ~70 B |
| Document after 200 one-second drags | ~809 KB | ~16 KB |
| Growth vs. interaction | unbounded with time held | bounded per gesture |

**Browser — encoded Yjs document, Chromium (CDP), `/local` editor, 100 node drags, sampled after a forced GC:**

| Metric | Start | After 100 drags | Per drag |
| --- | --- | --- | --- |
| Encoded Yjs document | 4.8 KB | 14.1 KB | **+95 B — bounded ✓ (this PR)** |

The encoded **Yjs document — the measured freeze driver — is now bounded** (~95 B/drag, vs. unbounded per-frame before).

**A separate node-render churn — not the CRDT path, out of scope here (see *Known follow-ups*).** Profiling also showed DOM/heap growth on one gesture: selecting a *different* node and dragging it. Heap snapshots (each forces a full GC) on that gesture:

| Gesture | DOM retained | JS heap retained |
| --- | --- | --- |
| Select a different node **and drag it** | ~66 / gesture | ~120 KB / gesture |
| Click-to-select only (no movement) | 0 | 0 |
| Drag the already-selected node (no selection change) | 0 | 0 |

This is in the **node render layer, not the CRDT path** — a bare node with no Apollon content shows none, so the editor's freeze fix above is unaffected. It is *not* a confirmed unbounded leak: React-Flow maintainers attribute this class of memory growth to **app-side render patterns** — non-memoized custom nodes and expensive work that re-runs on every render (e.g. on drag) — and find it typically transient and reclaimed rather than leaked. See [xyflow/xyflow#4973](https://github.com/xyflow/xyflow/issues/4973) (closed not-reproducible, with that guidance) and [#4943](https://github.com/xyflow/xyflow/issues/4943) (closed, could-not-reproduce). The one genuine select-during-drag bug, [#4716](https://github.com/xyflow/xyflow/issues/4716), was **fixed in 12.3.2** — we ship 12.3.6. Reducing per-render churn in the node components (fewer whole-`state.nodes` subscriptions) is a worthwhile follow-up optimization, kept out of this freeze fix.

### Implementation notes

**Save only the settled position of a drag/resize.** During a gesture the editor no longer commits on every frame; it commits once, on release. On-screen movement is unaffected (local UI state, separate from what's persisted). The skip is keyed on the **actual cause** — it happens only when an undo manager is present, which is exactly single-user modelling, where the undo manager would otherwise pin every per-frame write. In collaboration there is no undo manager, so the per-frame writes are reclaimed automatically and are what drives the live remote drag. **Collaboration is untouched.**

**Undo/redo fully preserved.** A whole drag is a *single* undo step and the history is **never capped** — once the per-frame buildup is gone, keeping full history is cheap (~2,000 edits ≈ 230 KB). A supporting cleanup on the single-user path updates only the map keys that actually changed instead of clearing and rewriting the whole map.

**Alignment-guide overlay (review follow-up).** The guide `<line>`s carry an infinite CSS animation; rendering `guides` directly unmounted them on every drag-stop (and remounted per frame, as the key embedded the live position). They are now drawn from a fixed pool of index-keyed slots that stay mounted — repositioned in place while active, inert (`display:none`) when unused — so no animated line is detached per gesture. Visual output is unchanged.

**Teardown during version preview (review follow-up).** On route teardown the cleanup flushed the autosaver, but a still-open preview kept it paused and the editor's model was the previewed snapshot — so a pending pre-preview edit was skipped (or the stale snapshot would have been persisted). Teardown now exits preview first (`setPreviewMode(false)` resyncs the editor from the live Yjs doc, `exitPreview()` clears the pause), so the flush persists the real diagram. Covered by a regression test.

**More robust autosave (standalone webapp).** Autosave is change-debounced and flushes on teardown, and it now **keeps an edit dirty until a PUT actually succeeds**: an unresolved revision mismatch is no longer reported as saved, and an edit made while a version preview is open is retried once preview exits.

### Steps for testing

1. **Unit + webapp tests:** `pnpm --filter @tumaet/apollon test` and `pnpm --filter @tumaet/webapp test`
2. **Performance budget** (Chromium + Firefox): `pnpm --filter @tumaet/webapp exec playwright test --project=perf --workers=1` (and `test:perf:firefox`) — asserts the encoded document stays bounded across many drags.
3. **Single-user:** drag a node back and forth on a ~30-node diagram for a minute — stays responsive; then undo many steps — it walks all the way back.
4. **Alignment guides:** drag a node so it lines up with another — the guide lines still appear while aligned and vanish on release, with no flicker.
5. **Preview teardown:** edit, open a version preview, then navigate away while still in preview — the edit is persisted (reload shows it).
6. **Collaboration** (`?view=COLLABORATE`, two tabs): drag a node — the other tab sees it move **live**, exactly as before.

### Screenshots / screencasts

n/a — no visual change. Alignment guides, collaboration, and undo/redo behave exactly as before (the visual-regression suite passes unchanged).

### Known follow-ups (out of scope)

- **Node-render churn on select-and-drag** (table above): ~66 DOM + ~120 KB heap per "select a different node and drag it" gesture (clicks and same-node drags show none; a bare node shows none). Not the CRDT path and not a confirmed unbounded leak — React-Flow maintainers attribute this class to app-side render patterns and find it typically transient ([#4973](https://github.com/xyflow/xyflow/issues/4973), [#4943](https://github.com/xyflow/xyflow/issues/4943), both closed not-reproducible; the real select-during-drag bug [#4716](https://github.com/xyflow/xyflow/issues/4716) is fixed in 12.3.2 ≤ our 12.3.6). A worthwhile follow-up is to cut per-render churn (e.g. drop the whole-`state.nodes` subscription that re-renders every node each drag frame), kept out of this freeze fix.
- Collaboration keeps its original per-frame save/broadcast (its buildup is auto-reclaimed without an undo manager, and far milder); if a long *collaborative* session ever needs tuning it should be throttled, keeping drag live, not restructured.
- Durable save against a hard tab-kill (server owning the converged document) and the browser's 64 KiB unload-request cap are server-side follow-ups.
- A latent, currently-inert width-clamp bug in the class node (a measured-height drop) is noted for a separate fix.

### Checklist

- [ ] Linked to a related issue (if applicable) — _from a maintainer bug report; no tracking issue._
- [x] Added a changeset whose summary is written in the user's voice — `@tumaet/apollon` + `@tumaet/webapp`
- [x] PR title's Conventional Commit type (`fix`) matches the kind of change — it groups the release note under **Bug Fixes**
- [x] Tests added or updated
- [ ] Documentation updated (if applicable) — _N/A: no public API change._
- [ ] Screenshots or screencasts attached (if a UI change) — _N/A: no visual change._


standalone/webapp/src/services/createDiagramAutosaver.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+        if (err instanceof ApiError && err.code === "REVISION_MISMATCH") {
+          // Single-editor mismatch (or retries exhausted): adopt the server's
+          // hint so the NEXT save carries a fresh If-Match rather than
+          // clobbering a concurrent writer with no guard.
+          const meta = err.meta as { currentHeadRev?: number } | undefined
+          if (typeof meta?.currentHeadRev === "number") {
+            lastHeadRev = meta.currentHeadRev
+          }
+          return
```
Claudia-Anthropica: @FelixTJDietrich [high] This path treats a `REVISION_MISMATCH` as a successful save: `putOnce()` returns, then `save()` calls `onSaved()` and clears `diagramIsUpdated`. In single-editor mode the PUT never reached the server, so the local edit is no longer dirty and the unload flush will skip it; keep it dirty and retry after adopting `currentHeadRev` instead.

In `standalone/webapp/src/services/createDiagramAutosaver.ts`, `REVISION_MISMATCH` returns from `putOnce()` and is treated as a successful save. Fix this by keeping the autosaver dirty and retrying with the adopted `currentHeadRev` instead of calling `onSaved()` for an unsaved model; add a regression test for single-editor mismatch.

</details>

standalone/webapp/src/services/createDiagramAutosaver.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+        throw err
+      }
+    }
+  }
+
+  const save = async () => {
+    clearTimers()
+    if (!dirty || opts.isPaused()) return
```
Claudia-Anthropica, line 146: @FelixTJDietrich [medium] When a debounce/maxWait fires during version preview, `save()` clears both timers and returns while leaving `dirty` true. After the user exits preview there is no timer left to retry, so the edit can remain unsaved indefinitely until another model change or teardown; re-arm the save when paused, or flush the autosaver when preview exits.

In `standalone/webapp/src/services/createDiagramAutosaver.ts`, `save()` clears timers before checking `opts.isPaused()`, so dirty edits that hit a preview pause are never retried after preview exit. Fix this by scheduling or flushing the pending dirty save when the editor unpauses; add a test covering paused save followed by unpause without a new model change.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The Yjs-side drag fix looks solid, but the new autosaver currently marks a couple of unsaved cases as settled. Please fix the revision-mismatch and preview-pause paths inline so dirty edits cannot silently sit or be cleared without a successful PUT.

standalone/webapp/src/services/createDiagramAutosaver.ts
Claudia-Anthropica: @FelixTJDietrich Still seeing this one in the current head: the single-editor mismatch still returns from `putOnce()`, so `save()` calls `onSaved()` even though no PUT succeeded. The regression test at `createDiagramAutosaver.test.ts` also locks in that unsaved first edit as settled instead of keeping it dirty or retrying.

standalone/webapp/src/services/createDiagramAutosaver.ts
Claudia-Anthropica: @FelixTJDietrich This is still open too. A save attempt that fires while preview is active still clears both timers and leaves the dirty flag with no retry path when preview exits unless another edit or teardown happens later.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two autosaver issues from the previous review are still present in the current head. I replied on both threads with the remaining gaps; please keep dirty edits dirty until a PUT actually succeeds and add a retry path after preview pauses.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit fixes the in-flight serialization race, but the two existing autosaver correctness threads are still unresolved in this head. Single-editor revision mismatch is still counted as saved without a successful PUT, and preview-paused dirty edits still have no retry path after unpause.

standalone/webapp/src/services/createDiagramAutosaver.ts
FelixTJDietrich: Fixed in 24c3e73b. `putOnce()` now returns whether the model was actually persisted; an unresolved `REVISION_MISMATCH` (single-editor, or rebase retries exhausted) returns `false`, so `save()` does not call `onSaved()` and keeps the edit dirty — it retries on the next change or flush with the adopted `currentHeadRev` as `If-Match`. A regression test asserts a single-editor mismatch stays dirty (a flush with no new change re-PUTs the adopted rev) and that exhausted rebases never report the edit saved.

standalone/webapp/src/services/createDiagramAutosaver.ts
FelixTJDietrich: Fixed in 24c3e73b. The trailing-edge save timer now re-arms itself while a version preview is open (instead of firing, finding it paused, and dropping the timer), so a dirty edit persists once preview exits even with no further model change. Regression test: `persists a paused dirty edit after preview exits, with no new change`.

standalone/webapp/src/services/createDiagramAutosaver.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

standalone/webapp/src/services/createDiagramAutosaver.ts
Claudia-Anthropica: @FelixTJDietrich Nice, that handles it.

Claudia-Anthropica (approved):

standalone/webapp/src/pages/ApollonWithConnection.tsx
```diff
@@ -305,9 +286,12 @@ export const ApollonWithConnection: React.FC = () => {
       abort.abort()
       setEditor(undefined)
       wsManagerRef.current?.cleanup()
-      if (syncIntervalRef.current) {
-        clearInterval(syncIntervalRef.current)
-      }
+      // Persist any pending debounced edits before tearing down (e.g. SPA
+      // navigation): flush() reads the model synchronously, so it captures the
+      // latest state while the editor is still alive, then dispose() stops it.
+      void autosaverRef.current?.flush()
```
Claudia-Anthropica, line 302: @FelixTJDietrich [high] `flush()` still no-ops while `isPaused()` is true, so this teardown path drops a dirty pre-preview edit if the user navigates away while a version preview is open. The preview overlay has replaced `editor.model`, but the Yjs doc still has the live diagram; exit preview or otherwise read the live Yjs model before flushing on teardown so the pending edit reaches the server.

In `standalone/webapp/src/pages/ApollonWithConnection.tsx`, cleanup calls `autosaver.flush()` while the autosaver is paused during version preview, so pending dirty edits are not persisted on route teardown. Fix this by restoring the editor out of preview or otherwise flushing the live Yjs/current model before disposing the autosaver, and add a regression test that edits, enters preview, navigates away, and persists the edit.

</details>

library/lib/components/AlignmentGuides.tsx
```diff
@@ -38,9 +43,10 @@ export const AlignmentGuides = () => {
         height: "100%",
         pointerEvents: "none",
         zIndex: 999,
+        visibility: hasGuides ? "visible" : "hidden",
       }}
     >
-      {guides.map((guide: AlignmentGuide) => {
+      {(guides ?? []).map((guide: AlignmentGuide) => {
```
Claudia-Anthropica: @FelixTJDietrich [medium] This keeps the host `<svg>` mounted, but `clearGuides()` still makes `(guides ?? [])` empty and React removes all of the animated `<line>` elements on every drag stop. If Firefox is retaining detached SVG nodes, the guide lines are the nodes with the SVG animation and they still churn per gesture; keep/reuse the last rendered line elements while hiding the overlay, or render stable line slots instead of clearing the subtree to empty.

In `library/lib/components/AlignmentGuides.tsx`, the component hides the SVG but still unmounts every guide `<line>` when `guides` is empty. Fix this by keeping or reusing the rendered line nodes across empty-guide states while toggling visibility, so drag stop does not detach animated SVG nodes.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The Yjs and autosaver fixes are much stronger now, but I found two remaining edge cases: teardown while preview is active can still lose a pending edit, and the alignment-guide SVG change does not actually keep the animated guide nodes mounted. Please address those before this lands.

library/lib/components/AlignmentGuides.tsx
Claudia-Anthropica: @FelixTJDietrich Still open in the current head. The component still renders `(guides ?? []).map(...)`, so when `clearGuides()` empties the array React removes all guide `<line>` nodes; only the wrapper SVG stays mounted.

standalone/webapp/src/pages/ApollonWithConnection.tsx
Claudia-Anthropica: @FelixTJDietrich Still open in the current head. Cleanup still calls `flush()` while `preview` keeps `isPaused()` true, and `save()` returns before reading or persisting anything, so route teardown during preview can still drop the dirty edit.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two open threads from the last review are still unresolved in the current head. Please fix the preview-teardown autosave path and the alignment-guide line-node churn before this lands.

standalone/webapp/src/pages/ApollonWithConnection.tsx
FelixTJDietrich: Fixed in 4c7112ec. Teardown now exits version preview **before** flushing: `setPreviewMode(false)` resyncs the editor from the live Yjs doc (so `getModel()` returns the real diagram, not the previewed snapshot) and `exitPreview()` clears the pause, so `flush()` persists the pending pre-preview edit instead of skipping it. Regression test added in `ApollonWithConnection.test.tsx` ("persists a pending edit when navigating away during a version preview"): edit → enter preview → navigate away → asserts the **live** model is PUT and the stale snapshot is not.

library/lib/components/AlignmentGuides.tsx
FelixTJDietrich: Good catch — keeping only the wrapper `<svg>` mounted still unmounted the animated `<line>`s every drag-stop (and remounted them per frame, since the React key embedded the live position). Fixed in e8659d9d: `AlignmentGuides` now renders a **fixed pool of index-keyed line slots** that stay mounted across gestures — each slot is repositioned in place as guides move, and goes inert (`display: none`, no animation) when unused, so no animated line node is ever detached. I first tried reuse-last-rendered via a ref, but reading a ref during render is rejected by the React Compiler lint rule (`react-hooks/refs`), so the stable-slot pool is the clean form. Visual output is unchanged; the visual-regression suite passes.

Claudia-Anthropica: @FelixTJDietrich Nice, that handles it.

standalone/webapp/src/pages/ApollonWithConnection.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):

PR #764, opened 2026-06-16
feat(library): migrate to React 19 and enable the React Compiler

FelixTJDietrich:
### Summary

Upgrades the whole monorepo from **React 18.3 → 19.2** and turns on the **React Compiler** (library, webapp, and both VS Code webviews). The compiler auto-memoizes components, which lets us delete a layer of hand-written performance plumbing and the `forwardRef` ceremony, and — guided by the compiler's new lint rules — fix a real correctness bug along the way.

Highlights:

- **`<Apollon>` now takes `ref` as a plain prop** (React 19 ref-as-prop); the `forwardRef` wrapper is gone. Consumers passing a `ref` need no changes.
- **Popover anchoring fixed at the root.** ~50 node components + the edge toolbar read `someRef.current` *during render* to position their popover — which is `null` on first render and never re-renders, so the popover was briefly anchorless. They now use a small `usePopoverAnchor` hook (callback ref → state), which the compiler's `react-hooks/refs` rule requires and which fixes the latent bug. The edge anchor is lifted out of 13 edge types into `CommonEdgeElements`.
- **Manual memoization removed** — `useMemo`/`useCallback`/`memo` wrappers and a hand-rolled stale-closure hash (`pointsKey`) are deleted now that the compiler handles them.
- **Net −70 lines** despite the new hook + test.

Release notes for `@tumaet/apollon` / `@tumaet/webapp` are in the changeset.

### Fixes found in review

Two regressions reported in review, both the **same root cause** — a value read non-reactively during render that the React Compiler then memoizes (so it never re-reads) — and fixed the same way (subscribe reactively):

- **Alignment ("helper") lines rendered offset.** `AlignmentGuides` read the viewport via the non-reactive `getViewport()` during render → stale once memoized. Now subscribes to the live viewport with `useStore(transform)`.
- **Edit-popover controls showed stale values** — class association multiplicity/role didn't update, BPMN node inputs/dropdowns didn't reflect edits, and Swap didn't update the popover. The node/edge edit popovers read `getEdge`/`getNode` imperatively during render. Now they read reactively via shared `useReactiveEdge` / `useReactiveNode` / `useReactiveNodeName` hooks (17 popovers). Edge-toolbar Edit/Delete and the Swap control also got `aria-label`s.

Verified working (so they were not regressed): live collaboration (`?view=COLLABORATE`, two-client presence + cursors) and diagram export — webapp SVG/PNG/JSON and the **server-side** `conversion-service` (`exportModelAsSvg`) integration test.

### Implementation notes

- **Pins** (`pnpm-workspace.yaml` catalog): `react`/`react-dom` 19.2.7, `@types/react(-dom)` 19.2.x, `eslint-plugin-react-hooks` 7.1.1, `babel-plugin-react-compiler` 1.0.0 (pinned exact — its output isn't SemVer-stable yet). `@emotion/*` → 11.14 (the 11.12 release carries Emotion's React 19 ref fix; the library peer floor moves to `^11.12.0`). Kept Vite 6 / `@vitejs/plugin-react` 4.3.x deliberately (the compiler wires through Babel; the Vite 8 / plugin-react 6 oxc path is a separate migration).
- **Library publishing:** compiled with `target: "19"`, so the peer build emits `import { c } from "react/compiler-runtime"` (React 19's built-in) and keeps it external — no polyfill is bundled. `@tumaet/apollon` targets React 19 (`peerDependencies: ^19.0.0`). Bumped as a **minor**: the component's call-site API is unchanged (ref-as-prop is transparent), only the React peer requirement moved.
- **Lint calibration:** `eslint-plugin-react-hooks` v7 adds the compiler rules at `error`. `set-state-in-effect` stays `error` globally, scoped to `warn` only in the files where the pattern is inherent and correct (post-paint SVG measurement, Yjs subscription sync, post-commit tracking) — new occurrences elsewhere still fail. A few `immutability`/`refs`/`purity` false positives (imperative editor API in effects, latest-closure refs) carry single-line, documented disables.
- **Scope discipline:** intentionally did **not** bump MUI, `react-router`, or `@xyflow/react` — their existing versions already support React 19, so those would be unrelated churn. (`@tumaet/apollon` keeps MUI; the MUI→Base UI migration is a separate effort.)
- **Follow-ups:** `useEdgeLineJumps` now depends on its producers keeping their point arrays referentially stable (documented at the call site, true today, not type-enforced). The edge toolbar's prop wiring is covered by TypeScript + the shared-hook test rather than a full render test (a `CommonEdgeElements` render test would need 3 store providers to assert "it renders").

### Steps for testing

1. `pnpm install && pnpm build` — library (both passes), server, and webapp build clean.
2. `pnpm test` — 950 library unit tests; `pnpm --filter @tumaet/webapp test` — 167 webapp tests.
3. `pnpm lint && pnpm format:check` — clean (0 errors; remaining warnings are the intentional `exhaustive-deps` / scoped `set-state-in-effect`).
4. Optional: `npx react-compiler-healthcheck@latest` in `library` / `standalone/webapp` → 259/259 and 82/82 components compile; React DevTools shows the "Memo ✨" badge on `<Apollon>` and the node/edge components.
5. **Edit popovers** (the review fixes): open a Class diagram, select an association, edit source/target **multiplicity & role** → labels update on canvas and the controls stay in sync after **Swap**; edit a **BPMN** Task name/type/marking. Drag a node so alignment guides appear → guides sit on the node edges.
6. Smoke-test the editor (webapp + both VS Code webviews): create/edit a diagram, open element popovers, drag edges — popovers anchor correctly on first open.

### Screenshots / screencasts

N/A — no UI changes; this is a runtime/build upgrade and behavior-preserving refactor. The one user-visible improvement (popovers anchoring on first open instead of after a later render) is a bug fix, not a visual change.

### Checklist

- [x] Linked to a related issue (if applicable) — N/A
- [x] Added a changeset (`@tumaet/apollon` minor, `@tumaet/webapp` minor)
- [x] Tests added or updated (new E2E `class-edge-multiplicity-role` for popover editing/Swap; `usePopoverAnchor` behavioural test; `<Apollon>` ref test green against ref-as-prop; full E2E suite 103 passing)
- [ ] Documentation updated (if applicable) — N/A
- [ ] Screenshots or screencasts attached (if a UI change) — N/A


tamang29:
1) ClassDiagram's multiplicity and role state is not updating on edit.
Edit: All the edge label updates are not working

https://github.com/user-attachments/assets/270f8be5-63c0-4a45-bbbf-93e29345ef33

2) Input field, dropdown selection in BPMN is not working:
- Try updating Task name, type, marking
- cannot change to initial state
- selected state is not updated.
- most of the edge label update and input fields for any kind of diagram is not working as expected
Screenshot 2026-06-16 at 16 51 45

3) **Export service for external consumer is not working**: tested locally

PR #766, opened 2026-06-18
feat(library): undo/redo in shared (collaborative) diagrams

FelixTJDietrich:
### Summary

Shared (collaborative) diagrams had no undo/redo: the undo manager was created only for single-user editing, so in a collaboration session ⌘/Ctrl+Z did nothing and the toolbar buttons never appeared. This enables undo/redo in collaboration — scoped per-user — without re-introducing the editor freeze that #763 fixes.

The freeze and live collaboration were entangled: dragging wrote the node into the shared document on every animation frame, and an always-on undo manager pins each frame, growing the document without bound. So naively enabling undo in collaboration would bring the freeze back. The fix keeps the live experience but changes where the per-frame data flows: transient drag/resize frames are no longer persisted in either mode, and a peer's in-progress movement travels over the ephemeral awareness channel (the same one that powers live cursors) instead of the document; only the settled position/size is committed, as a single edit.

### Release note

> Adds undo/redo to shared (collaborative) diagrams. Undo and redo now work during a collaboration session, scoped to each person's own edits — undo reverts only your own changes, never a teammate's, and never overwrites an element someone else is editing. A whole drag or resize counts as a single step, your selection comes back with the change you undo, and everyone keeps seeing each other's edits move in real time.

(`@tumaet/apollon`, `minor` → **Features**.)

### Implementation notes

- **Undo is per-user ("local undo").** The `Y.UndoManager` tracks only locally-authored writes (`trackedOrigins`), so ⌘Z reverts your edits and never a collaborator's — matching how Figma, Google Slides, and the Yjs docs recommend collaborative undo should behave. Undoing an element a teammate concurrently changed safely no-ops (Yjs's default `ignoreRemoteMapChanges`); selection is restored via stack-item meta. Undo/redo transactions were already broadcast to peers by the sync layer — this turns the manager on.
- **Live gestures move to awareness.** `useRemoteDraggingNodes` overlays peers' in-progress positions/sizes onto React Flow's controlled `nodes`, so remote drags stay live with no per-frame document write and nothing entering undo history.
- **Nested-resize freeze, also closed.** Resizing a node inside a container auto-grows the parent via a per-frame full-node `replace` that bypassed #763's transient guard — a latent freeze for nested nodes that this change would otherwise have extended to collaboration. It's now skipped per-frame (a self-clearing guard on React Flow's live dragging/resizing flags) and committed once at resize end.
- **Trade-offs:** the live-drag broadcast lives in the diagram store rather than the React layer like the other awareness writes, because `onNodesChange` is the only seam that sees both drag *and* resize transient frames; documented at the call site. The overlay reader follows the same effect+subscription pattern as the rest of `CollaborationLayer` for consistency.

**Addressed in review (both threads resolved):**
- Peer-supplied `draggingNodes` is now sanitized in `narrowState` (a malformed value could otherwise throw in the overlay reader and break collaboration for everyone reading that peer).
- A resizing child's parent auto-grow is now broadcast live over awareness too, so the container grows smoothly for peers instead of jumping at settle.

**Honest caveats (follow-ups, not blockers):**
- The unit tests cover the store/CRDT semantics and freeze-safety; the remote-overlay *rendering* (awareness → React Flow) has no automated coverage — please run the manual collaboration steps. A Playwright collab-drag test is a good follow-up.
- If a peer disconnects mid-drag, their overlay clears when their awareness entry times out (~30 s) and the node returns to its last committed position.

**Stacked on #763** (`apollon-firefox-freeze`).

### Steps for testing

1. **Unit tests:** `pnpm --filter @tumaet/apollon test` — see `library/tests/unit/perfProbe.test.ts` (collaboration undo, nested-resize live broadcast, preview-gate, selection restore) and `collaboration.test.ts` (peer-payload sanitizing).
2. **Collaboration (two tabs, `?view=COLLABORATE`):** drag a node in tab A — tab B sees it move live; drop, then ⌘/Ctrl+Z in A reverts it. Edit in B, then undo in A — B's edit is untouched.
3. **Nested resize:** resize a flowchart node inside a container — the container grows live for peers and the session stays responsive.
4. **Single-user regression:** drag/resize for a minute — responsive; undo walks all the way back.

### Screenshots / screencasts

No visual change. In collaboration the existing undo/redo toolbar buttons and ⌘Z / ⌘⇧Z shortcuts now work as they do in single-user; live remote drag looks identical to before.

### Checklist

- [ ] Linked to a related issue — _not tied to a specific issue; addresses a gap noticed alongside the #763 freeze work._
- [x] Added a changeset whose summary is written in the user's voice (`@tumaet/apollon`, `minor`)
- [x] PR title's Conventional Commit type (`feat`) matches the kind of change — groups the release note under **Features**
- [x] Tests added or updated
- [ ] Documentation updated — _N/A: no public API change; undo/redo was already exposed and now works in collaboration._
- [ ] Screenshots or screencasts — _N/A: no visual change._


library/lib/hooks/useRemoteDraggingNodes.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const buildOverlay = (
+  states: Map<number, CollaborationState>,
+  localClientId: number
+): RemoteDraggingOverlay => {
+  const overlay: RemoteDraggingOverlay = new Map()
+  for (const [clientId, state] of states) {
+    if (clientId === localClientId) continue
+    const draggingNodes = state.draggingNodes
```
Claudia-Anthropica, line 42: @FelixTJDietrich [medium] This assumes every peer-supplied `draggingNodes` value is an iterable array. `YjsSync.narrowState()` currently leaves this field as-is, so a malformed or stale awareness state like `{ draggingNodes: {} }` throws here and breaks the collaboration layer for everyone reading that peer. Please narrow this field before it reaches the overlay reader, the same way viewport/following are sanitized.

In `library/lib/hooks/useRemoteDraggingNodes.ts`, `buildOverlay` iterates `state.draggingNodes` without validating the peer-supplied awareness value. Fix this by sanitizing `draggingNodes` in `library/lib/sync/yjsSync.ts` so only arrays of entries with a string `id`, finite numeric `position.x`/`position.y`, and optional finite numeric or null `width`/`height` are exposed; drop malformed values.

</details>

library/lib/store/diagramStore.ts
```diff
@@ -490,13 +606,82 @@ export const createDiagramStore = (
...
+              (n) => n.dragging || n.resizing
+            )
+            const resizeSettled = filteredChanges.some(
+              (c) => c.type === "dimensions" && c.resizing === false
+            )
+
             transactStore(() => {
               for (const change of filteredChanges) {
                 if (change.type === "add" || change.type === "replace") {
+                  if (change.type === "replace" && gestureInFlight) continue
```
Claudia-Anthropica, line 692: @FelixTJDietrich [medium] This skips the per-frame parent `replace` writes during a nested resize, but those frames also never get published into awareness above. In collaboration that means a peer resizing a child broadcasts only the child dimensions; the remote container stays at its old size and then jumps when `resizeSettled` reconciles. Keep the document skip, but include these gesture-time `replace` items in the transient awareness payload.

In `library/lib/store/diagramStore.ts`, gesture-time parent `replace` changes are skipped from Yjs persistence but are not broadcast over awareness. Fix this by adding in-flight `replace` changes that carry changed geometry to `draggingNodesPublisher` while still skipping their Yjs writes, so remote peers see parent auto-grow live during nested resize.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The main undo/redo direction looks solid and the focused library tests pass locally. I’m requesting changes for two collaboration edge cases: the new awareness payload needs narrowing, and nested-resize parent auto-grow is no longer broadcast live to peers.

library/lib/hooks/useRemoteDraggingNodes.ts
FelixTJDietrich: Good catch — fixed in bc925e00. `draggingNodes` is now narrowed in `YjsSync.narrowState` via a new `sanitizeDraggingNodes` (mirroring `sanitizeCollaborationViewport`): a non-array value returns `null`, and malformed entries (missing string `id`, non-finite `position`, non-finite `width`/`height`) are dropped, so the overlay reader only ever iterates well-formed data. Covered by new unit tests in `collaboration.test.ts`.

library/lib/store/diagramStore.ts
FelixTJDietrich: Agreed — that was a real regression in the live experience, not just an edge case. Fixed in bc925e00: in-flight `replace` geometry (the parent auto-grow) is now included in the awareness payload while still being skipped from the per-frame Yjs write, so peers see the container grow live and it no longer jumps at `resizeSettled`. The nested-resize test now asserts the parent grow is broadcast up to its final size in collaboration.

library/lib/hooks/useRemoteDraggingNodes.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

library/lib/store/diagramStore.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed, and CI is green. Approving.

Claudia-Anthropica (approved):
@FelixTJDietrich Re-reviewed the current head after the latest commits. The collaboration undo/redo path and awareness-based live gesture handling look consistent, and CI is green.

Claudia-Anthropica (approved):
@FelixTJDietrich Re-reviewed the current head after the main merge and added two-peer undo tests. The collaboration undo path and transient awareness handling still look sound; focused library tests and library lint pass locally.

PR #770, opened 2026-06-19
ci: parallelize e2e and move visual/perf suites to a nightly workflow

FelixTJDietrich:
### Summary

Apollon's per-PR CI was dominated by the end-to-end job at **~10 min**. The root cause was a single line in the Playwright config — `workers: 1` in CI — which forced the entire functional suite to run one test at a time, even though the runners have 4 cores.

This PR:

1. **Lets the functional e2e suite run in parallel** (`workers: "50%"`, Playwright's recommended CI default — 2 workers on a 4-vCPU runner).
2. **Tiers the heavier, slower-moving guards to a nightly workflow** — visual regression and the chromium document-growth budget — so they no longer gate every PR.

The **Firefox** document-growth budget — the regression guard for the Apollon-in-Artemis exam freeze (#763) — **stays on every PR**. Firefox is the engine the freeze was reported on, and the byte/write metric it checks is engine-independent, so the chromium copy is only a cross-check and is the one perf job that moves to nightly.

Net effect: the slowest PR job runs roughly half as long, with **no loss of the correctness gates that matter**.

### Release note

n/a — no Changesets-tracked package changed (CI / test-tooling only).

### Implementation notes

- **Playwright config** (`playwright.config.ts`): `workers: 1 → "50%"`, `retries: 2 → 1`. The `perf` / `perf-firefox` projects pin `workers: 1` themselves — serial measurement is the one place the new global parallelism must not leak — so the now-redundant `--workers=1` is dropped from the `test:perf` scripts (the config is the single source of truth).
- **Composite action** (`.github/actions/setup`): collapses the repeated `pnpm + Node + frozen install` blocks shared across the PR-health and nightly jobs into one place. It runs *after* `actions/checkout` — a local composite can't perform the initial checkout itself.
- **Nightly workflow** (`.github/workflows/nightly-checks.yml`): runs visual + chromium-perf on a 03:00 UTC cron and on demand (**Run workflow** against any branch). It opens a single **deduplicated tracking issue** if a *scheduled* run fails, so a red nightly isn't silently ignored.
- **Why no PR label for on-demand visual:** an earlier draft used a `run-visual` label, but that needs a `pull_request` trigger that posts a permanently-skipped check on every PR (and the label didn't exist). "Run workflow against the branch" gives the same capability with zero per-PR noise.
- **Out of scope (follow-ups):** the per-PR Docker image build (multi-arch incl. emulated arm64, via the shared `ls1intum/.github` workflow) is the largest remaining per-PR cost but needs an org-level change; and a release-time guard that blocks publishing while a nightly-failure issue is open.

> **Branch-protection note for maintainers:** `main` currently has no required status checks. The intended required check is the gate job, whose **rendered name is `PR Health Gate`** (not the job id `pr-health-gate`). Set that one as required, and do **not** mark any `nightly-checks` job required — they never run on PRs.

### Steps for testing

1. On this PR, confirm **PR Health Checks** runs **without** a `visual-regression-tests` job, that `e2e-tests` uses multiple Playwright workers, and that it is meaningfully faster than before.
2. Confirm **`perf-firefox-tests` still runs and passes** (the exam-freeze guard).
3. Confirm **`nightly-checks` does not appear as a check on this PR** (it's schedule/manual only — this is the point of the redesign).
4. After merge: **Actions → Nightly Checks → Run workflow**, pick a branch, and confirm visual + chromium-perf run green. Optionally break a nightly job on a throwaway branch to confirm the deduplicated failure issue is opened.

### Screenshots / screencasts

n/a — no UI change.

### Checklist

- [ ] Linked to a related issue — n/a (no tracking issue; proactive CI cleanup)
- [x] No changeset needed — this PR doesn't touch a Changesets-tracked package's behavior (CI / test config only)
- [x] PR title's Conventional Commit type (`ci`) matches the kind of change
- [ ] Tests added or updated — n/a (this changes how tests *run*, not test logic)
- [x] Documentation updated (visual-tests, contributor overview, deployment docs)
- [ ] Screenshots / screencasts — n/a (no UI change)


Claudia-Anthropica (approved):
@FelixTJDietrich Looks good, nice cleanup of the PR-health path while keeping the Firefox perf guard in place. CI is green on this SHA.

PR #781, opened 2026-06-22
fix: make diagram sidebar more compact for iOS app

tamang29:
Related issue: #779 
### Summary
- Compact the diagram element sidebar on mobile viewports.
- Reduce whitespace between palette entries in portrait and landscape.
- Preserve the 44px minimum touch target and existing desktop layout.
- Automatically update the layout when device orientation changes.
- Add unit and E2E regression coverage.

### Release note
See more of the canvas on mobile with a compact element palette in both portrait and landscape.

### Screenshots / screencasts

Screenshot 2026-06-22 at 16 27 12
Screenshot 2026-06-22 at 16 27 33

### Checklist

- [x] Linked to a related issue (if applicable)
- [x] Added a changeset whose summary is written in the user's voice (`pnpm changeset`, [how](https://ls1intum.github.io/Apollon/contributor/development/release-notes/)) — or this PR doesn't touch a Changesets-tracked package (`@tumaet/apollon`, `@tumaet/webapp`, `@tumaet/server`)
- [x] PR title's Conventional Commit type (`feat`/`fix`/…) matches the kind of change — it groups the release note
- [x] Tests added or updated
- [ ] Documentation updated (if applicable)
- [x] Screenshots or screencasts attached (if a UI change)

PR #780, opened 2026-06-22
feat(webapp): make the iOS app App Store-ready to replace the legacy app

tamang29:
### Summary

This makes the redesigned Capacitor-based iPhone/iPad app ready to **replace the legacy native Apollon app on the same App Store listing** (`de.tum.cit.ase.apollon`), and does the supporting work that a clean, professional store submission needs. Apollon stays **web-first** — the web app is the primary product; iPhone/iPad, VS Code, and the embeddable library are surfaces of one platform.

It started as an icon/splash refresh (#780) and grew into the whole replacement. It's a large PR, but a cohesive one; the pieces are:

- **Legacy branding continuity** — the native launch screen is a byte-for-byte reproduction of the old app's (Apollon wordmark + TUM logo, same colour and artwork), and the Capacitor splash matches it so the cold-launch reads as one familiar brand.
- **Automated App Store screenshots** — real iPhone 17 Pro Max + iPad Pro 13″ captures via `fastlane snapshot`/XCUITest (not mockups), normalised to App-Store-safe sRGB with no alpha, optionally framed in official Apple product bezels, generated in CI. Screenshots seed this repo's README-hero diagram and two existing fixtures **without changing any release template**.
- **Metadata + release automation** — localised `fastlane/metadata`, and lanes that set almost everything code-first (category, age rating, export compliance, review contact) so little is clicked in App Store Connect.
- **Native polish** — a root-cause fix for iOS safe-area handling (cards no longer tuck under the floating nav bar; the header seats correctly on iPad), and the status bar is now shown (theme-aware) like a first-class native app.
- **Privacy alignment** — the web app no longer requests its font from Google; Inter is self-hosted, so no diagram session contacts a third party. The privacy statement is updated to cover web + mobile.
- **Public content** — web-first README/docs/landing, a new user **Support** page, and contributor docs.
- **Over-the-air live updates** — self-hosted, first-party, signed web-bundle updates so interface fixes can ship without an App Store review (details below).

Removing the committed personal signing team (a reviewer-blocker) and adding the screenshot UI-test target are the only changes to the Xcode project.

### Release note

Rebuild the iPhone and iPad app on the redesigned editor with the familiar Apollon branding, and stop the web app from loading its font from Google — it is now self-hosted.

### Implementation notes

[...]

PR #782, opened 2026-06-25
feat(library): add a fully-external build entry and externalize Yjs

krusche:
### Summary

Apollon's default build is a single self-contained file — perfect for a `<script>` tag or CDN, but it inlines React, MUI, xyflow, and Apollon's own runtime deps. For a **host that embeds Apollon with a bundler** (Angular, Vue, Svelte, React), that means shipping — and being unable to de-duplicate or audit — a second private copy of each. And because Yjs is a CRDT whose runtime relies on module-level singletons, a host that already uses Yjs ends up running two Yjs copies on the same page.

This PR fixes both, **without changing the existing entries**:

1. **New `@tumaet/apollon/external` entry** — the same imperative `ApollonEditor` API as the default entry, but with **every** runtime dependency external (the React family plus `@dnd-kit` / `zustand` / `uuid` / `@chenglou/pretext`; `yjs` / `y-protocols` were already external). The host's own bundler then resolves, de-duplicates, and gets full supply-chain / SBOM attribution for each one from its `node_modules`, instead of a copy buried invisibly in Apollon's bundle.
2. **`yjs` and `y-protocols` are now required peer dependencies of every build.** They are Apollon's document model and undo/redo engine (used by every editor, collaboration or not), and they rely on module-level singletons — so one shared instance is the correct contract for any host that already uses Yjs.

The default `@tumaet/apollon` (self-contained) and `@tumaet/apollon/react` entries are behaviourally unchanged.

| Subpath | Dependencies | Use when |
| --- | --- | --- |
| `@tumaet/apollon` _(default)_ | all bundled (except Yjs) | No bundler — vanilla JS, `<script>`, CDN. |
| `@tumaet/apollon/react` | React family external | A React host that wants the `<Apollon>` component and shares its own React/MUI. |
| `@tumaet/apollon/external` _(new)_ | everything external | A **bundler** host of any framework that wants one shared, fully auditable copy of every dependency. |

### Release note

Two packaging improvements for embedding hosts. First, `yjs` and `y-protocols` are now required peer dependencies instead of being bundled, so your app and Apollon share a single Yjs instance — no duplicate payload and no cross-instance document errors. Second, a new `@tumaet/apollon/external` entry exposes the same imperative `ApollonEditor` API as the default entry but leaves **every** dependency external (React, MUI, emotion, xyflow, `@dnd-kit`, `zustand`, `uuid`, `@chenglou/pretext`, …) — so a bundler host of any framework resolves and de-duplicates each one from its own `node_modules` and gets full supply-chain / SBOM visibility, instead of a copy inlined invisibly into the bundle. The default `@tumaet/apollon` (self-contained) and `@tumaet/apollon/react` entries are unchanged. Action required only if you adopt the new entry or the Yjs peer: install the corresponding peers (most package managers do this automatically).

### Implementation notes

- **`vite.config.ts`** — a third build pass (`LIB_EXTERNAL=true`) emits `dist/external/index.js`, externalizing the React family **and** Apollon's own runtime deps. It reuses the root `dist/index.d.ts` (same public surface as the default entry), so no second declaration file is emitted. `yjs` / `y-protocols` are external in all three passes.
- **`library/package.json`** — adds the `./external` export; moves `yjs` / `y-protocols` from `dependencies` to required `peerDependencies` (the React-family peers remain `optional`), and keeps them in `devDependencies` so the workspace still builds and tests. Adds a `size-limit` budget for the new entry (≈38 KB brotli — everything is external).
- **Source maps were intentionally dropped.** An earlier revision enabled `sourcemap: true`, but that does **not** deliver the SBOM goal: a consumer that disables vendor source maps ignores them, and a dependency inlined-but-undeclared has no version to match against an advisory. It also added ~12 MB to the published tarball. The `/external` entry is the structural fix instead.
- **Compatibility / action required (semver `minor`).** The default entry's old "zero peers" promise becomes "install `yjs` + `y-protocols`." npm 7+ and pnpm 8+ add missing peers automatically; **Yarn (all versions) and pnpm < 8 do not**, and CDN / import-map users must declare them explicitly. A missing peer surfaces as a clear module-not-found error, not silent breakage.
- **Adopting the new entry (host side, separate change):** a bundler host switches its import to `@tumaet/apollon/external` and provides the now-external peers it doesn't already have. Its existing SBOM tooling (`cdxgen`, Syft, …) then attributes every Apollon dependency natively from its own `node_modules` — no Apollon-side SBOM artifact needed.

### Steps for testing

1. `pnpm --filter @tumaet/apollon build` → emits `dist/index.js`, `dist/react/react.js`, and the new **`dist/external/index.js`** (with `dist/index.d.ts` shared by `.` and `/external`). _(verified locally — green)_
2. `pnpm --filter @tumaet/apollon size` → all five budgets pass; the `/external` entry is ≈38 KB brotli. _(verified locally — green)_
3. Confirm `dist/external/index.js` `import`s `react`, `@mui/material`, `yjs`, `@dnd-kit/*`, `zustand`, … as bare specifiers (no inlined copy).
4. `pnpm --filter @tumaet/apollon lint` and `… test` — green in CI (`lint-and-format-check`, `library-node22-compat`, `visual-regression-tests`).
5. In a real bundler host: `import { ApollonEditor } from "@tumaet/apollon/external"`, then confirm exactly one resolved copy of each dependency in the lockfile.

### Screenshots / screencasts

n/a — packaging / build + docs change, no UI.

### Checklist

- [x] Linked to a related issue (if applicable) — n/a
- [x] PR title's Conventional Commit type matches the kind of change (`feat` — a new public entry)
- [ ] Tests added or updated — n/a (packaging change; covered by existing build / size / lint / test gates)
- [x] Documentation updated — README + docs (overview "Three builds, one API", install "Fully external build", quickstart, Angular)
- [x] Screenshots or screencasts attached (if a UI change) — n/a


library/package.json
```diff
@@ -104,7 +104,9 @@
     "@mui/material": "^6.4.0",
     "@xyflow/react": "^12.9.0",
     "react": "^19.0.0",
-    "react-dom": "^19.0.0"
+    "react-dom": "^19.0.0",
+    "y-protocols": "^1.0.6",
```
Claudia-Anthropica, line 117: @krusche [medium] Adding these as required peers changes the standalone install contract, but the README and docs still say the default build has no peer deps; the vanilla/CDN example also does not show how `yjs` and `y-protocols` resolve. Users following those pages can install or import only `@tumaet/apollon` and then hit unresolved peer imports.

In `library/package.json`, `yjs` and `y-protocols` are now required peers but the install documentation still says the standalone build has zero peer dependencies. Update `library/README.md` and the affected docs under `docs/library/` to include `yjs` and `y-protocols` in install guidance, including CDN/import-map guidance for no-bundler examples.

</details>

Claudia-Anthropica (changes requested):
@krusche The build change itself checks out locally: library build and size passed, and the generated artifacts keep Yjs imports external. I am requesting changes because the public install docs still promise a zero-peer standalone setup, which is no longer true with these required Yjs peers.

library/package.json
Claudia-Anthropica: @krusche Thanks for updating the npm install/docs pages. The CDN/vanilla path is still the missing piece: `docs/library/embedding/vanilla.md` and the matching README/homepage snippets still import only `https://esm.sh/@tumaet/apollon@4.8.0` and never document how the required `yjs`/`y-protocols` peers resolve for no-bundler users. Please either add the esm.sh peer-resolution/import-map guidance there or update the snippets so the peers are explicit.

Claudia-Anthropica (changes requested):
@krusche The npm install docs are mostly updated now, but the CDN/vanilla path is still missing the peer-resolution guidance from the original thread. Leaving this as changes requested until those no-bundler snippets explain or make explicit how `yjs` and `y-protocols` resolve.

FelixTJDietrich:
Refreshed the title and description to match where this landed.

**TL;DR** — two packaging changes for hosts that embed Apollon with a bundler: `yjs` / `y-protocols` become required peers (one shared Yjs instance instead of two), and a new `@tumaet/apollon/external` entry leaves *every* dependency external so a host of any framework resolves, de-duplicates, and fully audits each one from its own `node_modules`. The default (`.`) and `/react` entries are unchanged.

**Source maps were dropped on purpose** — they don't deliver the SBOM goal (a consumer with `sourceMap.vendor: false`, like Artemis, ignores vendor maps; a dep that's inlined-but-undeclared has no version to match against an advisory) and added ~12 MB to the tarball. The `/external` entry is the structural fix.

**Action required for consumers:** the default entry is no longer zero-peer — install `yjs` + `y-protocols`. npm 7+ and pnpm 8+ add them automatically; Yarn and pnpm < 8 don't. A missing peer is a clear module-not-found, not silent breakage. Shipped as a `minor`.

**CI status (transparent):** everything is green except `e2e-tests` (and the `PR Health Gate` that depends on it). The single failing test — `activity-swimlane.spec.ts › lanes can be reordered by dragging in the popover` — is a `dnd-kit` drag interaction that is **failing on most current PRs (#774, #777, #778, #780, #781), not just this one**, and this PR touches neither the webapp nor that test (`git diff origin/main -- standalone/webapp/` is empty). So it's a pre-existing flaky test on `main`, unrelated to this change; a rerun doesn't fix it. It's worth quarantining/stabilising separately so it stops blocking unrelated PRs.

**For Artemis (separate, host-side):** import `@tumaet/apollon/external` and add the peers it doesn't already have; its existing `cdxgen` SBOM pipeline then sees every Apollon dependency natively — no Apollon-side SBOM artifact required.

FelixTJDietrich:
Pushed `f29a2ebf` — finishes the install docs for the new Yjs peer:

- root `README.md` install command now includes `yjs y-protocols`;
- the CDN/no-bundler examples (docs *Vanilla* page **and** the published `library/README.md`) now use an import map + `?external=yjs,y-protocols` so the page shares a single Yjs instance instead of silently loading a second copy.

Docs-only, so no changeset. The three other entries' install instructions and the peer tables were already correct.

library/package.json
Claudia-Anthropica: @krusche The dedicated vanilla guide and library README now cover the esm.sh peer resolution, thanks. The homepage Vanilla JS tab in `docs/src/pages/index.tsx` still shows the CDN snippet without any note that esm.sh resolves the required `yjs`/`y-protocols` peers, so that last part of the thread is still open. Please add the same one-line guidance near that snippet or make the peers explicit there.

Claudia-Anthropica (changes requested):
@krusche Most of the docs are fixed now, including the dedicated vanilla guide and library README. I am keeping changes requested because the homepage Vanilla JS CDN snippet still lacks the peer-resolution note from the open thread.

library/package.json
Claudia-Anthropica: @krusche Still open: the homepage install command now includes the npm peers, but the Vanilla JS tab itself still only shows the esm.sh import and does not say that esm.sh resolves the required `yjs`/`y-protocols` peers. Please add that one-line note next to the Vanilla JS snippet, or make those peer imports explicit in the snippet.

Claudia-Anthropica (changes requested):
@krusche The homepage npm install line is updated, but the open CDN/Vanilla JS point is still not addressed in the tab that shows the snippet. Keeping this as changes requested until that snippet explains the esm.sh peer resolution or imports the peers explicitly.

library/package.json
FelixTJDietrich: Addressed in 7ccfc602 — the homepage Vanilla JS tab now has a one-line comment next to the import noting that esm.sh serves the required `yjs`/`y-protocols` peers, matching the vanilla embedding guide and the library README.

One clarification on the premise, though: the bare esm.sh import was never actually broken for no-bundler users. esm.sh rewrites a package's dependency imports (peers included) to esm.sh-hosted URLs automatically, so `import … from "https://esm.sh/@tumaet/apollon"` pulls working `yjs`/`y-protocols` with no extra step. I verified this in a headless-browser load of the built bundle: the editor mounts and renders with the peers resolved and a single Yjs instance — no unresolved-import error. So this is a clarity improvement rather than a correctness fix. (The npm/bundler paths, where you do install the peers, are documented separately on the install page.)

Resolving as the requested note is now in place.

Claudia-Anthropica: @FelixTJDietrich Good point, that makes sense on esm.sh peer rewriting. The added note covers the docs clarity issue I was asking for. Looks good now, thanks.

Claudia-Anthropica (approved):
@krusche All requested docs feedback is addressed now. The thread is resolved and the latest clarification looks good to me.

FelixTJDietrich (approved):
Changes look good and work as expected, I ensured that the examples in the docs are working. I will do one more minor adjustment to the docs before I merge

library/vite.config.ts
```diff
@@ -25,35 +25,74 @@ function emitFontLicense(): Plugin {
...
+  "@emotion/react",
+  "@emotion/styled",
+  /^@mui\/material(\/.*)?$/,
+  "@xyflow/react",
+]
+
+// Apollon's own non-framework runtime deps. Inlined everywhere except the fully
+// external build, where the host resolves them — one shared copy and full SBOM
+// visibility instead of a copy buried inside the bundle.
+const RUNTIME_DEPS = [/^@dnd-kit\//, "zustand", "uuid", "@chenglou/pretext"]
```
Claudia-Anthropica: @krusche [medium] `zustand` is only listed as an exact external id, but the source imports `zustand/middleware` and `zustand/shallow`; Rollup treats those as different ids, so the `/external` build still inlines that code. I reproduced it with `LIB_EXTERNAL=true pnpm --filter @tumaet/apollon exec vite build`: `dist/external/index.js` imports `zustand` but also contains the `[zustand devtools middleware]` implementation. Match the subpaths too so the fully external build actually keeps all of Zustand external.

```suggestion
const RUNTIME_DEPS = [/^@dnd-kit\//, /^zustand(\/.*)?$/, "uuid", "@chenglou/pretext"]
```

In `library/vite.config.ts`, `RUNTIME_DEPS` externalizes only the exact `zustand` id, so `zustand/middleware` and `zustand/shallow` are bundled into `dist/external/index.js`. Change the Zustand entry to a regex that matches `zustand` and all subpaths, then rebuild the external entry and verify those subpaths stay external.

</details>

Claudia-Anthropica (changes requested):
@krusche The docs follow-up is resolved, but the new `/external` build still bundles Zustand subpath helpers because the external list only matches the package root. I left the concrete fix inline.

library/vite.config.ts
FelixTJDietrich: Good catch — confirmed and fixed in 235b67eb. The source imports `zustand/middleware` (7×) and `zustand/shallow` (120×), and since Rollup treats subpaths as distinct ids the exact `"zustand"` left them inlined in `dist/external/index.js`. Switched to `/^zustand(\/.*)?$/` (mirroring the `@dnd-kit` entry).

Verified after the change: `dist/external/index.js` now imports `"zustand"`, `"zustand/middleware"`, and `"zustand/shallow"` as bare externals, and the size budget still passes. I also checked the other exact-string externals while here — `uuid`, `@chenglou/pretext`, `@emotion/*`, and `@xyflow/react` import no subpaths (the one `@xyflow/react/...css?inline` is a deliberate CSS inline), so they stay correct as exact ids.

Claudia-Anthropica: @FelixTJDietrich Nice fix, thanks. I verified the library build and the generated `/external` bundle now keeps `zustand`, `zustand/middleware`, and `zustand/shallow` as bare imports; size-limit still passes too.

Claudia-Anthropica (approved):
@krusche All requested feedback is addressed now. I verified the library build, the generated external bundle imports the Zustand subpaths as bare externals, and the size budget still passes.

FelixTJDietrich (approved):

PR #783, opened 2026-06-26
feat(library): snap activity-diagram lane resize to the grid

FelixTJDietrich:
### Summary

Dragging an activity-diagram swimlane's lane separator was pixel-continuous, so lane boundaries could land on arbitrary sub-pixel positions — out of step with the rest of the canvas, where nodes (`snapGrid`) and edge bends already snap to a 5px grid. This makes the lane separator snap to that same grid while you drag it.

The snap lives in the pure `resizeLaneDivider` util (new optional `snapGridPx` arg); the component passes `CANVAS.SNAP_TO_GRID_PX`. Lane offsets are measured from the swimlane's own origin, which React Flow already keeps grid-aligned, so the snapped boundary lines up with the absolute grid. The minimum-lane clamp still wins at the extremes.

This branch also fixes a genuinely flaky lane-reorder e2e test (separate commit) — see below.

### Release note

Activity-diagram lane separators now snap to the canvas grid while you drag them, so lane boundaries line up cleanly.

### Implementation notes

- Snapping rounds the *moved boundary* to the grid, not the raw delta, so the separator lands on a grid line regardless of where the drag started.
- Default `snapGridPx = 0` keeps `resizeLaneDivider` a no-op snap unless a grid is passed — existing callers/tests are unaffected.
- Clamp order is snap → min-extent, so a snap that would shrink a lane below `MIN_LANE_EXTENT` is still clamped.

**Flaky e2e fix (root cause, not a bandaid):** the lane-reorder test drove dnd-kit through a synthetic `page.mouse` drag, whose success depends on collision detection sampling the pointer mid-move — a race that drops the reorder under loaded CI. No number of extra steps/waits makes a pointer drag deterministic. The popover already registers a `KeyboardSensor`, so the test now uses it: pick the handle up (`Space`), wait for dnd-kit to mark it active (`aria-pressed`), move up one slot (`ArrowUp`), drop (`Space`). Verified locally 5/5 on a clean server and the full spec 7/7.

### Steps for testing

1. Open an Activity diagram, drop a swimlane, and drag a lane separator.
2. The separator advances in 5px steps and the lane edges align with where nodes snap.
3. Drag a separator hard against a neighbour — the neighbour floors at the minimum lane size instead of snapping past it.

### Checklist

- [x] PR title's Conventional Commit type (`feat`/`fix`/…) matches the kind of change
- [x] Tests added or updated

### DIFF
--- a/.changeset/swimlane-resize-grid-snap.md
+++ b/.changeset/swimlane-resize-grid-snap.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Activity-diagram lane separators now snap to the canvas grid while you drag them, matching how nodes and edge bends already snap. Lane boundaries line up cleanly instead of landing on arbitrary sub-pixel positions.
--- a/library/lib/nodes/activityDiagram/ActivitySwimlane.tsx
+++ b/library/lib/nodes/activityDiagram/ActivitySwimlane.tsx
@@ -7,6 +7,7 @@ import { ActivitySwimlaneProps } from "@/types"
 import { useDiagramStore } from "@/store"
 import { useShallow } from "zustand/shallow"
 import { getLaneOffsets, resizeLaneDivider } from "@/utils"
+import { CANVAS } from "@/constants"
 import { PopoverManager } from "@/components/popovers/PopoverManager"
 import { useDiagramModifiable } from "@/hooks/useDiagramModifiable"
 import {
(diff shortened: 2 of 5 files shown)


Claudia-Anthropica (approved):
@FelixTJDietrich The implementation is clean: snapping happens in the pure lane-resize helper, the component passes the canvas grid, and the existing no-snap behavior stays intact. Focused swimlane unit tests pass locally; the completed CI checks are green and the rest were still pending when I reviewed.

PR #786, opened 2026-06-30
chore(deps): modernize toolchain (Vite 8, TS 6, ESLint 10, Hono) + cut 14 direct deps (-139 pkgs)

FelixTJDietrich:
## Summary

A ruthless dependency-modernization pass over the whole monorepo: push the build toolchain to the latest working majors, migrate the server's runtime majors (including **Express 5 → Hono 4**), and cut dead/redundant packages — leaving the repo **measurably leaner, more current, and behavior-identical**.

**Net `−139` resolved packages** (2667 → 2528), **14 direct dependencies removed** vs 4 added, every runtime/published dependency **at npm latest**, and **zero new vulnerabilities** introduced (the one security-relevant bump, `ws`, actually *fixes* an advisory in the server runtime).

> Stacked on top of #785 (single-entry `@tumaet/apollon`), now merged to `main`; this PR rebases cleanly onto it with no conflicts.

---

## Headline outcomes

| | |
|---|---|
| **Toolchain** | Vite 6→**8** (Rolldown/Oxc) · `@vitejs/plugin-react` 4→**6** · `vite-plugin-dts` 4→**5** (+`@microsoft/api-extractor`) · TypeScript 5.7→**6.0** · ESLint 9→**10** (→ `@eslint-react`) |
| **Server runtime** | zod 3→**4** · redis 5→**6** (RESP2-pinned) · pino 9→**10** · dotenv 16→**17** · ulid 2→**3** · **Express 5 → Hono 4** |
| **Footprint** | **−139 packages** net · **14 direct deps cut** · webapp initial bundle ≈ **−47%** (the `@ionic/react` drop) |
| **Security** | **0 introduced**; `ws 8.18→8.21` **remediates** GHSA ws-DoS in the WS relay; all residual crit/high are inherited dev/build/test transitives |
| **Currency** | **Every runtime + published dependency at npm latest**; the only lag is dev tooling and items deliberately held by the repo's `minimumReleaseAge` |

---

## 1. Build toolchain & shared dev tooling

| Package | Before | After | Rationale |
|---|---|---|---|
| `vite` | 6.4.2 | **8.1.0** | Rolldown/Oxc bundler+minifier. Oxc drops `debugger` on minify, so the explicit `esbuild: { drop }` is gone. |
| `@vitejs/plugin-react` | 4.3.4 | **6.0.3** | Requires Vite 8's `vite/internal`; paired with the Vite bump. |
[...]

PR #787, opened 2026-07-01
fix(library): keep -webkit-backdrop-filter first so glass blur survives minification

FelixTJDietrich:
## What & why

After the toolchain modernization (#786, Tailwind `4.0.17 → 4.3.1` + Vite 6 → 8), the frosted-glass **chrome effect flickers and the blur is missing in production** — but only in Chrome/Firefox, not Safari.

### Root cause

Every hand-written glass surface declared the standard property *before* its vendor-prefixed twin, with the same value:

```css
backdrop-filter: var(--apollon-chrome-glass-blur);
-webkit-backdrop-filter: var(--apollon-chrome-glass-blur);
```

**lightningcss** (Tailwind v4's CSS optimizer, run on the production build via `@tailwindcss/vite`) collapses an identical-value vendor-prefix pair down to whichever declaration comes **last**. So the shipped CSS kept only `-webkit-backdrop-filter` — Safari blurs, but Chrome/Firefox render the bare translucent `--apollon-chrome-glass` tint. That also lets sharp page content + the dot-grid canvas bleed through the semi-transparent bars, which reads as the reported flicker on scroll/theme-repaint.

Verified against the **live production CSS** (`apollon.aet.cit.tum.de`): the standard `backdrop-filter:var(--apollon-chrome-glass-blur)` appears **0 times**, the `-webkit-` variant **4 times**. Reproduced locally by running the exact declaration pair through the lightningcss in the tree (1.29.2 & 1.32.0) with real browser targets — both drop the standard property.

### Fix

Reorder `-webkit-backdrop-filter` **first** across `packages/ui/src/styles/components.css` and `library/lib/styles/app.css`. lightningcss keeps both properties when the prefixed one is first (verified: re-running the rebuilt `dist` through lightningcss now yields 2 standard + 2 webkit decls). No visual/behavior change on Safari; restores the blur on Chrome/Firefox.

A guard test (`packages/ui/src/styles/backdropFilterOrder.test.ts`) fails on the broken order so the next dependency bump can't silently reintroduce it.

## Testing
- `pnpm --filter @tumaet/ui test` → 39 passed (incl. new guard)
- `pnpm --filter @tumaet/apollon exec vitest run tests/unit/cssVariableContract.test.ts` → 79 passed
- Confirmed the guard regex catches the broken order and passes the fixed order.
- Confirmed rebuilt `packages/ui/dist/components.css` survives production lightningcss minification with both properties intact.

## Notes
- `packages/ui/dist` is gitignored and rebuilt from source during the library/webapp build, so only source changes are committed.
- Changeset: `@tumaet/apollon` + `@tumaet/ui` patch (cascades to `@tumaet/webapp`/`@tumaet/server`).

## Out of scope (flagged, not fixed here)
Rolldown's CSS code-splitting ships the full token set + Tailwind base **twice** (`lib.css` ~86 KB and `index.css` ~135 KB both carry it). Separate perf cleanup, not causing this bug.

### DIFF
--- a/packages/ui/tests/backdropFilterOrder.test.ts
+++ b/packages/ui/tests/backdropFilterOrder.test.ts
@@ -1,0 +1,37 @@
+import { describe, it, expect } from "vitest"
+import { readFileSync } from "node:fs"
+import { resolve } from "node:path"
+
+// The CSS minifier collapses an identical-value vendor-prefix pair to the last
+// declaration, so `backdrop-filter` before `-webkit-backdrop-filter` ships as
+// webkit-only and the glass blur dies in Chrome/Firefox. This guards the order
+// in the hand-written CSS so it can't regress on a dependency bump.
+
+const REPO_ROOT = resolve(__dirname, "../../..")
+
+const CSS_FILES = [
+  "packages/ui/src/styles/components.css",
+  "library/lib/styles/app.css",
+]
+
+// Matches the broken order: an unprefixed decl immediately followed by the
+// same-value -webkit- decl. `-webkit-` first, or a lone decl, is safe.
+const BROKEN_ORDER =
+  /(?<![-\w])backdrop-filter:\s*([^;]+);\s*-webkit-backdrop-filter:\s*\1\s*;/g
+
+describe("backdrop-filter vendor-prefix ordering", () => {
+  for (const file of CSS_FILES) {
+    it(`${file} lists -webkit-backdrop-filter before the standard property`, () => {
+      const css = readFileSync(resolve(REPO_ROOT, file), "utf-8")
+      const offenders = [...css.matchAll(BROKEN_ORDER)].map((m) =>
+        m[0].replace(/\s+/g, " ").trim()
+      )
+      expect(
+        offenders,
+        `Put -webkit-backdrop-filter BEFORE backdrop-filter in ${file}, or ` +
+          `lightningcss drops the standard property and the blur breaks in ` +
+          `Chrome/Firefox.`
+      ).toEqual([])
+    })
+  }
+})
--- a/.changeset/glass-blur-library.md
+++ b/.changeset/glass-blur-library.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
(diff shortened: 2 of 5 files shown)


packages/ui/tests/backdropFilterOrder.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, it, expect } from "vitest"
+import { readFileSync } from "node:fs"
```
Claudia-Anthropica, line 2: @FelixTJDietrich [high] This test now lives under `src`, and `packages/ui/tsconfig.build.json` includes `src/**/*` for declaration emit. CI is failing in the `@tumaet/ui` build because this file pulls in Node-only globals (`node:fs`, `node:path`, `__dirname`) while that build config only has Vite/browser types. Move the guard out to `packages/ui/tests/` (Vitest already includes that folder) or exclude test files from the build config before merging.

In `packages/ui/src/styles/backdropFilterOrder.test.ts`, the new test is compiled by `packages/ui/tsconfig.build.json` because it lives under `src`, which breaks declaration emit on Node imports and `__dirname`. Move the test to `packages/ui/tests/backdropFilterOrder.test.ts` so Vitest still runs it but the UI declaration build no longer compiles it, then run `pnpm --filter @tumaet/ui run build:types`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The CSS order fix itself looks sound, but the new guard test currently breaks the UI/library build because it sits under `src` and gets picked up by declaration emit. Requesting changes for that CI failure; details inline.

packages/ui/tests/backdropFilterOrder.test.ts
FelixTJDietrich: Done — moved the guard to `packages/ui/tests/backdropFilterOrder.test.ts` (outside `tsconfig.build.json`'s `src/**/*` include) and fixed the repo-root path. `pnpm --filter @tumaet/ui run build:types` and the full `pnpm run build` both pass now, and Codacy is green.

PR #789, opened 2026-07-02
fix(library): stop the canvas jumping on the first palette drop

FelixTJDietrich:
### Summary

Dropping the **first** element from the palette onto an empty canvas made the whole view lurch: the element didn't land where you released it, and the canvas visibly jumped. Every element after the first behaved fine.

The culprit is how React Flow's imperative `fitView()` behaves in v12. It doesn't fit immediately — it sets an internal `fitViewQueued` flag and waits for nodes to exist before resolving (see `@xyflow/react`'s `setNodes` path). We call `fitView` twice at startup — once in `App.tsx`'s `onInit` and once via the editor's mount-time fit — but at that point the canvas is empty, so nothing resolves the queued fit. It sits there pending until the first `setNodes`… which is the node you just dropped. React Flow then dutifully re-frames that lone node, and the viewport jumps.

The fix makes `fitView` a no-op when there are no nodes to frame, at both call sites. An empty canvas keeps its natural default viewport `(0,0)` / zoom 1, and no stale fit is left queued to ambush the first drop.

### Release note

Dragging the first element from the palette onto an empty canvas no longer jumps the view — the element now lands where you drop it.

### Implementation notes

- `library/lib/App.tsx` — only call `instance.fitView()` in `onInit` when `instance.getNodes().length > 0`.
- `library/lib/apollon-editor.tsx` — the public `fitView()` retries until nodes are measured; bail out early when the store has zero nodes.

Both guards are conservative: fitting an empty canvas never produced a meaningful result anyway (it only ever queued), so nothing about the non-empty path changes. Only the first node was ever affected, because React Flow clears the queued flag once it resolves — which is why a narrower one-site fix would have left the other latent call able to reintroduce the jump.

### Steps for testing

1. Open the editor with an empty diagram (e.g. create a new local Class Diagram).
2. Drag any element from the palette and drop it somewhere off-centre.
3. The element lands under the cursor and the canvas does **not** pan/jump.
4. Drop a few more elements — behaviour is unchanged, and existing diagrams still fit-to-view on load as before.

Automated coverage: `standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts` records the viewport transform before and after the first drop and asserts it is unchanged. It fails on `main` (viewport shifts) and passes with this change.

### Screenshots / screencasts

n/a — no visual/UI surface changes; the fix removes an unwanted viewport movement. Behaviour is covered by the E2E test above.

### Checklist

- [x] Tests added or updated
- [x] Documentation updated (if applicable) — n/a
- [x] Screenshots or screencasts attached (if a UI change) — n/a, no UI surface change

### DIFF
--- a/standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts
+++ b/standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts
@@ -1,5 +1,5 @@
 import { test, expect, type Page } from "@playwright/test"
-import { waitForCanvasReady } from "../helpers/canvas"
+import { openFixtureInLocalEditor, waitForCanvasReady } from "../helpers/canvas"
 
 /**
  * Dropping the first palette element onto an empty canvas must not move the
@@ -7,41 +7,15 @@ import { waitForCanvasReady } from "../helpers/canvas"
  * the first node, re-framing it and jerking the canvas.
  */
 
-async function openTemporaryLocalDiagram(
-  page: Page,
-  diagramType = "ClassDiagram"
-) {
-  const modelId = "e2e-first-drop-model-id"
-  await page.goto("/")
-  await page.evaluate(
-    ({ id, type }) => {
-      const storeValue = JSON.stringify({
-        state: {
-          models: {
-            [id]: {
-              id,
-              model: {
-                id,
-                type,
-                assessments: {},
-                edges: [],
-                nodes: [],
-                title: "E2E Diagram",
-                version: "4.0.0",
-              },
-              lastModifiedAt: new Date().toISOString(),
-            },
-          },
-          currentModelId: id,
-        },
-        version: 0,
-      })
-      localStorage.setItem("persistenceModelStore", storeValue)
-    },
-    { id: modelId, type: diagramType }
-  )
-  await page.goto(`/local/${modelId}`)
-  await expect(page).toHaveURL(new RegExp(`/local/${modelId}$`))
+// An empty Class Diagram — no nodes, so the very next drop is the first node.
+const EMPTY_MODEL = {
+  id: "e2e-first-drop-model-id",
+  type: "ClassDiagram",
+  assessments: {},
+  edges: [],
+  nodes: [],
+  title: "E2E Diagram",
+  version: "4.0.0",
 }
 
 // Read the .react-flow__viewport transform matrix -> {zoom, x, y}.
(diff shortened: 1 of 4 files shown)


standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+ * viewport. A fitView issued while empty stays queued and otherwise fires on
+ * the first node, re-framing it and jerking the canvas.
+ */
+
+async function openTemporaryLocalDiagram(
+  page: Page,
+  diagramType = "ClassDiagram"
+) {
+  const modelId = "e2e-first-drop-model-id"
+  await page.goto("/")
```
Claudia-Anthropica: @FelixTJDietrich [medium] This seeds the fixture after `/` has already mounted and the persistence store has hydrated. In a clean local run the follow-up `/local/${modelId}` page renders "Diagram not found" and `waitForCanvasReady` times out, so the new regression never exercises the drag. Seed via the shared `openFixtureInLocalEditor` helper, or use `page.addInitScript` before the first navigation, then go directly to `/local/${modelId}`.

In `standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts`, `openTemporaryLocalDiagram` writes `persistenceModelStore` after navigating to `/`, so the already-mounted app can miss or overwrite the fixture and `/local/<id>` renders the not-found page. Fix this by seeding the fixture before navigation with the existing `openFixtureInLocalEditor` helper from `tests/helpers/canvas` or with `page.addInitScript`, then navigate directly to `/local/<id>`. Keep the model empty so the first-drop regression remains covered.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The code change looks scoped, but the new Playwright regression currently does not reach the canvas in a clean run because the fixture is seeded after the app has already hydrated. I reproduced this locally with `pnpm --dir standalone/webapp exec playwright test tests/e2e/canvas-first-drop-jump.spec.ts`, so I am requesting changes for the test setup.

standalone/webapp/tests/e2e/canvas-first-drop-jump.spec.ts
FelixTJDietrich: Good catch, thank you — you're right that the old setup raced hydration. In fact it was worse than a flake: with correct seeding the drag is genuinely exercised, which exposed that the earlier version was passing *vacuously*.

Fixed in a42b7f70: the empty model is now seeded with the shared `openFixtureInLocalEditor` helper (`addInitScript` before the first navigation), then we go straight to `/local/<id>`. I re-verified it's a real regression guard — it fails (~125px viewport jump) on the unguarded editor and passes with the fix.

PR #797, opened 2026-07-02
chore(library): silence api-extractor's non-actionable TypeScript-version notice

FelixTJDietrich:
### Summary

Every library build prints this, three times (once per `.d.ts` entry point):

```
Analysis will use the bundled TypeScript version 5.9.3
*** The target project appears to use TypeScript 6.0.3 which is newer than the
    bundled compiler engine; consider upgrading API Extractor.
```

It reads like an error but isn't. `vite-plugin-dts` rolls the library's `.d.ts` up with **@microsoft/api-extractor**, whose newest release (7.58.9) bundles TypeScript 5.9.3 and has **no release that supports TypeScript 6 yet**. So it analyses with its bundled compiler and warns. "Consider upgrading API Extractor" is non-actionable — there's nothing newer to upgrade to.

Crucially, the rolled `.d.ts` is **correct**: I diffed the output before/after and it's byte-identical, it type-checks under TS 6.0.3, and `@tumaet/ui` is still fully inlined (zero external leakage). So this is pure build-log noise.

### Fix

Suppress **only** that one notice via API Extractor's supported `messageCallback` invoke option (exposed by `vite-plugin-dts`'s `bundleTypes.invokeOptions`), matching the message id `console-compiler-version-notice` and marking it `handled`:

```ts
const dtsInvokeOptions: IExtractorInvokeOptions = {
  messageCallback(message) {
    if (message.messageId === "console-compiler-version-notice") {
      message.handled = true
    }
  },
}
```

I deliberately **kept** the neutral `Analysis will use the bundled TypeScript version 5.9.3` preamble — it's an honest breadcrumb that the public types are analysed by 5.9.3, so if a TS-6-only construct ever trips it up, the signal is still there. A `// Remove once api-extractor supports the project's TS major` comment marks the workaround for removal.

### Release note

n/a — no Changesets-tracked package's shipped output changed (the emitted `.d.ts` is byte-identical; this only quiets a build-log message).

### Implementation notes

Approaches ruled out on the way:
- **Upgrade api-extractor** — 7.58.9 is the newest; there is no TS-6-capable release. (The `8.0.0` on npm is an ancient mis-publish pinning TS ~3.1.6, not a real upgrade.)
- **`typescriptCompilerFolder`** — tested; no effect. API Extractor detects the *project's* TS version independently of which compiler it's told to use, so pointing it at 5.9.3 doesn't stop it noticing the project is on 6.0.3.
- **`messages` config** — only governs compiler/extractor/tsdoc messages, not `ConsoleMessageId` messages; `messageCallback` is the sanctioned lever for these.

### Steps for testing

1. `pnpm --filter @tumaet/apollon run build`
2. Build output no longer contains `consider upgrading API Extractor`; the neutral `bundled TypeScript version` preamble remains.
3. `library/dist/*.d.ts` is unchanged (byte-identical to `main`).

### Screenshots / screencasts

n/a — build-log change only.

### Checklist

- [x] No shipped output changed (emitted `.d.ts` byte-identical) — no changeset needed
- [x] PR title's Conventional Commit type matches the change (`chore`)
- [x] Verified locally: build clean, `.d.ts` unchanged, `lint:lib` passes, `vite.config.ts` type-checks
- [ ] Tests — n/a (build-config change)
- [x] Documentation — n/a


Claudia-Anthropica (approved):
@FelixTJDietrich This is narrowly scoped and matches the API Extractor invocation path used by the current d.ts bundler. CI is green, so approving as-is.

PR #798, opened 2026-07-02
feat(library): add composable localized overlay controls

FelixTJDietrich:
### Summary

This PR turns Apollon's built-in editor chrome into composable, localized overlay controls instead of fixed UI.

Embedding apps can now:

- use the default palette, zoom/history cluster, and minimap by omitting `<Apollon>` children;
- compose the built-ins explicitly with `<Apollon.Palette />`, `<Apollon.Zoom />`, `<Apollon.MiniMap />`, or `<ApollonDefaultControls />`;
- pass `null` / an empty fragment for a bare canvas, or omit a built-in to hide it;
- configure the same built-ins imperatively with `paletteControl`, `zoomControl`, `miniMapControl`, and `defaultControls`;
- add, move, restyle, replace, update, or remove controls through one overlay registry;
- override editor-owned UI strings through `labels`, `editor.setLabels(...)`, and `useLabels()`.

The built-ins and host chrome now share one measured layout layer: bands reserve room, band lanes stack predictably, corner slots float without reserving whole columns, same-side rail/corner extents are measured, and `fitView` frames diagram content around the resulting insets.

### Release note

Embedding apps can now compose, hide, move, replace, and localize Apollon's built-in palette, minimap, and zoom/history controls while built-in and host chrome share one measured overlay layout.

### Implementation notes

- Built-ins are ordinary registry records under reserved ids: `PALETTE_ID`, `ZOOM_ID`, and `MINIMAP_ID`.
- React hosts can use `<ApollonDefaultControls />` for the standard chrome or list individual built-ins as children; imperative hosts use descriptor factories or `addControl` / `updateControl` / `removeControl`.
- Reserved-id replacements are normal controls after replacement: built-in region validation follows the built-in descriptor, not the public id forever.
- `getControl(id)` returns an immutable options snapshot and intentionally omits the renderer, so readbacks cannot accidentally rewrite built-in internals.
- `labels` is a partial typed dictionary merged over English defaults and can be updated at runtime with `editor.setLabels(...)`; default node-type popover titles, select placeholders, and generated swimlane names are part of that seam.
- The minimap is a managed corner-slot control, so a short right rail or problem-statement rail no longer pushes it away from the canvas edge.
- Narrow layouts bound rail tracks and make oversized host rail content scroll inside its slot instead of widening the overlay grid.
- Bottom-corner controls align to their painted edge, so an expanded minimap no longer stretches a neighboring action island's measured inset.
- Sparse bottom action islands should use a bottom corner region with an inset; `footer` remains the full-width docked action-bar region.
- Selection-following controls default to `role="group"`; hosts should opt into `role="toolbar"` only when their children implement toolbar keyboard behavior.
- Mobile keyboard lifting is scoped to the editor whose canvas intersects the visual viewport obstruction.
- The size-limit check measures the published library files with peer/runtime dependencies externalized, matching the package's Vite library build semantics.

### Testing

Verified locally:

- `pnpm --filter @tumaet/apollon exec vitest run tests/unit/Apollon.test.tsx tests/unit/overlayApi.test.tsx`
- `pnpm --filter @tumaet/apollon run build`
- `pnpm --filter @tumaet/apollon run lint` (0 errors; existing warnings remain)
- `pnpm --filter @tumaet/apollon run size` (`dist/index.js`: 83.01 kB brotli vs 85 kB limit)
- `pnpm format:check`
- `CI=1 pnpm --filter @tumaet/webapp exec playwright test controls-layout controls-a11y --workers=1` (21 passed)

Manual smoke paths:

1. In Storybook `editor-controls-immersive--student-modeling`, `--tutor-assessment`, and `--exam-mode`, verify the minimap stays anchored to the canvas corner and bottom action islands do not push unrelated controls away.
2. Compose `<ApollonDefaultControls />` plus a custom `<ApollonControl>` child; verify the default chrome remains visible.
3. Replace a built-in at a reserved id, then call `editor.updateControl(id, { region })`; verify the replacement moves like a normal control.
4. Switch `labels` at runtime; verify zoom/minimap/palette labels and default popover labels update without remounting.

### Checklist

- [x] Added a changeset for the public `@tumaet/apollon` API change
- [x] PR title uses a valid Conventional Commit type/scope
- [x] Tests added or updated for overlay registry behavior, insets, layout geometry, responsive clipping, and accessibility
- [x] Documentation updated (`docs/library/api/overlay-controls.md`, `docs/library/api.md`, `library/README.md`, `docs/library/overview.md`)
- [x] Bundle size checked with the package's size-limit workflow


library/lib/chrome/BuiltInControls.tsx
```diff
@@ -0,0 +1,176 @@
...
+  return JSON.stringify({
+    region: c.region ?? null,
+    order: c.order ?? null,
+    inset: c.inset ?? null,
+    interactive: c.interactive ?? null,
+    visible: baseVisible && (c.visible ?? true),
+    className: c.className ?? null,
+    style: c.style ?? null,
+    props: c.props ?? null,
+    hasRender: !!c.render,
```
Claudia-Anthropica: @FelixTJDietrich [medium] Replacement renderers are not actually reactive here. If a host calls `editor.setControl("zoom", { render: nextRender })` or passes a new `<Apollon controls>` render callback with the same placement, the signature stays `hasRender: true`, the effect bails, and the overlay registry keeps calling the old render function. Keep the registration throttling, but store/call the latest render through a ref or otherwise force an update when the render callback changes.

In `library/lib/chrome/BuiltInControls.tsx`, replacement `render` callbacks for built-in controls are collapsed to `hasRender: true`, so changing the callback does not update the registered overlay renderer. Fix this by making registered render thunks call the latest render callback from a ref, or by including a safe render-version update path that refreshes the overlay renderer without causing registration loops.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The controls API direction looks solid, but the replacement renderer path currently breaks the reactive contract; see inline. GitHub's server-tests job is also red with empty SVG export assertions, though the suite passed locally for me after a server build, so that check needs a rerun or investigation before merge.

library/lib/chrome/BuiltInControls.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback addressed from my side, nice work.

library/lib/chrome/builtins/controls.tsx
```diff
@@ -0,0 +1,100 @@
...
+ * components (React). Placement/appearance overrides ride the shared
+ * {@link OverlayControlOptions} vocabulary; the reserved id and default `render`
+ * are fixed here.
+ */
+export const PALETTE_ID = "apollon:palette"
+export const ZOOM_ID = "apollon:zoom"
+export const MINIMAP_ID = "apollon:minimap"
+
+/** Placement/appearance a caller may override on a built-in (never its id). */
+export type BuiltInPlacement = Partial<Omit<OverlayControlOptions, "id">>
```
Claudia-Anthropica: @FelixTJDietrich [medium] `BuiltInPlacement` is too broad for the built-ins that reuse it. `Apollon.MiniMap region="header"` and `paletteControl({ region: "bottom-right" })` type-check today, but the minimap passes that value straight to React Flow `Panel` and the palette measurement bails unless it is inside `.apollon-overlay-band`, so those public placements render incorrectly. Restrict the placement types per built-in, or make the renderers actually handle every advertised region.

In `library/lib/chrome/builtins/controls.tsx`, `BuiltInPlacement` lets palette and minimap controls use overlay regions their renderers do not support. Fix this by defining per-built-in placement types that restrict `region` to supported values, or update the palette/minimap renderers and validation so every accepted region works correctly.

</details>

library/lib/components/Sidebar.tsx
```diff
@@ -177,6 +187,10 @@ export const Sidebar = () => {
       ref={asideRef}
       className="apollon-palette"
       data-testid="apollon-palette"
+      aria-label={labels.elementPalette}
```
Claudia-Anthropica, line 204: @FelixTJDietrich [medium] This localizes the palette's aria label, but the visible palette strings just below still hard-code "Model", "Select Elements", and the highlight hint. With the new `labels` API a host language switch still leaves part of the palette in English, which contradicts the i18n contract in the docs. Add label keys for these strings and use them here.

In `library/lib/components/Sidebar.tsx`, the palette view switch and highlight hint are still hard-coded English while the new labels API claims palette strings are overridable. Add `ApollonLabels` keys/defaults for those palette strings and render them from the active labels.

</details>

docs/library/api/overlay-controls.md
```diff
@@ -239,6 +321,71 @@ control never drags the diagram.
...
+  confirms and toasts in your own React tree (or your host modal system), not as
+  controls. The controls API is for chrome anchored to the canvas edges/corners.
+
+## i18n
+
+Every string the editor renders itself — palette / zoom / minimap tooltips and
+aria-labels, and the editing popovers — is a key in the typed `ApollonLabels`
+dictionary, shipped with English defaults. Override any subset:
+
+```tsx
```
Claudia-Anthropica, line 458: @FelixTJDietrich [medium] This fence is checked by `pnpm run check:doc-snippets`, and CI is failing here because `Apollon` is not in scope and `percent` is inferred as `any`. Either make the example self-contained with the needed import and callback type, or mark this illustrative snippet as `tsx no-check`.

In `docs/library/api/overlay-controls.md`, the i18n example starting at line 366 is a checked TypeScript fence but does not typecheck. Fix it by importing or otherwise declaring `Apollon` and typing `percent`, or change the fence to `tsx no-check` if the snippet is intentionally partial.

</details>

library/lib/styles/app.css
```diff
@@ -72,11 +72,39 @@
...
+  grid-template-columns: auto 1fr auto;
+  grid-template-rows: auto auto minmax(0, 1fr) auto auto;
+  grid-template-areas:
+    "header   header    header"
+    "topleft  topcenter topright"
+    "leftrail center    rightrail"
+    "botleft  botcenter botright"
+    "footer   footer    footer";
+  pointer-events: none;
+  padding-top: var(--safe-area-inset-top, 0px);
```
Claudia-Anthropica: @FelixTJDietrich [low] Codacy is failing on the new CSS style issues here: these zero fallbacks need to be unitless (`0`), and the new palette block comment at line 277 needs the expected blank line before it. Please clean those up so the static analysis check goes green.

In `library/lib/styles/app.css`, Codacy reports `length-zero-no-unit` on the new safe-area fallbacks and `comment-empty-line-before` before the palette comment. Replace the new `0px` fallbacks with `0` where they are zero lengths and add the missing blank line before the palette comment.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The controls API is coming together, but I’m requesting changes for a few public API/i18n issues and the red checks. CI is currently failing on doc snippets, size-limit (145.72 kB vs 85 kB for dist/index.js), visual BPMN snapshot (72,583 px diff), and Codacy style notices, so those need to be green before merge.

library/lib/chrome/builtins/controls.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

library/lib/components/Sidebar.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

docs/library/api/overlay-controls.md
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

library/lib/styles/app.css
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

FelixTJDietrich:
Follow-up audit/fix pass is pushed in `015c8327`.

What changed:
- tightened built-in control option types and runtime validation (palette rail-only, minimap corner-only, built-in invariants pinned);
- fixed overlay measurement for controls mounted before the `ResizeObserver` effect attaches;
- made `addControl` disposers ownership-safe and `getControl` return immutable snapshots;
- completed the i18n gaps called out in review (palette view labels, class type/method labels, minimap aria-label, swimlane resize label);
- fixed doc snippets, Codacy CSS notices, size-limit semantics, and the BPMN visual baseline.

Verification:
- Local: `pnpm lint && pnpm format:check && pnpm build && pnpm test`
- Local: `pnpm --filter @tumaet/apollon run size`
- Local: `pnpm run check:doc-snippets`
- Local: `pnpm --filter @tumaet/webapp exec playwright test controls-layout controls-a11y`
- Local: `pnpm --filter @tumaet/webapp exec playwright test tests/visual/diagrams.visual.spec.ts --grep BPMN`
- CI: all current PR checks are green, including [PR Health Gate](https://github.com/ls1intum/Apollon/actions/runs/28899889131/job/85735852081), [visual-regression-tests](https://github.com/ls1intum/Apollon/actions/runs/28899889131/job/85735495023), [library-node22-compat](https://github.com/ls1intum/Apollon/actions/runs/28899889131/job/85735495534), and [lint-and-format-check](https://github.com/ls1intum/Apollon/actions/runs/28899889131/job/85735495012).

library/lib/components/react/ApollonSelectionToolbar.tsx
```diff
@@ -0,0 +1,111 @@
...
+  // because it can't appear in a node id, so the round-trip can't fracture.
+  const selected = useStore((s) => {
+    const ids: string[] = []
+    for (const node of s.nodeLookup.values())
+      if (node.selected) ids.push(node.id)
+    return ids.join("\n")
+  })
+  const ids = selected ? selected.split("\n") : []
+  return (
+    <NodeToolbar
```
Claudia-Anthropica, line 64: @FelixTJDietrich [medium] This self-positioned path bypasses `ControlSlot`, so the toolbar never gets the `nopan`/`nodrag`/`nowheel` guard that every other interactive overlay control gets. Existing node toolbars add those guards explicitly; without them, dragging or wheel-interacting on host toolbar buttons can still pan/drag the canvas underneath. Add the guard on the `NodeToolbar` wrapper or wrap the portal in the same stop-propagation/nopan treatment.

In `library/lib/components/react/ApollonSelectionToolbar.tsx`, `<Apollon.SelectionToolbar>` renders through `NodeToolbar` as a self-positioned control and misses the interaction guards. Fix this by adding `nodrag nopan nowheel` and pointer/touch/mouse propagation guards to the toolbar wrapper around `RegionMount`, matching the existing node toolbar behavior.

</details>

library/lib/components/react/Apollon.tsx
```diff
@@ -219,7 +232,31 @@ export function Apollon(props: ApollonProps) {
         style={mergedStyle}
         data-theme={dataTheme}
       />
-      {children}
+      {children ?? <DefaultChildren />}
```
Claudia-Anthropica: @FelixTJDietrich [medium] `null` children still falls back to the default palette/zoom/minimap here. That breaks the composition contract for common conditional JSX like `<Apollon>{showChrome ? <Apollon.Zoom /> : null}</Apollon>`: when `showChrome` is false, users get the full default chrome instead of a bare canvas. Default only when `children` is truly omitted, not when an explicit empty composition renders `null`.

In `library/lib/components/react/Apollon.tsx`, `children ?? <DefaultChildren />` treats explicit `null` children as omitted children and restores the default chrome. Fix this by distinguishing omitted children from explicit `null`, for example rendering defaults only when `children === undefined`, and add a unit test for `<Apollon>{null}</Apollon>` staying bare.

</details>

library/lib/styles/app.css
```diff
@@ -248,54 +274,25 @@
...
-     under it can't cover it. Contained by the editor's own stacking context
-     (`.apollon-editor { isolation: isolate }`), so it stays below host-app
-     chrome. The body-portaled drag ghost renders above it regardless. */
-  z-index: 10000;
+  /* One gap in from the band's top/bottom edges (which already sit clear of the
+     header and bottom controls via the reserved insets) and one edge in from the
+     safe-area-adjusted left the band pins to. */
+  margin-top: var(--apollon-chrome-gap);
+  margin-bottom: var(--apollon-chrome-gap);
+  margin-left: var(--apollon-chrome-edge);
```
Claudia-Anthropica, line 289: @FelixTJDietrich [medium] The palette can now be placed in `right-rail`, but this margin is still hard-coded for the left rail. In a right rail it puts the breathing room on the canvas side and leaves the palette flush against the right edge/safe-area boundary, so the supported placement renders visibly wrong. Flip the edge margin when the parent region is `right-rail`.

In `library/lib/styles/app.css`, `.apollon-palette` always uses `margin-left` even though `paletteControl` supports `right-rail`. Fix this by adding a `[data-apollon-region="right-rail"] .apollon-palette` override that moves the edge margin to `margin-right` and clears `margin-left`; add coverage for right-rail palette placement.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The overlay controls API is in good shape overall, but I found three new runtime/API edge cases in the React composition path and supported palette placement. The inline comments cover the selection toolbar interaction guard, `null` children behavior, and right-rail palette spacing.

Claudia-Anthropica (approved):
@FelixTJDietrich All previously requested feedback is addressed and CI is green. Approving the follow-up.

library/lib/components/Sidebar.tsx
```diff
@@ -172,11 +188,22 @@ export const Sidebar = () => {
     )
   }
 
+  const paletteStyle: React.CSSProperties = {
+    ...(canvas.h ? { maxHeight: canvas.h } : null),
```
Claudia-Anthropica, line 193: @FelixTJDietrich [medium] The short-laptop BPMN palette still overflows after this cap: CI’s `Element palette › fits every element in view without scrolling (short laptop)` gets `scrollHeight - clientHeight === 2` on retry. Please leave a little rounding slack in the measured budget/max-height, or make `computePaletteLayout` use a budget that matches the final border-box height so the grid actually fits without scrolling.

In `library/lib/components/Sidebar.tsx`, the palette `maxHeight` uses the measured rail height but the resulting BPMN palette overflows by 2px in CI. Fix the palette height/layout budget so `scrollHeight - clientHeight <= 1` for the 1280x600 BPMN case and keep the existing no-scroll e2e assertion passing.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The overlay API changes look solid overall, but the latest CI is red. The e2e job still has the short-laptop BPMN palette overflowing by 2px, and visual-regression-tests reports a 47,270 px diff for `visual-bpmn.png`, so those need to be fixed or the snapshot regenerated before merge.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The inline follow-up threads are resolved from my side, but I cannot approve this SHA while the current visual-regression-tests check is still red. The failing annotation is still the BPMN snapshot (`visual-bpmn.png`) with 47,270 differing pixels in `standalone/webapp/tests/visual/diagrams.visual.spec.ts`, so the rendering or baseline needs to be brought back in sync before merge.

Claudia-Anthropica (approved):
@FelixTJDietrich All previously requested feedback is addressed and CI is green now. Approving the follow-up.

library/lib/apollon-editor.tsx
```diff
@@ -424,6 +448,29 @@ export class ApollonEditor {
...
+   *   read a built-in's current `region` after `updateControl`. The renderer is
+   *   intentionally omitted; replace a renderer explicitly with `updateControl`.
+   */
+  public getControl(id: string): OverlayControlSnapshot | undefined {
+    const control = this.overlayStore.getState().controls[id]
+    if (!control) return undefined
+    return Object.freeze({
+      id: control.id,
+      region: control.region,
+      inset: control.inset,
```
Claudia-Anthropica: @FelixTJDietrich [medium] `getControl()` still hands out the live `inset` object here, and `style` below has the same problem. The wrapper is frozen, but a host can mutate `editor.getControl(id)!.inset!.left` or `.style!.width` and silently change the registry without `updateControl()` or an inset recompute. Clone or freeze the nested option objects so the snapshot is actually detached from store state.

In `library/lib/apollon-editor.tsx`, `getControl()` returns `inset` and `style` by reference, so callers can mutate live overlay store state through the supposedly immutable snapshot. Fix this by cloning and freezing nested option objects before returning the snapshot, and add a unit test that mutating the returned `inset`/`style` does not affect a later `getControl()` result or overlay insets.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The API hardening mostly looks good, but `getControl()` still leaks mutable nested option objects from the registry. I left one inline comment for that contract issue.

Issue #105, opened 2020-06-04
User wants to make method abstract


julian-christl:
### Is your feature request related to a problem?
The user can't differenciate between implemented and abstract methods.

### Describe the solution you'd like
It's possible to define abstract methods (possibly with a checkbox?)

### Describe alternatives you've considered
There is no real alternative in Apollon

krusche:
I don't think we need this as part of Apollon

PR #799, opened 2026-07-02
feat(library): italic abstract classes/methods and lowercase UML class keywords

FelixTJDietrich:
### Summary

This makes UML class diagrams notation-correct and the class editor simpler to use.

- **Abstract classes show an _italic_ name** (UML 2.5.1 §9.2.4) instead of the invalid `«Abstract»` keyword — UML has no `«abstract»` keyword; abstractness is a property, not a stereotype.
- **Keywords use the exact lowercase UML spelling** — `«interface»` / `«enumeration»` (2.5.1 Table C.1), not the capitalized forms.
- **Methods can be marked abstract** (italic signature) from the class editor. Attributes can't — UML forbids it — so the control isn't offered there.
- **One "Class type" picker** replaces the old Abstract-checkbox + keyword-toggle pair: `Class · Abstract Class · Interface · Enumeration`. It mirrors the four palette tiles and every other node/edge "type" select, and — because the four options are a *lossless projection* of the underlying `{stereotype, isAbstract}` data — it makes the previously reachable invalid states (an italic `«interface»`, an abstract enumeration) impossible to create.
- **Old diagrams migrate automatically** on load, at every entry point.

The italic is a **real shipped Inter italic face**, so it renders the same in the editor and in **PNG, PDF and PowerPoint** exports. The editor's browser-synthesized slant can't be reproduced by the export renderers (resvg, jsPDF), so it was dropped on export — a real face fixes that at the source.

### Release note

Abstract classes and methods now render in italics that survive PNG/PDF/PowerPoint export, interfaces and enumerations use the correct lowercase UML keywords, and you set all of it from a single "Class type" menu in the class editor. _(Full user-facing note lives in the changeset.)_

### Implementation notes

**Real italic, everywhere it's measured or rendered.** `gen:fonts` subsets Inter Italic + Bold Italic (from the pinned Inter 4.0 release) into the three consumers that must agree byte-for-byte: the editor woff2, resvg's ttf `fontBuffers` (PNG), and the jsPDF VFS (PDF). Abstract text is now *measured* with italic metrics too, preserving the measured == rendered == exported invariant documented in `fontStack.ts`, so boxes still size correctly.

**One deliberate, disclosed scope line — keeping the bundle flat.** The italic faces are **not** base64-inlined into the self-contained SVG export (`exportFonts.ts`). Inlining them would roughly double that chunk (**+180 KB gzip**) to serve a single case: a raw `.svg` opened in a renderer that won't synthesize oblique (e.g. Inkscape). Browsers synthesize the slant from the embedded regular face, and our own PNG/PDF paths carry the real ttf, so the webapp bundle-size budget stays green with no ceiling bump. The only trade-off: abstract text is upright when a `.svg` is opened directly in a non-synthesizing external editor.

**The picker is a projection, not a new data model.** Storage stays orthogonal (`{stereotype, isAbstract}`); the popover just reads/writes a 4-way kind. Writing *both* fields on every change is what eliminates a real bug in the old two-control design — a stray `isAbstract` could survive a switch to Enumeration and paint an italic `«enumeration»` on canvas. A migration guard also clears such a stray modifier on load, and the header-height delta is derived from the `LAYOUT` constants instead of a magic `10`.

**Migration is idempotent and covers every hydration boundary** — constructor, `set model`, `importDiagram`, and the off-screen SVG export — so no load path bypasses it.

**Not a breaking change.** The reshaped class types (`ClassStereotype`, `ClassNodeProps`) are not part of the published API — the package entry only re-exports `typings.ts`, and they're emitted as ambient `declare`, not `export declare`. The published JSON schema leaves node `data` open, so old and new models both validate. `minor` is correct. One known edge is out of scope: a peer joining a *live* Yjs collaboration room during a mixed-version rollout hydrates from the CRDT, which bypasses the load-time migration — a deployment-window concern no semver level addresses.

### Steps for testing

1. Add a **Class**, open its editor (double-click, or the pencil in the toolbar). Cycle **Class type** through Class → Abstract Class → Interface → Enumeration. Confirm: Abstract Class → italic name, no keyword line; Interface/Enumeration → a `«interface»` / `«enumeration»` line; height only changes when a keyword line appears or disappears.
2. In the same editor, toggle a **method** to abstract → its signature renders italic. Note attributes offer no such control.
3. Export to **PNG** and **PDF** → abstract class names and method signatures are italic in both files.
4. Open an older diagram (e.g. a bundled template with `"stereotype": "Abstract"` / `"Interface"`) → it loads as an italic abstract class / lowercase `«interface»` with no manual step.

### Screenshots / screencasts

The class editor's classifier control changes from a checkbox + two-button toggle to a single **Class type** dropdown (each option previews its notation — italic name for Abstract Class, a `«keyword»` chip for Interface/Enumeration). The new popover and the exported-diagram rendering are both captured by the refreshed Playwright visual baselines in this PR (`node-edit.png` and the class-diagram snapshots). A maintainer can attach a quick screencast if desired.

### Checklist

- [x] Linked to a related issue (if applicable) — implements #105 (already closed)
- [x] Added a changeset whose summary is written in the user's voice (`@tumaet/apollon`, minor)
- [x] Tests added or updated
- [x] Ran `pnpm lint && pnpm format:check && pnpm build && pnpm test` locally — green (library 1221, ui 42, webapp 263)
- [x] Documentation updated (if applicable) — n/a; behavior is covered by the changeset
- [ ] Screenshots or screencasts attached — captured via the refreshed visual baselines; optional screencast left for a maintainer


standalone/webapp/tests/visual/how-to-use.visual.spec.ts
```diff
@@ -99,10 +99,10 @@ test.describe("How-to-use modal assets", () => {
     // names.
     await page.locator(".react-flow__node-toolbar svg").nth(1).click()
 
-    // The popover renders the stereotype ToggleGroup — wait for its first
-    // toggle button (CSS uppercases the "Abstract" label to "ABSTRACT").
+    // The popover renders the "Class type" Select — wait for its trigger to
+    // confirm the popover has painted.
     await expect(
-      page.locator(".apollon-stereotype-toggle").first()
+      page.locator('.apollon-select-trigger[aria-label="Class type"]')
```
Claudia-Anthropica, line 105: @FelixTJDietrich [medium] This spec now waits for the new Class type select, but the generated `how-to-use-node-edit.png` asset is unchanged, and the class diagram/template/svg-export baselines are also unchanged even though the fixtures still contain legacy `Abstract`/`Interface` values that now normalize to different rendering. CI’s visual projects will compare against stale images; please regenerate the affected Playwright snapshots/assets with this PR’s rendering.

In `standalone/webapp/tests/visual/how-to-use.visual.spec.ts`, the node-edit visual test now captures the new class type control but the corresponding generated PNG and class-diagram visual baselines are not updated. Regenerate and commit the affected Playwright visual assets/snapshots for the class editor, class diagram fixture, class SVG export fixture, and class templates touched by legacy stereotypes.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The implementation pieces I checked look coherent and the focused library tests pass, but this still needs changes: the visual baselines/assets were not regenerated for the changed class UI/rendering, and the title lint check is currently failing. Please update the affected PNG baselines/assets and fix the PR title casing.

standalone/webapp/tests/visual/how-to-use.visual.spec.ts
Claudia-Anthropica: @FelixTJDietrich This is partially fixed now: the class diagram/template and SVG export baselines were regenerated, but `standalone/webapp/assets/images/how-to-use-node-edit.png` is still unchanged and was not included in the follow-up commit. Since this spec is the source of that modal asset, please regenerate and commit that image too.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The title/CI issue is fixed and most visual baselines are refreshed, but one requested asset is still missing: `standalone/webapp/assets/images/how-to-use-node-edit.png` was not regenerated. Leaving changes requested until that modal image is committed.

standalone/webapp/tests/visual/how-to-use.visual.spec.ts
Claudia-Anthropica: @FelixTJDietrich Still not seeing `standalone/webapp/assets/images/how-to-use-node-edit.png` in the latest commit, so this requested asset update is still open.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit fixes the title and refreshes the Playwright snapshots, but `standalone/webapp/assets/images/how-to-use-node-edit.png` is still missing from the diff. Keeping this as changes requested until that generated modal asset is regenerated and committed.

standalone/webapp/tests/visual/how-to-use.visual.spec.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):

FelixTJDietrich:
Two commits landed after the approval — flagging them for the record since one is a genuine behavior change:

- **`39f899a3` fix(library): preload the italic faces before off-screen SVG export.** The static export path explicitly warms the Inter faces before its one-shot measure/render, but only loaded the upright `400`/`700`. Abstract classes/methods are italic, so that text could be measured and rendered in a fallback family until the italic face loaded — drifting box sizes in the exported SVG. Now the italic `400`/`700` faces are preloaded too. (Build + full library suite green.)
- **`84bd0d45` chore(webapp): scope visual-baseline refresh to the changed diagrams.** While chasing the modal-asset regen I briefly switched the baseline job to `--update-snapshots=all`, which rewrote 32 unrelated diagram baselines that differ from CI only sub-tolerance. Reverted those and the script change; the diff is now scoped to the 4 class-diagram baselines + 4 how-to-use modal images that this change actually affects.

CI is green on the new head (the earlier `e2e-tests` miss was an unrelated flake in `activity-swimlane.spec.ts`'s keyboard drag-reorder poll — green on re-run).

PR #800, opened 2026-07-02
feat: shape-aware freeform edge connection points

tamang29:
### Summary
Edge endpoints can be placed and reconnected **anywhere along a node's border**, snapping to the grid, with a dedicated drag handle for moving an endpoint's origin/destination. Attachment is **shape-aware** — the endpoint lands on the node's actual visible outline rather than its bounding box:

- **ellipse** — use-case ovals: on the curve, at the angle you aimed
- **parallelogram** — flowchart input/output: along the slanted edge
- **four-center** — event / gateway / interface nodes: the four N/E/S/W points
- **freeform-rect** — default box shapes: anywhere on the border
- **none** — legends, annotations, swimlanes: correctly not a connection target

### Release note

Place and reconnect edge endpoints anywhere along a node's border, snapping to the grid, with a drag handle for moving an endpoint. Endpoints attach to each node's real shape rather than its bounding box and match the handles a node shows: use-case ovals connect along their curve, the flowchart input/output parallelogram along its slanted outline, and round or diamond nodes — activity start/end, BPMN events and gateways, flowchart decisions, petri places, and component/deployment interfaces — at their four N/E/S/W points. Legends, annotations, and swimlanes are not connection targets, and endpoints can be reconnected onto container nodes (activity, package, pool, subsystem). Drawing a connection shows a dashed preview that lands exactly where the edge will attach, dragging an endpoint across empty canvas follows the cursor instead of snapping back, connection arcs respond as soon as you reach them, and a connection attaches to the nearest node when several sit close together.

### Implementation notes

A single `getConnectionMode(nodeType)` decides the mode; endpoints are stored as a `{side, ratio}` anchor on the bounding rectangle and projected onto the node's real shape at render. The drop resolver, ghost preview, and rendered edge share `pickNearestConnectable` / `distanceToRect` so a preview never promises a target the drop won't honour.

The stack was hardened by a full-scope audit against `main`:
- **Dead code** — removed an always-true feature flag and its dead branches, unused `edgeUtils` exports, and a dead switch fallback (now a `never` exhaustiveness guard).
- **Single source of truth** — the handle-slot map is derived from the `HandleId` enum instead of a hand-copied table.
- **Dedup** — extracted the shared `useFreeformEndpointNode` hook out of both path hooks; shared the nearest-node math.
- **Tests** — fixed a real e2e bug (a test scanned *all* edges instead of the one under test, so a broken connect could pass); converted flaky sleep-then-snapshot assertions to retrying `expect.poll`.

Verified: tsc, lint, format, schema-gen, unit (1238), and the edge e2e suites are green.

Follow-ups (tracked, not blocking): extract the ~120-line endpoint-drag orchestration still shared verbatim by the two path hooks; model connectable sides as a string-literal union to drop the remaining `Position` casts.

### Screenshots / screencasts

https://github.com/user-attachments/assets/9f227238-efe2-42c5-a76d-8557e09298a0

### Checklist

- [x] Tests added or updated

### DIFF
--- a/library/lib/hooks/useConnect.ts
+++ b/library/lib/hooks/useConnect.ts
@@ -9,20 +9,20 @@ import {
 } from "@xyflow/react"
 import { useCallback, useRef } from "react"
 import {
-  findClosestHandle,
   generateUUID,
   getDefaultEdgeType,
-  getFreeformAnchorFromPoint,
   getSideHandleIdForPosition,
+  type FreeformEdgeAnchor,
 } from "@/utils"
-import { DiagramNodeTypeRecord } from "@/nodes"
+import { getEdgeAnchorFromPoint } from "@/utils/connectionModes"
 import { useDiagramStore, useMetadataStore } from "@/store/context"
+import { useFreeformDropTarget } from "./useFreeformDropTarget"
 import { useShallow } from "zustand/shallow"
 
 const withEndpointAnchor = (
   data: Edge["data"],
   endpoint: "source" | "target",
-  anchor: ReturnType<typeof getFreeformAnchorFromPoint> | null
+  anchor: FreeformEdgeAnchor | null
 ): Edge["data"] => {
   const nextData = { ...((data ?? {}) as Record<string, unknown>) }
   const key = endpoint === "source" ? "sourceAnchor" : "targetAnchor"
@@ -124,8 +124,8 @@ const isSameNodeSameHandleArea = (
 export const useConnect = () => {
   const startEdge = useRef<Edge | null>(null)
   const connectionStartParams = useRef<OnConnectStartParams | null>(null)
-  const { screenToFlowPosition, getIntersectingNodes, getInternalNode } =
-    useReactFlow()
+  const { screenToFlowPosition, getIntersectingNodes } = useReactFlow()
+  const resolveDropTarget = useFreeformDropTarget()
   const { setEdges, addEdge, edges } = useDiagramStore(
     useShallow((state) => ({
       setEdges: state.setEdges,
@@ -144,22 +144,14 @@ export const useConnect = () => {
 
   const defaultEdgeType = getDefaultEdgeType(diagramType)
 
-  const isFourHandleNode = useCallback(
-    (nodeType?: string) =>
-      nodeType === DiagramNodeTypeRecord.componentInterface ||
-      nodeType === DiagramNodeTypeRecord.petriNetPlace ||
-      nodeType === DiagramNodeTypeRecord.petriNetTransition ||
-      nodeType === DiagramNodeTypeRecord.sfcTransitionBranch,
-    []
-  )
   const getDropPosition = useCallback(
     (event: MouseEvent | TouchEvent) => {
       const { clientX, clientY } =
         "changedTouches" in event ? event.changedTouches[0] : event
-      return screenToFlowPosition(
-        { x: clientX, y: clientY },
-        { snapToGrid: false }
-      )
+      // Inherit the editor's grid snapping so a new edge snaps its endpoint to
+      // the grid exactly like reconnect does; screenToFlowPosition is called
+      // without overriding snapToGrid.
+      return screenToFlowPosition({ x: clientX, y: clientY })
     },
     [screenToFlowPosition]
   )
@@ -228,48 +220,22 @@ export const useConnect = () => {
       try {
         if (!connectionState.isValid) {
           const dropPosition = getDropPosition(event)
-          const intersectingNodes = getIntersectingNodes({
-            x: dropPosition.x - 5,
-            y: dropPosition.y - 5,
-            width: 10,
-            height: 10,
-          })
-
-          if (intersectingNodes.length === 0) return
-
-          const fromNodeId = connectionState.fromNode?.id
-          const nodeOnTop =
-            intersectingNodes.findLast((node) => node.id !== fromNodeId) ??
-            intersectingNodes[intersectingNodes.length - 1]
-
-          const internalNodeData = getInternalNode(nodeOnTop.id)
-
-          if (
-            !internalNodeData ||
-            nodeOnTop.width == null ||
-            nodeOnTop.height == null
+          const nodeOnTop = resolveDropTarget(
+            dropPosition,
+            connectionState.fromNode?.id
           )
-            return
+          if (!nodeOnTop) return
 
-          const targetRect = {
-            x: internalNodeData.internals.positionAbsolute.x,
-            y: internalNodeData.internals.positionAbsolute.y,
-            width: nodeOnTop.width,
-            height: nodeOnTop.height,
-          }
-          const targetAnchor = getFreeformAnchorFromPoint(
+          const targetRect = nodeOnTop.rect
+          // Shape-aware: the anchor lands on the node's real shape (oval curve,
+          // diamond vertex, interface centre, …), not just its bounding box.
+          const targetAnchor = getEdgeAnchorFromPoint(
+            nodeOnTop.type,
             dropPosition,
             targetRect
           )
-          const targetHandle = targetAnchor
-            ? getSideHandleIdForPosition(targetAnchor.side)
-            : findClosestHandle({
-                point: dropPosition,
-                rect: targetRect,
-                useFourHandles: isFourHandleNode(nodeOnTop.type),
-              })
-
-          if (!targetHandle) return
+          if (!targetAnchor) return // node is not a connection target (mode "none")
+          const targetHandle = getSideHandleIdForPosition(targetAnchor.side)
 
           if (startEdge.current) {
             const updatedEdge = edges.find(
@@ -354,9 +320,7 @@ export const useConnect = () => {
       defaultEdgeType,
       edges,
       getDropPosition,
-      getInternalNode,
-      getIntersectingNodes,
-      isFourHandleNode,
+      resolveDropTarget,
       setEdges,
       stopConnectionGuidance,
     ]
--- a/library/lib/hooks/useConnect.ts
+++ b/library/lib/hooks/useConnect.ts
@@ -223,7 +223,11 @@ export const useConnect = () => {
             const sourceNodeId = connectionState.fromNode!.id
             const sourceHandleId = connectionState.fromHandle?.id
 
-            if (sourceNodeId === nodeOnTop.id) {
+            // Disallow loop from a handle to the same handle on the same node.
+            if (
+              sourceNodeId === nodeOnTop.id &&
+              sourceHandleId === targetHandle
+            ) {
               return
             }
 
--- a/library/lib/hooks/useConnect.ts
+++ b/library/lib/hooks/useConnect.ts
@@ -8,10 +8,38 @@ import {
   OnEdgesDelete,
 } from "@xyflow/react"
 import { useCallback, useRef } from "react"
-import { findClosestHandle, generateUUID, getDefaultEdgeType } from "@/utils"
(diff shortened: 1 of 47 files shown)


library/lib/hooks/useStepPathEdge.ts
```diff
@@ -266,15 +444,27 @@ export const useStepPathEdge = ({
     return result
   }, [basePath])
 
+  const hasStoredManualPoints = Boolean(data?.points && data.points.length > 0)
+  const hasLocalManualPoints = customPoints.length > 0
+  const shouldPreferComputedPath =
+    freeformAnchorsEnabled && computedPoints.length === 2
```
Claudia-Anthropica: @tamang29 [high] This makes fresh step edges drop user-created bends whenever the computed route is straight. After the first bend, `data.points` is stored, but the next render still sets `shouldPreferComputedPath` because `computedPoints.length === 2`; the effect below then clears `points`, which is exactly why the E2E `edge-fresh-bend` tests are failing. Only prefer/clear the computed path when there are no manual points, or when the endpoint-drag commit explicitly decides to clear them.

In `library/lib/hooks/useStepPathEdge.ts`, `shouldPreferComputedPath` ignores stored manual bend points and causes the effect below to clear `data.points` after a first bend. Fix this by preserving stored/local manual points unless the endpoint-drag commit explicitly clears them, and keep the existing fresh-bend E2E behavior passing.

</details>

library/lib/hooks/useConnect.ts
```diff
@@ -175,11 +223,7 @@ export const useConnect = () => {
             const sourceNodeId = connectionState.fromNode!.id
             const sourceHandleId = connectionState.fromHandle?.id
 
-            // Disallow loop from a handle to itself, but allow loops to other handles.
-            if (
-              sourceNodeId === nodeOnTop.id &&
-              sourceHandleId === targetHandle
-            ) {
+            if (sourceNodeId === nodeOnTop.id) {
```
Claudia-Anthropica: @tamang29 [medium] This now blocks every new self-loop, even when the user starts and ends on different handles or sides. The old code explicitly allowed same-node loops unless the handle was identical, and Apollon still has self-loop handling in layout and labels. Keep the same-node plus same-handle guard instead of rejecting all same-node drops.

In `library/lib/hooks/useConnect.ts`, the new-edge drop fallback rejects every same-node connection. Fix this by restoring the previous same-node/same-handle-only rejection, adapted for freeform anchors, so users can still create self-loops between different node sides or anchors.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The freeform endpoint direction looks useful, but this currently regresses manual edge bends and same-node loop creation. CI is also failing in e2e-tests for the bend persistence regression; see the inline comments.

library/lib/hooks/useConnect.ts
Claudia-Anthropica: @tamang29 Nice, that handles it.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @tamang29 Looks good now, thanks.

Claudia-Anthropica (approved):
@tamang29 All feedback addressed, nice work.

library/lib/hooks/useStraightPathEdge.ts
```diff
@@ -31,47 +44,226 @@ export interface StraightPathEdgeData {
...
     getEdgeMarkerStyles(type)
 
   const padding = markerPadding ?? EDGES.MARKER_PADDING
+  const allNodes = getNodes()
+  const sourceNode =
+    allNodes.find((node) => node.id === source) ?? getNode(source)!
+  const targetNode =
+    allNodes.find((node) => node.id === target) ?? getNode(target)!
+  const sourceAbsolutePosition = useMemo(() => {
+    if (sourceNodePositionX != null && sourceNodePositionY != null) {
```
Claudia-Anthropica, line 168: @tamang29 [medium] Anchored straight edges resolve the node rectangle from `useDiagramStore` when an anchor exists, so they ignore the remote-drag overlay that is only applied to the React Flow `nodes` prop. In collaboration, a peer dragging or resizing a node will see the node move live while these anchored endpoints stay on the old persisted position until the drag settles. Resolve the anchor from the displayed React Flow node (`getNodes()` / node lookup) and use the diagram store only as a fallback or subscription.

In `library/lib/hooks/useStraightPathEdge.ts`, anchored endpoints prefer persisted `useDiagramStore` node geometry over the React Flow node geometry. Fix this by resolving `sourceRect` and `targetRect` from the displayed React Flow nodes first so remote drag/resize overlays move anchored endpoints live; keep store data only as a fallback/subscription.

</details>

library/lib/hooks/useStepPathEdge.ts
```diff
@@ -155,19 +243,105 @@ export const useStepPathEdge = ({
...
-  const targetNode = getNode(target)!
   const allNodes = getNodes()
+  const sourceNode =
+    allNodes.find((node) => node.id === source) ?? getNode(source)!
+  const targetNode =
+    allNodes.find((node) => node.id === target) ?? getNode(target)!
 
   const sourceAbsolutePosition = useMemo(() => {
-    if (!sourceNode) return { x: sourceX, y: sourceY }
+    if (sourceNodePositionX != null && sourceNodePositionY != null) {
```
Claudia-Anthropica, line 260: @tamang29 [medium] Same issue for step edges: once `sourceAnchor` or `targetAnchor` exists, this branch pins endpoint geometry to persisted store positions instead of the nodes React Flow is currently displaying. Remote collaboration drag/resize frames are deliberately awareness-only, so anchored step edges visibly detach from the moving node until the settled write lands. Prefer the displayed React Flow node geometry here as well.

In `library/lib/hooks/useStepPathEdge.ts`, anchored step endpoints use persisted `useDiagramStore` node geometry before React Flow display nodes. Fix this by computing anchored rectangles from the displayed React Flow nodes first so collaboration remote drag/resize overlays keep endpoints attached during live movement; use the store values only as fallback/subscription.

</details>

Claudia-Anthropica (changes requested):
@tamang29 The implementation is close, and the new coverage across diagram types is useful. There is one collaboration regression in the anchored endpoint geometry though: freeform endpoints resolve from persisted store node positions, so they stop following remote live-drag overlays until the drag settles. See the two hook comments inline.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @tamang29 This is still using the store geometry first: `sourceAbsolutePosition` and `targetAbsolutePosition` return the `useDiagramStore` coordinates before falling back to the displayed React Flow node, and the rect dimensions do the same. Please flip that order so the displayed node drives anchored endpoints during live remote drag/resize, with the store values only as fallback/subscription.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @tamang29 Same here: the displayed React Flow node is found, but the memo still returns `sourceNodePositionX/Y` and `targetNodePositionX/Y` from the store first, and the width/height prefer store values too. Step-edge anchors need to resolve from the displayed node first so collaboration overlays keep the endpoint attached while the node is moving live.

Claudia-Anthropica (changes requested):
@tamang29 The two anchored endpoint geometry threads are still open: both hooks still prefer persisted store coordinates/dimensions over the displayed React Flow nodes. Please flip that resolution order so collaboration drag/resize overlays keep endpoints attached live.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @tamang29 This is still open after the latest refactor. `sourceAbsolutePosition` and `targetAbsolutePosition` return the store coordinates before `getPositionOnCanvas(sourceNode, allNodes)`, and `sourceRect` / `targetRect` still prefer the store widths and heights. The render-time anchor rectangle needs to be built from the displayed React Flow node first; the extracted endpoint lookup only fixes drag snap target selection.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @tamang29 Same state here after the latest commits. `sourceAbsolutePosition` and `targetAbsolutePosition` still short-circuit on the store values, and the rect dimensions still prefer `sourceNodeWidth` / `targetNodeWidth` over the displayed nodes. Please make the step-edge render anchor use the React Flow node rect first, with store geometry only as fallback/subscription.

Claudia-Anthropica (changes requested):
@tamang29 The new commits clean up the endpoint lookup refactor, but the two anchored endpoint geometry threads are still open. Both straight and step edges still resolve render-time anchor rectangles from store geometry before React Flow display nodes, so collaboration drag/resize overlays can still detach endpoints.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @tamang29 Still open on the current head. The only new commit after the last review touched changesets, and this hook still returns the store coordinates and dimensions before the displayed React Flow node geometry.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @tamang29 Still open on the current head. The latest commit did not touch this hook, and step-edge anchors still prefer `sourceNodePositionX/Y`, `targetNodePositionX/Y`, and the store widths/heights over the displayed React Flow nodes.

Claudia-Anthropica (changes requested):
@tamang29 The latest commit only consolidates changesets, so the two anchored endpoint geometry issues are still open. Keeping this as changes requested until straight and step edge anchors resolve from the displayed React Flow node before store geometry.

library/lib/hooks/useStraightPathEdge.ts
FelixTJDietrich: Fixed in 77368fee. `sourceAbsolutePosition`/`targetAbsolutePosition` now resolve from the displayed React Flow node (`getPositionOnCanvas(node, getNodes())`) first, and `sourceRect`/`targetRect` prefer the displayed node's width/height; the subscribed `useDiagramStore` geometry is only a fallback and the re-render subscription. So a peer's live drag/resize overlay keeps anchored straight endpoints attached during the movement. Local follow behaviour is unchanged (44 edge e2e green).

library/lib/hooks/useStepPathEdge.ts
FelixTJDietrich: Fixed in 77368fee — same change for step edges. The render anchor rectangle now comes from the displayed React Flow node first (position and dimensions), with store geometry kept only as fallback/subscription, so collaboration overlays keep anchored step endpoints attached while the node moves live.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

library/lib/hooks/useStepPathEdge.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@tamang29 All feedback addressed, nice work.

PR #801, opened 2026-07-06
fix(library): stop overhanging & hidden hit areas from stealing clicks (node handles, resize, edge toolbar)

FelixTJDietrich:
## Summary

Selecting an element in the editor could make its **invisible or overhanging hit areas steal clicks** meant for something else — the same class of bug fixed for the node toolbar in #791, now on a node's other floating affordances and on edges.

Two symptoms, one root cause (a hit area larger than the thing you can actually see):

1. **A selected node blocked the node beneath it.** A selected node is lifted above its neighbours (`elevateNodesOnSelect`), and its connection arcs (~14 px past the edge) and corner resize handles kept capturing the pointer, so you couldn't click an overlapping node exactly where those affordances overhang it.
2. **An edge selected itself from empty canvas.** The edge edit/delete toolbar is anchored to an always-present `<foreignObject>` sitting ~50 px *off* the edge line. That empty box captured clicks, so clicking blank canvas near an edge selected it. An edge should select only by clicking its line.

The fix gates each affordance so it captures the pointer **only when you're actually about to use it**, and makes the edge toolbar anchor transparent to the pointer while keeping its buttons clickable.

## Release note

Clicking near an edge or an overlapping node now behaves as expected: an edge is selected only by clicking its line, and selecting a node no longer blocks clicks on a node overlapping it.

## Implementation notes

**Node handles & resize controls (`app.css`).** On a fine pointer these are armed (`pointer-events: all`) only while the node is `:hover`ed — every gesture that uses them starts by hovering — so a selected-but-not-hovered node lets clicks fall through. This required explicitly overriding React Flow's unconditional `.connectionindicator { pointer-events: all }`.

**Touch parity (`@media (hover: none), (pointer: coarse)`).** Touch devices never fire `:hover`, so the rules above would leave a selected node's affordances ungrabbable (you couldn't start a connection or resize with a finger — a regression vs. `main`, where `.selected` armed them). A coarse-pointer fallback arms them on selection instead. A pure-touch pointer can't hover to disambiguate an overlap anyway; hybrid devices follow their **primary** pointer.

**Guidance mode stays inert.** Click-to-connect guidance drives connections from the circle handles and hides the arcs. The arc-`::before` hover/coarse rules are scoped `:not(.apollon-editor--connection-guidance)` so they can't out-specify the "arcs off" rule and re-arm an *invisible* arc that would swallow clicks over neighbours.

**Edge toolbar anchor (`CustomEdgeToolBar.tsx`).** The `<foreignObject>` is set `pointer-events: none`; its buttons re-enable via `.apollon-edge-toolbar > *`. Anchoring is geometric, so the popover still positions correctly.

## Steps for testing

1. Overlap two class nodes. Select one (elevating it) and click the other where they overlap — the second selects (previously the first's handles/resize swallowed the click).
2. Click blank canvas a few px beside an edge — nothing selects. Click the edge line — it selects. Select an edge and use its edit/delete buttons — still work.
3. On touch / mobile emulation: select a node and start a connection or resize with a finger — still works.
4. Enter click-to-connect guidance and hover nodes — no invisible arc steals the connection.

## Testing done

- New Playwright specs reproduce each bug (they fail on `main`) and lock in the fix: `node-affordance-click-through.spec.ts` (overlap block, resize block, hover-connect guard, touch grab, guidance-arc inertness) and `edge-toolbar-anchor-click-through.spec.ts` (anchor doesn't select, line still selects, buttons still work).
- Full webapp e2e suite green (139 passed); library lint/build clean; main-entry size unchanged (45.4 kB / 85 kB budget — CSS-only, no JS impact).

## Checklist

- [x] Tested the changes (new e2e + full suite)
- [x] Changesets added (`@tumaet/apollon` patch × 2)
- [x] No new lint errors; bundle size within budget
- [x] No user-facing API change; no migration needed

### DIFF
--- a/library/lib/styles/app.css
+++ b/library/lib/styles/app.css
@@ -651,6 +651,7 @@
   background-color: var(--apollon-primary, #3e8acc);
   opacity: 0.4;
   box-sizing: border-box;
+
   /* Off by default; only grabbable while the node is hovered (or in click-to-
      connect guidance mode) — see the hit-area note near the hover rules below. */
   pointer-events: none;
@@ -736,11 +737,13 @@
 .react-flow__handle.connectionindicator {
   pointer-events: none;
 }
+
 .react-flow__node:hover .react-flow__handle,
-.react-flow__node:hover .react-flow__handle.connectionindicator,
-.apollon-editor--connection-guidance .react-flow__handle.connectionindicator {
+.apollon-editor--connection-guidance .react-flow__handle.connectionindicator,
+.react-flow__node:hover .react-flow__handle.connectionindicator {
   pointer-events: all;
 }
+
 /* The arc ::before is armed on hover too, but scoped OUT of guidance mode: its
    selector would otherwise out-specify the guidance "arcs off" rule below and
    re-arm an invisible (opacity:0) arc, swallowing clicks over neighbours — the
--- a/library/lib/styles/app.css
+++ b/library/lib/styles/app.css
@@ -651,7 +651,9 @@
   background-color: var(--apollon-primary, #3e8acc);
   opacity: 0.4;
   box-sizing: border-box;
-  pointer-events: all;
+  /* Off by default; only grabbable while the node is hovered (or in click-to-
+     connect guidance mode) — see the hit-area note near the hover rules below. */
+  pointer-events: none;
   cursor: crosshair;
 
   /* --arc-scale is set by DefaultNodeWrapper to 1/zoom so the indicator keeps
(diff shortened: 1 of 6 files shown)


library/lib/styles/app.css
```diff
@@ -651,7 +651,9 @@
   background-color: var(--apollon-primary, #3e8acc);
   opacity: 0.4;
   box-sizing: border-box;
-  pointer-events: all;
+  /* Off by default; only grabbable while the node is hovered (or in click-to-
```
Claudia-Anthropica, line 655: @FelixTJDietrich [low] Codacy is failing here because `comment-empty-line-before` expects a blank line before this comment. Add the blank line after `box-sizing: border-box;` so the static-analysis check goes green.

In `library/lib/styles/app.css`, Codacy flags the comment before the new `pointer-events` declaration because it is not separated from the previous declaration. Insert a blank line before that comment so `comment-empty-line-before` passes.

</details>

library/lib/styles/app.css
```diff
@@ -711,24 +713,48 @@
...
+
+/* React Flow marks every connectable handle `.connectionindicator` and gives it
+   `pointer-events: all` unconditionally; override that so handles follow the
+   hover-gated model above. Click-to-connect guidance mode is exempt (its circle
+   handles are clicked without hovering their node); guidance hides the arcs, so
+   only `.connectionindicator` is re-enabled there, not the arc ::before. */
+.react-flow__handle.connectionindicator {
+  pointer-events: none;
+}
+.react-flow__node:hover .react-flow__handle,
```
Claudia-Anthropica, line 741: @FelixTJDietrich [low] Codacy also flags this selector with `rule-empty-line-before`. Add the required blank line between the previous rule and this one.

In `library/lib/styles/app.css`, Codacy flags the hover handle rule because it starts immediately after the previous rule. Insert the required blank line before this selector so `rule-empty-line-before` passes.

</details>

library/lib/styles/app.css
```diff
@@ -711,24 +713,48 @@
...
+.react-flow__node:hover .react-flow__handle.connectionindicator,
+.react-flow__node:hover .react-flow__handle.apollon-arc-handle::before,
+.apollon-editor--connection-guidance .react-flow__handle.connectionindicator {
```
Claudia-Anthropica: @FelixTJDietrich [low] Codacy's `no-descending-specificity` rule wants the guidance selector before the more specific hover selectors in this group. Reorder or split the selector list without changing the `pointer-events: all` behavior.

In `library/lib/styles/app.css`, Codacy flags the pointer-events selector list because `.apollon-editor--connection-guidance .react-flow__handle.connectionindicator` comes after the more specific hover selectors. Reorder or split this rule so the guidance selector is defined before the hover-specific selectors while preserving the same `pointer-events: all` behavior.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The behavior change and E2E coverage look solid, but Codacy is red on the changed CSS. Please clean up the three style annotations inline so the PR is mergeable.

library/lib/styles/app.css
FelixTJDietrich: Fixed in 39a7be75 — added a blank line after `box-sizing: border-box;` so the comment is separated. Codacy is now green on the latest commit. Resolving.

library/lib/styles/app.css
FelixTJDietrich: Fixed in 39a7be75 — inserted a blank line before the `.react-flow__node:hover .react-flow__handle` rule (also restores the file's usual block spacing). Codacy now passes. Resolving.

library/lib/styles/app.css
FelixTJDietrich: Fixed in 39a7be75 — reordered the group so guidance's `.connectionindicator` (0,3,0) precedes the `:hover` `.connectionindicator` (0,4,0); same declarations and specificities, no behavioural change. Codacy now passes. Resolving.

PR #802, opened 2026-07-06
fix(library): sharper edge-connection preview and endpoint handles

FelixTJDietrich:
### Summary

Stacked on top of the freeform-endpoints branch, this polishes how a connection
*previews* while you draw or reconnect it, in three user-visible ways plus one
consolidation that makes the preview trustworthy:

1. **Drawing a connection** now shows a clear dashed preview. It stays hidden for
   a stray click or tiny drag and appears once you pull away from the node, then
   snaps its end onto the exact point where the edge will attach.
2. **Reconnecting an edge's endpoint** across empty canvas now lets the edge
   follow your cursor — step-routed, keeping the edge's own arrowheads — instead
   of the endpoint staying pinned to its node until you happen to hover a target.
   Release over empty canvas and it reverts; no accidental reconnect.
3. **The endpoint reconnect grip** hugs the edge tip at every zoom level, instead
   of floating ~22px off it (which looked detached when zoomed out).

Under the hood, the preview and the actual attachment now go through **one**
shared drop-target resolver, so the preview can never promise a spot the drop
won't honour.

### Release note

Drawing a connection now shows a clear dashed preview that appears once you drag away from a node and lands exactly where the edge will attach. Dragging an edge's endpoint across empty canvas follows your cursor instead of snapping back, and the endpoint grip stays on the edge tip instead of drifting away when you zoom out.

### Implementation notes

- **One resolver, no lying preview.** A new `useFreeformDropTarget` hook is the
  single source of truth for "which node does this drop attach to." Both
  `useConnect.onConnectEnd` (the commit) and `ReconnectConnectionLine` (the
  preview) use it, so they share the same hit slop and z-order tie-break.
  Previously the ghost used a wider halo and a smallest-area tie-break, so a
  release just outside a node could show "snapped here" and then attach nothing.
- The ghost now reads React Flow's `fromNode` prop to skip the source node,
  instead of scanning every node to guess it (and drops a whole-`nodeLookup`
  store selector + an `infer` conditional type in favour of the exported types).
- The free endpoint preview reuses the existing `preserveOrthogonalEdgePoints`
  routing and sets **no** commit, so a release over empty canvas reverts with no
  state leak; listener add/remove stays symmetric with the bend-drag handler.
- The grip is anchored to the endpoint plus a small marker clearance; the
  invisible hit-target is unchanged, so grabbing to reconnect is exactly as easy.

**Considered and rejected:** shrinking the handles proportionally when zoomed out
— pressure-tested a `sqrt` down-scale and it left the bend handle ~6px at 40%
zoom (too small to grab), so the constant on-screen size is deliberately kept.

### Steps for testing

1. Open a class diagram.
2. Start a connection by dragging from a node's edge: no preview flashes for a
   tiny drag; a dashed preview appears once you pull away and snaps onto whatever
   node you hover — exactly where the edge lands on release.
3. Release over empty canvas: no edge is created.
4. Select an edge and drag one of its endpoints across empty canvas: the edge
   follows the cursor and snaps back to where it was when you release on empty
   canvas; drag onto another node to reconnect there.
5. Zoom out to ~40%, select an edge: the round reconnect grips sit right on the
   endpoints (not floating in the gap).

### Checklist

- [x] Added a changeset whose summary is written in the user's voice (`.changeset/sharp-connection-preview.md`)
- [x] Tests added or updated (`edge-ghost-preview.spec.ts`, `edge-handle-zoom.spec.ts`, `EdgeEndpointMarkers.test.tsx`)
- [x] Ran `pnpm lint && pnpm format:check && pnpm build && pnpm test` locally — green (library unit 1227 pass; edge e2e 50 pass)
- [ ] Documentation updated (not applicable)
- [ ] Screenshots or screencasts attached (please add in the UI — see note above)


Claudia-Anthropica (approved):
@FelixTJDietrich The edge preview/refinement logic looks solid: the drop resolver is shared between preview and commit, the empty-canvas reconnect preview stays non-committing, and CI is green. Approving as-is.

FelixTJDietrich:
Combined into #803 — its commits (the connection-preview refinements) were rebased on top of #800 together with the shape-aware work, so #803 now targets `feat/freeform-class-edge-endpoints` directly and supersedes this PR.

PR #803, opened 2026-07-06
feat(library): shape-aware edge connection points + connection-preview refinements

FelixTJDietrich:
### Summary

Edge endpoints attached to a node's bounding **rectangle** regardless of its real shape, and every node accepted a connection anywhere on that box no matter which handles it showed. This PR makes connection points **shape-aware** and **consistent with the handles you see**, and tightens the whole connect / reconnect interaction.

The rule: *a node's connectable points match the handles it renders.*

| Mode | Nodes | Behaviour |
| --- | --- | --- |
| `four-center` | activity start/end/merge, BPMN events & gateways, flowchart decision, petri place & transition, component & deployment interface, SFC transition branch | connects at exactly the four **N/E/S/W** points (these render only the four side handles) |
| `ellipse` | use-case oval | connects **anywhere along the curve** (it renders the full handle set) |
| `parallelogram` | flowchart input/output | connects **anywhere along the slanted outline**, with no gap on the left/right sides |
| `none` | color legend, title/description, swimlane | **not a connection target** |
| `freeform-rect` | everything else — classes, activities, tasks, actors, containers, forks, data objects, … | anywhere on the border (default) |

### Connection interaction

- **Grabbable arcs** — a connection arc responds as soon as you reach it, without first hovering the node body.
- **Nearest node** — a connection attaches to the node nearest the drop, not the top-most one, when several sit close together. Applies to both a new connection and reconnecting an endpoint.
- **Grid** — a new connection snaps to the grid the same way dragging an existing endpoint does.
- **Ghost preview** — drawing a connection shows a dashed preview that appears once you drag clear of the source and lands exactly where the edge will attach; the use-case oval draws a solid snap dot at the live attach point on the curve.
- **Straight edges** (use-case, petri-net, syntax-tree) — the reconnect grip sits on the edge and follows its angle; no arrow gap on a small endpoint drag.
- **Container nodes** (activity, package, pool, subsystem) are valid reconnect targets.

### Release note

Edges attach to each node's real shape, and a node's connectable points match the handles it shows: round and diamond nodes connect at their four N/E/S/W points, use-case ovals anywhere along their curve, and the flowchart input/output parallelogram anywhere along its slanted outline; legends, annotations and swimlanes are not connection targets. Drawing a connection shows a dashed preview that lands exactly where the edge attaches, connection arcs are easier to grab, and a connection snaps to the nearest node — and to the grid — consistently whether you are drawing a new edge or moving an existing endpoint.

### Implementation notes

- **Single source of truth** `getConnectionMode(nodeType)` in `connectionModes.ts`, kept in sync with the nodes that pass `FOUR_WAY_HANDLES_PRESET`.
- **No data-format change / no migration.** The stored anchor stays `{side, ratio}`; the mode only changes how that anchor renders (project onto the oval, shear onto the parallelogram, snap to a side midpoint). Commit, ghost preview and rendered edge all go through the same `getEdgeAnchorFromPoint` / `getEdgeAnchorPoint`, so preview == landing.
- **One drop resolver** (`useFreeformDropTarget`) shared by the commit and the ghost; it picks the node nearest the drop (distance to its box, 0 when inside) and gates out `none` targets. The reconnect drag-commit uses the same nearest-node rule.
- **Arcs** are armed on their own hit area so a direct approach grabs them; a selected, elevated node's arcs are disarmed while it isn't hovered so they don't cover an overlapping node.
- **Straight grips** rotate to the outward edge direction (step edges keep rotation 0). Container reconnect prefers a child under the pointer and falls back to the parent.

### Steps for testing

1. Use-case diagram: connect onto an oval aiming at its top-right — the endpoint lands on the curve, the ghost's snap dot matches, and the diagonal edges' reconnect grips tilt along the edge.
2. Activity/BPMN: a start/end/event/gateway offers exactly four points and connections snap to them.
3. Flowchart input/output: connect along the slanted left/right outline with no gap.
4. Place two nodes close together and connect/reconnect near the boundary — it attaches to the nearer node.
5. Move straight to a node's connection arc and start a connection without hovering the body first.
6. Swimlanes and annotations are not connection targets; classes, tasks, actors and containers still connect freely along their border.

### Checklist

- [x] Added changesets in the user's voice
- [x] PR title's Conventional Commit type matches the change
- [x] Tests added — `connectionModes.test.ts` (per-mode math); `shape-aware-connections.spec.ts` (oval, parallelogram, container reconnect, nearest-node, direct-approach grab, click-through); `edge-ghost-preview.spec.ts`
- [x] Ran lint, format, build, unit + edge e2e locally
- [ ] Documentation updated (not applicable)
- [ ] Screenshots or screencasts attached (please add in the UI)

> Stacked on #800 (`feat/freeform-class-edge-endpoints`).


library/lib/utils/connectionModes.ts
```diff
@@ -0,0 +1,168 @@
...
+  // exactly those four N/E/S/W points (keep this list in sync with the nodes
+  // that pass FOUR_WAY_HANDLES_PRESET).
+  activityInitialNode: "four-center",
+  activityFinalNode: "four-center",
+  activityMergeNode: "four-center",
+  bpmnStartEvent: "four-center",
+  bpmnIntermediateEvent: "four-center",
+  bpmnEndEvent: "four-center",
+  bpmnGateway: "four-center",
+  flowchartDecision: "four-center",
```
Claudia-Anthropica, line 48: @FelixTJDietrich [medium] This map misses `flowchartInputOutput`. That node hides the wrapper handles and renders only four custom handles, with left/right inset to match the parallelogram, so falling through to `freeform-rect` still lets freeform drops attach anywhere on the bbox, including empty slanted corners. Add it to the shape-aware mode and make `getFourCenterPoints()` return the same inset left/right points as `FlowchartInputOutput.tsx` uses.

In `library/lib/utils/connectionModes.ts`, `flowchartInputOutput` falls through to `freeform-rect` even though `library/lib/nodes/flowchart/FlowchartInputOutput.tsx` renders only four custom handles inset by 10px on the slanted sides. Add a shape-aware mode for `flowchartInputOutput` and update `getFourCenterPoints()` so its left/right points match those custom handle positions, then add a unit test for the parallelogram anchor points.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The shape-aware anchor plumbing is mostly coherent, but one custom four-handle node is still falling through to rectangle anchoring. Please fix the flowchart input/output case inline before merge.

library/lib/utils/connectionModes.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The parallelogram-specific mode keeps the endpoint on the visible outline and covers the original empty-corner issue.

Claudia-Anthropica (approved):

library/lib/utils/connectionModes.ts
```diff
@@ -0,0 +1,247 @@
...
+
+// The guiding rule: a node's connectable points must MATCH the handles it
+// renders. Nodes that render the full handle set stay `freeform-rect` (the
+// default) and connect anywhere along their border. Nodes that render only the
+// four side handles (FOUR_WAY_HANDLES_PRESET) connect at exactly those four
+// points. The use-case oval is the one shape that renders full handles but
+// isn't a rectangle, so it gets `ellipse`. Only the exceptions are listed here.
+const MODE_OVERRIDES: Record<string, ConnectionMode> = {
+  // Not connectable — legends / annotations / partition containers.
+  colorDescription: "none",
```
Claudia-Anthropica, line 32: @FelixTJDietrich [medium] This `none` list still misses `bpmnAnnotation`, even though the changeset and testing notes say annotations are not connection targets. Since `BPMNAnnotation.tsx` still renders the default wrapper handles, users can still connect to an annotation through React Flow’s valid-handle path, which bypasses `getEdgeAnchorFromPoint()` entirely. Add the BPMN annotation node to the non-connectable path and hide or disable its handles too.

In `library/lib/utils/connectionModes.ts`, BPMN annotations are described as non-connectable but `bpmnAnnotation` still falls through to `freeform-rect`, and `library/lib/nodes/bpmn/BPMNAnnotation.tsx` still renders default handles. Fix this by adding `bpmnAnnotation` to the `none` mode and updating the BPMN annotation node so its handles cannot start or end connections. Add a unit or e2e assertion that connecting to a BPMN annotation creates no edge.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The shape-aware plumbing is otherwise coherent and CI is green, but BPMN annotations are still connectable. I left the concrete blocker inline.

library/lib/utils/connectionModes.ts
Claudia-Anthropica: @FelixTJDietrich This is still open on the current head: `bpmnAnnotation` is not in `MODE_OVERRIDES`, `BPMNAnnotation` still uses `DefaultNodeWrapper` without `hiddenHandles`, and the `none` unit test still does not cover it. Leaving this unresolved.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The BPMN annotation blocker is still present on this head, so I’m keeping changes requested. Please add the BPMN annotation type to the non-connectable path and hide or disable its handles.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The BPMN annotation blocker is still open on this head, and the last commit did not touch that path. Keeping changes requested; the existing unresolved thread has the concrete fix.

library/lib/utils/connectionModes.ts
Claudia-Anthropica: @FelixTJDietrich Still unchanged on this head: `bpmnAnnotation` is missing from `MODE_OVERRIDES`, `BPMNAnnotation` still renders the default handles, and the none-mode test still does not cover it. Keeping this thread open.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The BPMN annotation blocker is still open on this head, so I am keeping changes requested. CI is green, but the bpmnAnnotation node still needs to be routed through the non-connectable path and its handles need to be hidden or disabled as described in the existing thread.

library/lib/utils/connectionModes.ts
FelixTJDietrich: Fixed in ca343836. Added `bpmnAnnotation: "none"` to `MODE_OVERRIDES`, and `BPMNAnnotation.tsx` now passes `hiddenHandles={true}` so it renders no handles — closing both the freeform anchor path and React Flow's native valid-handle path. Covered by a unit assertion (`getConnectionMode` → `none` and `getEdgeAnchorFromPoint` → `null`) and an e2e that a drop onto an annotation creates no edge and it exposes zero handles.

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):

PR #804, opened 2026-07-07
refactor(library): audit the freeform + shape-aware connection stack vs main

FelixTJDietrich:
Re-scoped the A+ bridge to the **whole `feat/freeform-class-edge-endpoints` → `main` diff** (46 files), not just the shape-aware slice — a full-scope websearch-grounded rubric + principal-engineer audit (33 agents) + adversarial verification. This branch lands the fixes into #800's branch. Every **high** finding that is safe to fix is closed; the honest remaining items are named below rather than papered over.

## Closed

**Dead code**
- Removed the hardcoded `freeformAnchorsEnabled = true` flag and every provably-dead `!freeformAnchorsEnabled` branch in `useStepPathEdge` (manual-points fallback, guard, ternary, memo deps) — behaviour-preserving.
- Removed the unused `getNormalizedHandleOffset` / `getNormalizedHandleOffsetPercent` exports (no callers).
- `getFreeformAnchorPoint`: replaced the dead trailing `top` fallback with a `never` exhaustiveness guard so a new `Position` fails the build instead of silently rendering as top.

**Type safety**
- Dropped the `getNode(...)!` non-null assertions in both path hooks that the surrounding null guards already contradict.
- Derived `handleSlotBySide` from the `HandleId` enum's declaration order instead of re-listing all 36 ids + slots as a second untyped source of truth.

**Duplication / minimalism**
- Extracted `useFreeformEndpointNode` (shared `getNodeRect` + `findFreeformEndpointNode` + `FREEFORM_ENDPOINT_SNAP_RADIUS_PX`) — these were byte-identical in both path hooks.
- Both hooks + the drop resolver now share `pickNearestConnectable` / `distanceToRect` (removed the inline point-to-rect re-implementations).

**Test correctness & races**
- **Bug:** the oval-curve test scanned the endpoint nearest the oval across ALL edges — a pre-existing edge could satisfy the assertion even if the connection never formed. Now isolates the newly-drawn edge.
- Converted the flaky sleep-then-read-snapshot assertions to retrying `expect.poll` across the follows / ghost / grab / retarget / same-node specs (negative "no edge forms" cases keep their settle). Verified stable across repeated no-retry runs.
- Trimmed the follows e2e 25 → 4 representative cases; split a collapsed `before && after` assertion.

**Verified:** tsc (0 errors), lint (0), unit (280), and 44 edge e2e (freeform-anchor, edge-interactions, shape-aware, ghost-preview, fresh-bend, handle-zoom, hit-test) — green.

## Honestly remaining (not claiming a clean A+)
Two larger refactors of tested, working code, deliberately left for a focused follow-up rather than rushed:
- **Endpoint-drag orchestration** — the ~120-line `handleEndpointPointerDown` (pointermove/up + `setEdges` commit) is still near-duplicated across the two hooks, differing only in the drag-preview shape. Extracting a shared orchestrator (resolve/preview/commit callbacks) is the main remaining minimalism/DRY win.
- **`Position` typing** — `FreeformEdgeAnchor.side` as the `Position` enum forces ~9 `"top" as Position` casts and keeps the `sideToHandleId` identity bridge; modelling connectable sides as a closed string-literal union removes both.

Plus minor: `getNodes()` called in render then in memo deps; `onConnectStart` left un-memoized while its siblings are wrapped; `EdgeEndpointMarkers` unit tests assert re-derived pixel arithmetic; the per-spec localStorage readers could live in the shared canvas helper.


Claudia-Anthropica (approved):
@FelixTJDietrich This looks solid; the shape-aware anchor logic, preview/commit alignment, and non-connectable-node handling are covered well. Approving as-is.

Claudia-Anthropica (approved):
@FelixTJDietrich The new connection-mode helper, freeform drop targeting, and preview/commit paths line up cleanly now. I did not find any new issues on this SHA, and the focused changed-unit tests pass under Node 24.15.0.

Claudia-Anthropica (approved):
@FelixTJDietrich The new commit still looks clean: the shared connection-mode helper, drop resolver, and preview/commit paths stay aligned across the shape-specific cases. I ran the focused library unit tests for the changed geometry coverage, and CI is green, so approving as-is.

PR #806, opened 2026-07-09
feat: a VS Code custom editor for .apollon files, and the theming fixes it exposed

FelixTJDietrich:
### Summary

The VS Code extension is rewritten as a `CustomTextEditorProvider` over `.apollon` files. Rather than a bespoke webview that owned the document, VS Code now owns it — so save, auto-save, undo, dirty state, hot exit, diffs, split editors and branch switches all work because they are the editor's, not because we reimplemented them.

Embedding the library in a real host immediately exposed three bugs in it. All three only appear when the theme is scoped to the editor's **own mount node** — `dataTheme`, a `theme` override, or a host stylesheet — and are invisible to the webapp, which themes the document root. They are fixed here because the extension is unusable without them. Separately, `fitView` never reserved the device safe area, so on a phone the camera framed nodes under the notch and the home indicator.

| Area | Commits |
| --- | --- |
| **The custom-editor rewrite** | `feat(vscode-extension)` |
| **Library fixes it exposed** | `fix(library)` × 3 — safe area, scoped theme, popover titles |
| **Correctness found in review** | `fix(vscode-extension)` × 2 — lifecycle holes, PNG export |
| **Structure** | `refactor(vscode-extension)` — split the provider |
| **Plumbing** | `build`, `build(webapp)`, `test(library)`, `style` |

### Release note

Apollon diagrams now open in a real VS Code editor, with saving, undo and diffs working the way they do for any other file — and embedded editors correctly follow a dark or custom theme in their popups, inputs and drag previews.

### Implementation notes

**Why the extension shrank by two workspaces.** It previously ran two extra pnpm workspaces (`editor/`, `menu/`) and mutated the user's global `files.associations` on every activation. Both are gone: one webview workspace, and no writes to user settings. The extension has never been published (`CHANGELOG.md`: *"First Marketplace release pending"*), so there is no installed base holding a stale `files.associations` entry and nothing to migrate. That is the only reason it was safe to simply delete that code rather than clean up after it.

**Writes are range edits, not rewrites.** The document is edited through `jsonc-parser`'s `modify`/`applyEdits` at the model's own path, so a hand-annotated `.apollon` file keeps its comments, trailing commas, sibling keys, indentation and line endings, and a text editor open on the same document keeps its cursor and a sane undo granularity. `detectFormatting` reads the document's own indent and EOL rather than imposing ours. Tests pin the wiring, not just the detection — deleting `formattingOptions` from `modelEdits` turns three of them red.

**An empty file is a state, not a failure.** `readDocument` returns `empty | model | invalid`. An empty `.apollon` offers a diagram-type picker; an unparseable one explains itself instead of throwing. `parseModel` checks only the `type` discriminator — structural validation belongs to the library, which rejects a bad model on mount with a better message than this extension could reconstruct from a JSON shape.

**The three library theming bugs.** Each has a one-line root cause worth stating precisely, because each is easy to "fix" somewhere that doesn't work:

1. An **unregistered custom property substitutes its `var()` at the element that declares it**, then inherits the computed result. The `--home-*` aliases the shared `@tumaet/ui` primitives paint from were declared only at `:root`, so under a scoped mount they had already frozen to the light document. Re-declared on `.apollon-editor`, exactly as the chrome ramp already was.
2. The UA stylesheet's `input { color: fieldtext }` is a **declaration on the element**, which beats inheritance, and `fieldtext` follows `color-scheme` — the OS appearance, not the editor's theme. Tailwind's Preflight resets this; the Tailwind-free bundle the embeds ship does not, so the primitives now reset `color` themselves.
3. Body-portaled popups copy **resolved token values** and never recomputed them, so a theme switch left an open menu — and every later one on the same anchor — painting the old palette.

For (3) the popups now share **one** `MutationObserver` behind `useSyncExternalStore` instead of one per popup (an editor keeps dozens of tooltips mounted). It deliberately watches `class` but not `style` on `<body>`, because a palette drag writes `body.style.overflow` on every grab and would otherwise repaint every popup. The drag ghost, which portals to `<body>`, carries both the `data-theme` attribute and the resolved values — the attribute alone is not enough for a mount themed by inline properties or a host stylesheet, which is precisely the VS Code case.

Nothing tied the bridge to `tokens.css`, so an alias added to one and not the other was invisible until an embed rendered light input borders on a dark surface. A test now asserts the bridge is a superset of what `tokens.css` scopes to `.apollon-editor`.

**PNG export produced 0-byte files for large diagrams.** The webview rasterized its own PNGs through `` → `<canvas>` → `toBlob` — the exact pipeline the library abandoned in #667. Past the browser's canvas-area cap (~16 MP on WebKit, ~268 MP on Chrome) `toBlob` returns `null` with no exception, so the export silently wrote nothing. It now goes through `@tumaet/apollon/export`, which renders into resvg's own wasm buffer, as the webapp already does. This was a duplicate of library code, and the known-broken variant of it.

The renderer and its wasm binary load on first export, so the canvas keeps its startup path. Instantiating the module needs `wasm-unsafe-eval` and fetching it needs `connect-src`; a nonce does not propagate to `import()`-ed chunks, so `script-src` also carries `cspSource`. This was verified against the CSP the host actually emits, rendered from `webviewHtml.ts` rather than retyped: with `wasm-unsafe-eval` a real PNG comes back, and without it the module fails to compile — so the check can fail.

**860 kB of dead weight left the VSIX.** `@tumaet/apollon/export` also exports `svgToPdf`, and a bundler keeps *every* export of a dynamically imported module, since nothing constrains what the namespace object is read for. So the PNG import dragged jsPDF, svg2pdf, html2canvas and dompurify into a code path the webview can never take. `pngRenderer.ts` narrows the namespace to one binding and they shake out; an eslint rule stops a future import from silently re-inflating it. Separately, `.vscodeignore`'s `node_modules/**` was anchored to the root and matched no nested tree.

**The provider was doing three jobs.** `ApollonEditorProvider` was 481 lines. Two subsystems moved out, leaving 253:

- **`DocumentSync`** — the document↔canvas engine, previously five interlocking mutable variables in a closure. It is the code where a bug loses a user's work, and it now has a name, a lifecycle and a doc comment that fits it.
- **`DiagramExporter`** — render request/response correlation, the workspace-trust guard, the file write, cancellation.

`protocol.ts` and `diagramTypes.ts` moved to `src/shared/`, which the webview imports too. That boundary was real but unenforced: nothing stopped a `vscode` import from breaking the webview bundle. An eslint rule enforces it now, and both rules were checked by writing the violating import and watching them fire.

Naming, while the files were open: the `config` message carried one field and promised a bag → `autoExportChanged`. The customEditor viewType `apollon.diagram` differed from the tree view id `apollon.diagrams` by one character → `apollon.diagramEditor`.

**What review caught, and what it means.** Three defects in the first push, all now covered by tests that fail without their fix:

- Closing a tab inside the 300 ms commit debounce cleared the timer and dropped the edit with it. VS Code never saw a dirty document, so there was no save prompt — the change was silently gone. **This was a data-loss bug.** The pending model is now written on disposal, unless the document is already closed (closing without saving is a choice; resurrecting it would undo that).
- An external write to a *transiently invalid* document — mid-typing in a split JSON editor — returned before cancelling the pending canvas write, so the debounce fired 300 ms later and overwrote what had just been typed.
- The manual export command bypassed the workspace-trust guard that auto-export applied, even though the manifest promises image exports stay off until the workspace is trusted. The guard now sits on the shared write path.

The extension had **no** coverage of the provider at all — only of the pure document functions — which is precisely why these survived. `tests/vscodeStub.ts` now stands in for the `vscode` module: events fire synchronously and `applyEdit` really rewrites the document, so tests assert on the text the document ends up holding rather than on which API happened to be called. Auto-export writes files to a user's disk on every save and *nothing asserted a file was ever written*; the round trip, both payload encodings, the failure path, the trust refusal, the scaffold and its non-empty guard are now pinned too. 44 extension tests, up from 31.

Every one was falsified: eleven targeted mutations of the source, each turning red exactly the tests that guard the behaviour it broke, and green again on revert.

**CI gap closed.** `pr-health-checks.yml` linted `vscode-extension` but never typechecked, tested or built it, so its unit tests were decorative. A `vscode-extension-checks` job now runs `typecheck`, `test` and `build:all` against a built library, and it is wired into the `PR Health Gate`.

**Safe area.** The overlay grid's padding is the safe area, and chrome insets are measured from each control's own box, which begins inside that padding — so the two stack without double-counting. `respectInsets: false` and a per-side `padding` override both still opt out of chrome; neither can opt out of the safe area, which is a hardware constraint rather than breathing room.

**Trade-off — the CSS size budget.** `standalone/webapp` capped gzipped CSS at 35 kB and `main` had grown to exactly 35,001 B, so the next CSS change of any size would have failed the check on its own. This PR adds 56 B. Raised to 36 kB rather than shaving bytes out of a correctness fix; the current 35.06 kB still leaves the budget able to catch a real regression.

**Marketplace metadata.** The manifest had no `icon`, which the Marketplace requires as a PNG of at least 128×128 (`media/` held only the monochrome activity-bar glyph, which is recoloured by the theme and would have listed as a black square). `media/icon.png` is the project's own brand asset downscaled to 256×256. `apollon.autoExport` gains `enumDescriptions` and `scope: resource`, since export writes are per-folder. `Other` leaves `categories`.

### Deliberately not in this PR

An audit for duplication turned up four more things. None are this branch's doing, and folding them in would make it harder to review, not better:

- **`standalone/webapp`'s `diagramTypes` identity map** re-declares all 13 diagram types purely to hand back string constants identical to the library's runtime `UMLDiagramType`. Unlike the extension — where importing the runtime value would drag the browser editor into the Node host bundle, which is why `src/shared/diagramTypes.ts` legitimately declares its own — the webapp already ships the editor, so this is free to delete. Its sibling `shortLabelsByType` and `tilesByType` are typed `Partial<Record<…>>`, so a 14th diagram type would silently render with no label rather than failing the build.
- **The model schema version is a bare string in five places.** `"4.0.0"` is the repo-wide convention (the v3→v4 converter, the webapp store, every fixture) and the extension follows it; the outlier is `library/lib/apollon-editor.tsx`, whose `model` getter emits `"4.1.0"`. `UMLModel["version"]` is typed `` `4.${number}.${number}` ``, so nothing catches the disagreement. The library should export the constant.
- **`DocumentSync.write()`** rolls `written` back if `applyEdit` is refused. If an unrelated external edit lands inside that await, the rollback restores text the document no longer holds. This behaviour is carried over verbatim from the pre-refactor closure, not introduced here. Refused `applyEdit` is rare and the window is a microtask; still worth a ticket.
- **`knip` runs in no workflow**, so the config in `knip.json` only gates whoever runs it locally. (It is green here, for the first time — the dead `nodeTypeLabel` export it was flagging turned out to be a stale duplicate of the i18n label table, missing two diagram prefixes the live one also lacked. That is the `fix(library)` popover-titles commit.)

`vsce` is deliberately **not** a workspace dependency — the release workflow installs it into a temp prefix with `--ignore-scripts`, keeping its native credential store and Azure SDK out of every contributor's `pnpm install`. Recorded in `README.dev.md` so it doesn't get "fixed".

### Steps for testing

**The extension**

1. `pnpm install && pnpm run build:lib`
2. `pnpm --filter apollon-vscode run build:all`, then <kbd>F5</kbd> to launch an Extension Development Host.
3. Create an empty `diagram.apollon` and open it — you should get a diagram-type picker, not an error. Pick one, then <kbd>Cmd/Ctrl+Z</kbd>: you are back to an empty file.
4. Drag a node. Hit <kbd>Cmd/Ctrl+S</kbd>, then <kbd>Cmd/Ctrl+Z</kbd> — undo, dirty state and save behave like any text file.
5. Add a `// comment` and a sibling key to the JSON via **Reopen with Text Editor**, go back to the canvas, move a node, save. The comment, the sibling key and your indentation survive.
6. Set `apollon.autoExport` to `png` and save a large diagram. You get a real PNG next to the file, not a 0-byte one.
7. Switch VS Code between Light Modern, Dark Modern and both high-contrast themes with the canvas open. Popovers, inputs, tooltips and the palette follow, live.

**The library fixes** (these are the ones with regression coverage)

- `pnpm --filter @tumaet/apollon run test` — `portalTheme`, `portalThemeReactivity`, `draggableGhostTheme`, `nodeTypeLabel`, `fitView`.
- `pnpm --filter @tumaet/webapp run test:e2e -- scoped-theme-form-controls` — asserts the rendered `border-color` and `color` of an input inside a body-portaled popover under an inline dark theme.
- `pnpm --filter apollon-vscode run test` — the provider lifecycle, including the data-loss path above.

Every new test was falsified against the unfixed code before being kept: reverting the `.apollon-editor { --home-* }` block turns the border assertion red (it reads `0.84` luminance — a light border on a dark surface), and removing `text-inherit` turns the ink assertion red (`rgb(0, 0, 0)`).

Two caveats worth stating, because both are cases where a test passed for the wrong reason until it was checked:

- An early `vscode` stub modelled `applyEdit` as *refusing* a closed document. That made the "don't resurrect a closed document" test pass whether or not the guard existed. Real VS Code applies a `WorkspaceEdit` to files that aren't open — that is how cross-file refactors work — so the stub was corrected and the test now actually fails without the guard.
- The scoped-theme e2e test first used `data-theme="dark"`, which also sets `color-scheme: dark` and therefore made `fieldtext` white on its own — the fix was unfalsifiable. Rewritten to an inline theme; it then still passed, because the webapp loads Tailwind Preflight, which already resets `input { color }`. The bug is only observable in the Preflight-free `components.css` the embeds ship, which the test now loads directly.

### Screenshots / screencasts

⚠️ **Not attached.** This is a UI change and screenshots are required — they need to be captured from a running Extension Development Host (the editor across the four VS Code themes, the empty-file picker, and a popover with an input under a dark theme) and added before merge. Flagging rather than silently checking the box.

### Checklist

- [x] Tests added or updated

<sub>Five changesets ship with this PR: four `@tumaet/apollon` patches (safe area, scoped-theme form controls, drag ghost, popover titles), of which the form-controls one also bumps `@tumaet/ui`; plus one `@tumaet/webapp` patch (iOS keyboard). `apollon-vscode` is on the Changesets ignore list and versions through its own release workflow.</sub>

<sub>Note for the reviewer: the e2e suite is flaky under parallel load with retries disabled — `main` fails 1–2 different tests per run in that mode, as does this branch, on a different test each time. CI runs with retries, where this branch is 211 passed / 1 flaky. The palette-drag ghost tests were A/B'd specifically (this branch's `DraggableGhost` vs `main`'s, 18 repeats each across 4 workers): both 18/18.</sub>


vscode-extension/src/apollonEditorProvider.ts
```diff
@@ -0,0 +1,465 @@
...
+      this.active = { webview: panel.webview, document }
+    }
+
+    // Last: loading the document can only start once we are listening, or the
+    // webview's `ready` lands before anyone is there to answer it.
+    panel.webview.html = renderWebviewHtml(panel.webview, this.extensionUri)
+
+    panel.onDidDispose(() => {
+      if (timer) {
+        clearTimeout(timer)
```
Claudia-Anthropica: @FelixTJDietrich [critical] Clearing the debounce timer here drops any canvas edit that has not yet been written to the `TextDocument`. If a user drags a node or creates a diagram and closes the tab within 300ms, VS Code never sees a dirty document and the change disappears without a save prompt. Flush the pending model before disposal, or remove the window where edits live only in memory.

In `vscode-extension/src/apollonEditorProvider.ts`, `panel.onDidDispose` clears the commit debounce without applying `pending`, so quick tab closes can lose canvas edits. Fix this by ensuring pending model changes are committed to the VS Code `TextDocument` before the panel is disposed, or by removing the debounce window where edits are only held in memory.

</details>

vscode-extension/src/apollonEditorProvider.ts
```diff
@@ -0,0 +1,465 @@
...
+        if (event.document.uri.toString() !== document.uri.toString()) {
+          return
+        }
+        const text = document.getText()
+        if (text === committed) {
+          return
+        }
+        committed = text
+        const state = readDocument(text)
+        if (state.kind === "invalid") {
```
Claudia-Anthropica: @FelixTJDietrich [high] This invalid branch returns before clearing `pending` and the debounce timer. If a split text editor makes the JSON temporarily invalid while a canvas update is pending, the timer still fires and writes the stale canvas model over the user's text. Treat every external document change as superseding the pending canvas write before deciding whether to post an update to the webview.

In `vscode-extension/src/apollonEditorProvider.ts`, the invalid external-edit path leaves `pending` and `timer` intact. Fix this by clearing and canceling pending canvas commits for any external text change before returning on invalid JSON, so the debounce cannot overwrite split-editor edits.

</details>

pnpm-workspace.yaml
```diff
@@ -3,8 +3,7 @@ packages:
   - "packages/*"
   - "standalone/*"
   - "vscode-extension"
-  - "vscode-extension/menu"
-  - "vscode-extension/editor"
+  - "vscode-extension/webview"
```
Claudia-Anthropica, line 6: @FelixTJDietrich [high] The workspace rename is not reflected in the Dockerfiles, and Docker CI is failing before install because both images still copy `vscode-extension/menu/package.json` and `vscode-extension/editor/package.json`. Update `standalone/webapp/Dockerfile` and `standalone/server/Dockerfile` to copy the new `vscode-extension/webview/package.json` path instead.

In `standalone/webapp/Dockerfile` and `standalone/server/Dockerfile`, the install stage still copies deleted `vscode-extension/menu/package.json` and `vscode-extension/editor/package.json` files after the workspace was replaced by `vscode-extension/webview`. Replace those COPY lines with the new webview package path, or remove them if they are no longer needed, so Docker image builds pass.

</details>

vscode-extension/src/apollonEditorProvider.ts
```diff
@@ -0,0 +1,465 @@
...
+      return
+    }
+    const { webview, document } = this.active
+    if (!isOnDisk(document.uri)) {
+      void vscode.window.showInformationMessage(
+        "Save the diagram to a file before exporting an image."
+      )
+      return
+    }
+    await this.writeExport(document.uri, webview, format, { silent: false })
```
Claudia-Anthropica: @FelixTJDietrich [medium] Manual export bypasses the workspace-trust guard that auto-export uses below, so an untrusted workspace can still write sibling SVG/PNG files even though the manifest says image exports are disabled until trust. Put the `vscode.workspace.isTrusted` check on the shared export path, or add it here before requesting the render.

In `vscode-extension/src/apollonEditorProvider.ts`, `exportActiveDiagram()` writes sibling image files without checking `vscode.workspace.isTrusted`, while `runAutoExport()` does check it and the manifest promises exports are disabled in untrusted workspaces. Fix this by enforcing the workspace-trust guard for manual exports too, preferably in the shared `writeExport()` path.

</details>

vscode-extension/webview/src/App.tsx
```diff
@@ -0,0 +1,263 @@
...
+/** What the canvas shows, driven entirely by what the document holds. */
+type View =
+  | { kind: "loading" }
+  | { kind: "empty" }
+  | { kind: "invalid"; reason: string }
+  | { kind: "editor"; initial: UMLModel }
+
+function DiagramPicker() {
+  return (
+    <div className="apollon-vscode-notice">
```
Claudia-Anthropica, line 37: @FelixTJDietrich [medium] This PR changes visible VS Code editor states and scoped-theme UI behavior, but there are no real screenshots or screencasts attached. The Apollon review rules require captures for UI changes, so please add the Extension Development Host screenshots for the empty picker, the editor across VS Code themes, and a themed popover/input before merge.

For PR #806, visible UI changes lack the required review screenshots. Capture and attach real screenshots or screencasts from a running VS Code Extension Development Host covering the empty-file picker, editor theme variants, and a scoped dark popover/input; do not use generated or mock screenshots.

</details>

library/lib/styles/app.css
```diff
@@ -95,11 +82,11 @@
     "botleft  botcenter botright"
     "footer   footer    footer";
   pointer-events: none;
-  padding-top: var(--safe-area-inset-top, 0);
-  padding-left: var(--safe-area-inset-left, 0);
-  padding-right: var(--safe-area-inset-right, 0);
+  padding-top: var(--safe-area-inset-top, 0px);
```
Claudia-Anthropica: @FelixTJDietrich [low] Codacy is red on these direct padding fallbacks because they use `0px`. These three declarations are not inside `calc()` or `max()`, so use unitless `0` here and keep the typed `0px` fallbacks only where CSS math needs them.

In `library/lib/styles/app.css`, the direct `padding-top`, `padding-left`, and `padding-right` fallbacks use `0px` and trigger Codacy `length-zero-no-unit`. Change those three direct padding fallbacks to `0`, while keeping typed `0px` fallbacks in `calc()` or other CSS math contexts.

</details>

packages/ui/src/styles/components.css
```diff
@@ -190,13 +190,19 @@
 
   /* ── Input (data-slot="input") ────────────────────────────────────────── */
   /* h-9 (36px) matches the default button height for aligned form rows. */
+  /* `text-inherit` is load-bearing, not cosmetic: the UA stylesheet declares
```
Claudia-Anthropica, line 194: @FelixTJDietrich [nit] Codacy expects a blank line before this new block comment, so this contributes to the red static-analysis check. Add one blank line between the short `h-9` comment and this longer explanation.

In `packages/ui/src/styles/components.css`, the new `text-inherit` block comment is adjacent to the previous comment and trips Codacy `comment-empty-line-before`. Insert a blank line before the new block comment.

</details>

vscode-extension/webview/src/index.css
```diff
@@ -0,0 +1,203 @@
...
+  --apollon-border: var(
+    --vscode-contrastBorder,
+    var(--vscode-editorWidget-border, var(--vscode-panel-border, #ced4da))
+  );
+  --apollon-border-subtle: var(
+    --vscode-contrastBorder,
+    var(--vscode-panel-border, #e9ecef)
+  );
+
+  --apollon-danger: var(--vscode-errorForeground, #d33);
```
Claudia-Anthropica, line 91: @FelixTJDietrich [nit] Codacy flags the blank line before this custom property. Remove the extra blank line so the custom property group stays contiguous and the static-analysis check goes green.

In `vscode-extension/webview/src/index.css`, Codacy flags the empty line before `--apollon-danger` as `custom-property-empty-line-before`. Remove that blank line so the custom properties remain contiguous.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The custom-editor rewrite is headed in the right direction, but I found a data-loss window in the debounce lifecycle, stale Dockerfile workspace paths that are breaking Docker CI, and a trust guard gap on manual export. Codacy also has a few CSS style annotations to clear, and the required UI screenshots still need to be attached.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Still open: the invalid-document branch returns before clearing `pending`/`timer`, so the pending canvas write can still fire and overwrite text-editor edits.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Still open: `onDidDispose` still only clears the timer, so a pending canvas edit can be dropped before it reaches the `TextDocument`.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Still open: `exportActiveDiagram()` still reaches `writeExport()` without checking `vscode.workspace.isTrusted`; only auto-export is guarded.

pnpm-workspace.yaml
Claudia-Anthropica: @FelixTJDietrich Looks good now, the Dockerfiles copy `vscode-extension/webview/package.json`.

vscode-extension/webview/src/App.tsx
Claudia-Anthropica: @FelixTJDietrich Still open: I do not see screenshots in the PR body or comments yet, and the PR body still says they are not attached.

library/lib/styles/app.css
Claudia-Anthropica: @FelixTJDietrich Still open: the direct safe-area padding fallbacks are still `0px`, so this likely still contributes to the Codacy failure.

packages/ui/src/styles/components.css
Claudia-Anthropica: @FelixTJDietrich Still open: there is still no blank line before the new `text-inherit` block comment.

vscode-extension/webview/src/index.css
Claudia-Anthropica: @FelixTJDietrich Still open: the blank line before `--apollon-danger` is still there.

Claudia-Anthropica (changes requested):
@FelixTJDietrich One Docker issue is fixed and I resolved that thread, but the debounce/external-edit data-loss paths, manual export trust guard, missing screenshots, and Codacy CSS items are still open in the existing threads.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: Confirmed and fixed in 55aa18b8 — this was real data loss, not just a dropped timer. `onDidDispose` now commits `pending` while the document is still alive, so the edit lands, the document goes dirty, and VS Code's close/hot-exit takes over.

One nuance I added on top of the suggestion: the flush is skipped when `document.isClosed`. Closing without saving is a deliberate choice, and applying a `WorkspaceEdit` to a closed file would resurrect it — VS Code applies edits to files that are not open, which is how cross-file refactors work.

Covered by `tests/apollonEditorProvider.test.ts` (`closing the editor`). Removing the flush, or removing the `isClosed` guard, each turns exactly one test red.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: Confirmed and fixed in 55aa18b8. `pending = undefined; cancelCommit()` now runs for *every* external change, before the validity check — the invalid branch only decides whether to mirror the document back into the canvas, never whether to keep the debounce armed.

Test: `an external edit › cancels a pending canvas write even while the document is invalid`. It types `{ "title": "half-typed` into the document with a canvas edit pending, advances past the debounce, and asserts the half-typed text survives.

pnpm-workspace.yaml
FelixTJDietrich: Right, and thank you — this is exactly what Docker CI was failing on. Fixed in 9954cdfd: both Dockerfiles now copy `vscode-extension/webview/package.json`, and `.dockerignore` had the same two stale source paths.

Verified by building both images locally (`docker build -f standalone/webapp/Dockerfile .` and the server equivalent) rather than by pushing and waiting. Docker Build Gate is green.

Worth noting for the future: the image builds run in a workflow outside `PR Health Gate`, so a full local `lint && build && test` can be green while this is red.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: Confirmed and fixed in 55aa18b8. The guard moved to the shared `writeExport()` path, which both entry points go through, and it is now removed from `runAutoExport()` so there is exactly one place that enforces it. The manual command additionally surfaces a warning explaining why nothing was written — silently doing nothing on an explicit user action would be worse.

Test: `an untrusted workspace › refuses a manual export and explains why`, which also asserts the canvas is never even asked to render.

vscode-extension/webview/src/App.tsx
FelixTJDietrich: Agreed, and I have not attached them — I cannot capture a running Extension Development Host from here, and I would rather leave the checklist box unticked than fill it with anything generated. Called out explicitly in the PR body under *Screenshots / screencasts*: the empty-file picker, the editor across the four VS Code themes, and a scoped-dark popover with an input still need real captures before merge.

library/lib/styles/app.css
FelixTJDietrich: Agreed, and your reasoning is the right one. Fixed in 55aa18b8: the three direct `padding-*` fallbacks are now unitless `0`, and the `0px` fallbacks stay inside the `calc()` on the next line, where a unitless `<number>` would invalidate the expression. Added a one-line comment there so the asymmetry does not look accidental and get "cleaned up" later.

packages/ui/src/styles/components.css
FelixTJDietrich: Fixed in 55aa18b8 — blank line added between the `h-9` note and the `text-inherit` explanation.

vscode-extension/webview/src/index.css
FelixTJDietrich: Fixed in 55aa18b8 — blank line removed so the custom-property group stays contiguous.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. The disposal path now flushes the pending model before teardown without resurrecting already closed documents.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now. Cancelling before the validity check closes the overwrite path I was worried about.

pnpm-workspace.yaml
Claudia-Anthropica: @FelixTJDietrich Looks good, thanks for checking the Docker builds too.

vscode-extension/src/apollonEditorProvider.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now. Putting the trust check in `writeExport()` covers both manual and auto export cleanly.

vscode-extension/webview/src/App.tsx
Claudia-Anthropica: @FelixTJDietrich Still open then. I agree generated screenshots would be worse, but the review requirement is real captures before merge, so I’m leaving this thread unresolved.

library/lib/styles/app.css
Claudia-Anthropica: @FelixTJDietrich Looks good now, the direct fallbacks are unitless and the typed zero is limited to the CSS math case.

packages/ui/src/styles/components.css
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

vscode-extension/webview/src/index.css
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The code issues I had flagged are addressed and CI is green now. I'm keeping changes requested only because the required real screenshots or screencasts for the VS Code UI/scoped-theme changes are still missing.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: The provider has since been split. This now lives in `DocumentSync.dispose()` (`src/documentSync.ts`): it cancels the debounce, then writes the pending model while the document is still alive, skipping only a document that was already closed. Covered by `writes a canvas edit that the debounce had not committed yet` and `leaves a document that was already closed alone`.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: Now `DocumentSync.onDocumentChanged()` (`src/documentSync.ts`), which drops the pending model and cancels the timer unconditionally, before it decides whether the new text is worth mirroring. Covered by `cancels a pending canvas write even while the document is invalid`.

vscode-extension/src/apollonEditorProvider.ts
FelixTJDietrich: The trust guard now sits in `DiagramExporter.write()` (`src/diagramExporter.ts`), the single path both the command and auto-export take. Covered by `refuses in an untrusted workspace and explains why`, which also asserts the canvas is never asked to render.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The code feedback I checked is addressed, and the targeted VS Code extension tests pass locally. I’m keeping changes requested because the required real screenshots or screencasts for the VS Code UI and scoped-theme changes are still missing, and the GitHub checks for this SHA are still pending.

FelixTJDietrich:
Correction to this PR's description, found while debugging the release run it triggered ([#808](https://github.com/ls1intum/Apollon/pull/808)).

Under *Implementation notes* I wrote that the extension "has never been published (`CHANGELOG.md`: *First Marketplace release pending*), so there is no installed base holding a stale `files.associations` entry and nothing to migrate."

That is wrong. I trusted the changelog instead of querying the Marketplace. `tumaet.apollon-vscode` has been listed since **2024-11-18**, last at **0.0.17**. I downloaded that VSIX and read its bundle: it writes `"files.associations": { "*.apollon": "json" }` into **global** user settings on every activation. An installed base exists, and it carries that line.

The conclusion mostly survives, but not for the reason I gave:

- `.apollon` files still open in the Apollon editor — the custom editor matches on `filenamePattern`, not on language, so the stale setting cannot hijack it.
- Reopening one as text shows it as JSON rather than under this extension's `apollon` language, grammar and file icon.
- Nothing removes the setting. Deleting it restores both.

Deleting the code that wrote it was still right; claiming there was nobody to migrate was not. #808 corrects `CHANGELOG.md` and asks whether we want a one-time, self-deleting migration.

Two related things #808 fixes, both surfaced by merging this PR:

1. `Release VS Code Extension` fired on merge and tried to publish `1.0.0`, because it wakes on any edit to `vscode-extension/package.json` and this PR edited the icon, categories and settings schema. **Nothing was published** — the publish job crashed first, on a missing `.nvmrc`.
2. That crash (the publish job never checks the repo out) is also fixed there.

PR #809, opened 2026-07-09
feat(ci): version the VS Code extension with the rest of Apollon

FelixTJDietrich:
> Follows #808 (merged), which stopped a stray manifest edit from publishing. This PR makes the version itself meaningful.

### Summary

`vscode-extension/package.json` has said `1.0.0` at **every commit I can trace back to 2026-02-03**. Nobody chose it; it is `npm init`'s default. The Marketplace listing it supposedly describes is at `0.0.17`, published 2024-11 — so that field has never driven a release. Meanwhile `@tumaet/apollon`, `@tumaet/webapp` and `@tumaet/server` are all at `5.1.0`.

`apollon-vscode` joins them in a Changesets `fixed` group. It carries Apollon's version from now on.

### What happens next

Nothing, until you merge a **Version Packages** PR. Then all four move together:

| Package | Before | After |
| --- | --- | --- |
| `@tumaet/apollon` | `5.1.0` | `5.1.1` |
| `@tumaet/webapp` | `5.1.0` | `5.1.1` |
| `@tumaet/server` | `5.1.0` | `5.1.1` |
| `apollon-vscode` | `1.0.0` | **`5.1.1`** |

**No version is hand-edited here**, and that is deliberate. Changing `vscode-extension/package.json`'s version in *this* PR would fire `Release VS Code Extension` on merge and publish immediately — the premature release #808 exists to prevent. Changesets aligns a `fixed`-group member to the group's highest version on its own, so `1.0.0 → 5.1.1` happens inside the Version Packages PR, and merging *that* is the deliberate release.

`0.0.17 → 5.1.1` is monotonic, so the Marketplace accepts it. `version-monotonicity.yml` is satisfied for the same reason.

### The trade, stated plainly

`fixed` means **a changeset touching only the extension still bumps and republishes the library and the standalone images**, at a new version, with identical content. That is the cost of one product number.

The alternative is `linked`, which bumps only what changed but lets the numbers drift apart again — which is the thing we're fixing. I've gone with `fixed` and written the trade into `docs/contributor/deployment/npm-publishing.md` so the next person doesn't rediscover it. **Say the word and I'll switch to `linked`.**

Two smaller consequences:

- **`Version Bump (VS Code extension)` is deleted.** It existed only to hand-bump a package Changesets was told to ignore. Extension work now records a `pnpm changeset` like everything else.
- **`apollon-vscode` is no longer in `ignore`, which was also what kept it away from npm.** Nothing publishes it there — `release.yml` runs Changesets in version-only mode — but I've written that constraint into the docs rather than leave it implicit. It ships to the Marketplace, never to npm.

### CHANGELOG

`vscode-extension/CHANGELOG.md` is now what Changesets writes, matching `library/CHANGELOG.md`. Its prose preamble had to go: Changesets inserts each release **directly after the first line**, so anything under the heading gets shoved beneath every future release. I verified this — the first simulation produced exactly that mess.

The `0.0.17` migration note moved into the changeset. Review caught that it would have gone nowhere: `release-vscode-extension.yml` built its GitHub Release with `--generate-notes` and a static body, ignoring the changelog the library and standalone tracks both feed through `extract-changelog.mjs`. Fixed in `6d45a1c` — the extension now uses the same path, so the note reaches the release body:

```
# apollon-vscode

## 5.1.1

### Patch Changes

- Apollon for VS Code now carries the same version as the rest of Apollon…
  If you have used the extension before, version `0.0.17` wrote
  "files.associations": { "*.apollon": "json" } into your user settings…
```

### Verification

Every claim above was run, not reasoned. In a throwaway git worktree off this branch, with the real `pnpm exec changeset version`:

- All four packages land on `5.1.1`; `@tumaet/ui` (not in the group) goes `0.0.1 → 0.0.2` as before; `apollon-vscode-webview` stays ignored and private.
- The generated `vscode-extension/CHANGELOG.md` has the shape shown above.
- `node scripts/check-release-taxonomy.mjs` — in sync.
- `node scripts/sync-library-version.mjs --check` — CDN refs match.
- `pnpm exec changeset status` — parses the new config, lists the five patches.
- `actionlint .github/workflows/*.yml` — **0 findings** across every workflow (was 2 pre-existing before #808).
- `prettier --check` — clean.

One thing I could not run locally: `changeset version` with the real `changelog-github` generator, which resolves each changeset's commit on GitHub and fails on an unpushed one. I swapped in the default generator for the simulation; the GitHub generator only changes how release lines are *worded*, not which versions are computed. It runs for real in the Version Packages PR.

### Steps for testing

1. Merge this.
2. `release.yml` opens (or updates) the **Version Packages** PR. Confirm it contains `apollon-vscode: 1.0.0 → 5.1.1` alongside the other three.
3. Merging that PR is the release. It changes `vscode-extension/package.json`, so `Release VS Code Extension` fires, `check` sees the version changed, and the VSIX is built, attested and published to the Marketplace + Open VSX as `5.1.1`.
4. Before that, worth one `workflow_dispatch` of `Release VS Code Extension` with `dry_run: true` to prove #808's `.nvmrc` fix on a real runner — it can only fail where there is no checkout.

### Checklist

- [ ] Tests added or updated

<sub>No tests: this is release configuration. It was instead exercised end-to-end with the real `changeset version` in a throwaway worktree, as described above.</sub>


.changeset/apollon-vscode-joins-the-product-version.md
```diff
@@ -0,0 +1,7 @@
+---
+"apollon-vscode": patch
+---
+
+Apollon for VS Code now carries the same version as the rest of Apollon, so the extension you install and the library it renders with are one number.
+
+If you have used the extension before, version `0.0.17` wrote `"files.associations": { "*.apollon": "json" }` into your user settings every time it started. Nothing writes or needs it now. Diagrams open in the Apollon editor either way, but while that line remains, opening one as text shows it as plain JSON instead of an Apollon diagram. Deleting it restores the file icon and syntax highlighting.
```
Claudia-Anthropica, line 7: @FelixTJDietrich [medium] This migration note will land in `vscode-extension/CHANGELOG.md`, but the VS Code release job still ignores that changelog and creates the GitHub Release with `--generate-notes` plus a static note. That means the `0.0.17` settings cleanup guidance does not actually reach the `apollon-vscode@...` release body, even though this PR moves the text out of the manual changelog preamble. Please feed the generated VS Code changelog into the release notes path.

In `.changeset/apollon-vscode-joins-the-product-version.md`, the migration note is only consumed into `vscode-extension/CHANGELOG.md`, while `.github/workflows/release-vscode-extension.yml` does not include that changelog in GitHub Release notes. Fix this by updating the VS Code release workflow to build its release notes from `scripts/extract-changelog.mjs "$VERSION" vscode-extension` and append the existing install/Marketplace footer, mirroring the library/standalone fallback behavior.

</details>

.changeset/config.json
```diff
@@ -3,8 +3,10 @@
...
   "access": "public",
   "baseBranch": "main",
-  "fixed": [["@tumaet/webapp", "@tumaet/server"]],
+  "fixed": [
+    ["@tumaet/apollon", "@tumaet/webapp", "@tumaet/server", "apollon-vscode"]
+  ],
   "updateInternalDependencies": "patch",
   "privatePackages": { "version": true, "tag": false },
-  "ignore": ["@tumaet/apollon-docs", "apollon-vscode", "apollon-vscode-webview"]
+  "ignore": ["@tumaet/apollon-docs", "apollon-vscode-webview"]
```
Claudia-Anthropica, line 11: @FelixTJDietrich [medium] With `apollon-vscode` removed from `ignore`, the contributor release-notes guide is now wrong: it still says the extension is excluded and VS Code-only changes should skip changesets. That would make future extension PRs miss the version/changelog entry this new flow depends on; while updating it, also remove the remaining old wording in `npm-publishing.md` that says the library is never dragged up, because the fixed group now deliberately can do that.

In `.changeset/config.json`, `apollon-vscode` is now Changesets-tracked, but the docs still describe the old exclusion/manual flow. Fix `docs/contributor/development/release-notes.md` so user-visible VS Code extension changes require a changeset, and update `docs/contributor/deployment/npm-publishing.md` to remove contradictory old cascade wording such as `the library is never dragged up`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The fixed-group config itself behaves as described; I verified `changeset status` and a temp `changeset version` simulation produce `5.1.1` across the four packages. I’m requesting changes because the VS Code release job still does not publish the generated changelog text, and a contributor guide still tells people to skip VS Code changesets.

.changeset/apollon-vscode-joins-the-product-version.md
FelixTJDietrich: Correct, and it invalidated a claim in this PR's own description. Fixed in 6d45a1c.

`release-vscode-extension.yml` now builds the release body the way the library and standalone tracks always have: `node scripts/extract-changelog.mjs "$VERSION" vscode-extension` into `--notes-file`, with `--generate-notes` kept only as the fallback when a version carried no changeset. The `release` job gains `fetch-depth: 0`, because the script recovers each entry's Conventional Commit type from the SHA in the changelog line to group the notes.

Verified against the exact CHANGELOG that `changeset version` generates: the script emits `### Bug Fixes` / `### Dependencies` sections carrying the `0.0.17` settings guidance, and returns empty for a version with no section so the fallback engages.

Worth noting `npm-publishing.md` already asserted the release body for **each track** came from `extract-changelog.mjs`. That was untrue for the extension before this commit; now it is true.

.changeset/config.json
FelixTJDietrich: Both right. Fixed in 6d45a1c.

`docs/contributor/development/release-notes.md` said three lines carry changesets and told authors to *skip* one for VS Code-only work — which is exactly the changeset the new flow needs them to write. It now says four lines share one version via the `fixed` group, and that a VS Code-only change needs a changeset because it is what earns the extension its version bump and its Marketplace release notes. The `#ignore` sentence now names only the docs site and the webview sub-package, which is what remains there.

`npm-publishing.md` claimed `cascade-standalone-bump.mjs` "raises a floor only ... and the library is never dragged up". Inside a `fixed` group that is no longer true in either direction: a bump earned by any member raises all of them, so a standalone-only or extension-only change does drag the library up. The paragraph now says so plainly and calls it the price of one product number. The cascade script is kept as a safety net, but it is no longer the mechanism.

.changeset/apollon-vscode-joins-the-product-version.md
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

.changeset/config.json
Claudia-Anthropica: @FelixTJDietrich Nice, that handles it.

Claudia-Anthropica (approved):

FelixTJDietrich:
Rebased onto `main` (`8e01555`, `9ff0482`). #808 was squash-merged as `d16a68d`, so this branch was still carrying the pre-squash commit — same content, different SHA, hence the conflict. Dropped it with `git rebase --onto origin/main 8eae0bb4`; no textual conflicts.

Re-verified after the rebase rather than assuming, since the merge base moved:

- #808's work is intact on the result (version guard, `.nvmrc` sparse checkout, `fetch-depth: 2`), and this branch's only change to `release-vscode-extension.yml` is the release-notes path.
- `actionlint` clean across all workflows; `check-release-taxonomy`, `sync-library-version --check`, `prettier`, `changeset status` all pass.
- The full `pnpm changeset:version` still lands `@tumaet/apollon`, `@tumaet/webapp`, `@tumaet/server` and `apollon-vscode` on `5.1.1`.
- On that simulated version-bump commit the guard reads `before=1.0.0 now=5.1.1` and correctly proceeds to publish, and `extract-changelog.mjs` renders the extension's release body.

PR #810, opened 2026-07-09
refactor: name every private workspace @tumaet/<role>

FelixTJDietrich:
Follow-up to #806 / #808 / #809. Pure naming; no behaviour change.

## The rule

The scope carries the organization, the name carries the role. `@tumaet/apollon` is the product; `@tumaet/ui`, `@tumaet/webapp` and `@tumaet/server` are named for the job they do. Two workspaces broke that rule for no reason:

| Before | After |
| --- | --- |
| `apollon-vscode-webview` | `@tumaet/vscode-webview` |
| `@tumaet/apollon-docs` | `@tumaet/docs` |

The webview was unscoped purely because it inherited the look of its sibling — it is `private`, has no `engines.vscode`, and is never packaged as a VSIX. The docs site repeated the product name that the scope already implies.

## What is deliberately *not* renamed

`apollon-vscode` keeps its unscoped name, for two independent reasons — either alone is decisive:

1. **A scope is illegal.** The Marketplace requires an extension `name` to match `[a-z0-9][a-z0-9-]*`; `vsce package` rejects `@tumaet/apollon-vscode` outright.
2. **A rename would orphan every install.** Marketplace identity is `publisher.name`. A different name is a *new listing*, and the 38 existing installs would never see another update.

`tumaet.apollon-vscode` already *is* the scoped name, spelled the way the Marketplace spells scopes. That reasoning is now recorded in `project-structure.md` and `AGENTS.md` so nobody "fixes" it later.

Also corrects a line in `AGENTS.md` left stale by #806: it still described `menu/` and `editor/` sub-packages, which that PR replaced with a single `webview/`.

## Why this is safe

Nothing depends on either package via `workspace:*` — both are reached only through `pnpm --filter`. The rename is inert to resolution, and **`pnpm-lock.yaml` is untouched**, because pnpm keys importers by path rather than by name.

The one place a stale name would have been load-bearing is `.changeset/config.json#ignore`. I falsified that rather than assuming: putting the old names back makes `changeset status` die with a `ValidationError` from `@changesets/config`, so a half-done rename fails loudly at release time instead of silently un-ignoring the docs site.

## Verification

Run locally against this branch:

- `pnpm install` — lockfile unchanged
- `pnpm -r list` — all 8 workspaces resolve under the new names
- `pnpm --filter @tumaet/vscode-webview` / `--filter @tumaet/docs` — both filters resolve
- `pnpm --filter @tumaet/apollon run build && pnpm --filter apollon-vscode run build:all` — extension builds through the renamed webview filter
- `pnpm --filter apollon-vscode lint | typecheck | test` — clean; **44/44** tests pass
- `pnpm --filter @tumaet/docs run typecheck` — clean (the exact command `pr-health-checks.yml` runs)
- `pnpm exec changeset status` — config accepted
- `pnpm format:check` — clean

No changeset: both packages are private and excluded from every release line, so nothing user-visible changes.


Claudia-Anthropica (approved):
@FelixTJDietrich This looks good to me. The renamed workspace filters resolve locally, Changesets accepts the new ignore entries, and the docs/VS Code checks that exercise these names are passing.

PR #812, opened 2026-07-09
fix(vscode): publish the extension under the aet-tum publisher

FelixTJDietrich:
The `5.1.1` release reached npm but died at `vsce publish`:

```
Publishing 'tumaet.apollon-vscode v5.1.1'...
Access Denied: ca588d25-… needs the following permission(s) on the resource
/TUMAET/apollon-vscode: Make changes to, share, or view certificate of an existing extension
```

This is not a PAT to repair. It is the wrong publisher.

## What the archaeology showed

Both Apollon listings on the Marketplace were published from **`ls1intum/apollon-vscode`**, a repository that has been archived since #806 absorbed it into this monorepo. Their VSIX manifests both carry `repository: github.com/ls1intum/apollon-vscode`.

| Listing | Version | Installs | Last updated |
| --- | --- | --- | --- |
| `tumaet.apollon-vscode` | 0.0.17 | 38 | 2025-01-04 |
| `aet-tum.apollon-extension` | 1.0.0 | 10 | 2026-03-02 |

Its maintainers migrated from `tumaet` to `aet-tum` in March, stranding the 38 installs themselves. `aet-tum` is the organization's live publisher — it also owns `aet-tum.iris-thaumantias`, and the matching Open VSX namespace already exists with `artemis-scorpio` and `iris-thaumantias` in it. `tumaet` publishes nothing else, on either registry.

Meanwhile this monorepo's manifest has read `publisher: tumaet` since the extension was first committed in February. It has always pointed at the dead listing; nothing had ever tried to publish it until now.

New identity: **`aet-tum.apollon-vscode`** (the ID is free). The git tag stays `apollon-vscode@X.Y.Z`, the pnpm workspace stays `apollon-vscode`, and `.changeset/config.json` is untouched — only `publisher` changes.

## The duplication that let this drift

`vsce` and `ovsx` derive what they publish from the manifest's `publisher` and `name`. The workflow **restated** `tumaet.apollon-vscode` in three separate places: the Marketplace version query, the install footer, and the release-body links. Plus the tag was hardcoded as `apollon-vscode@…`. So the listing the workflow *queried* and the listing `vsce` *wrote to* were free to disagree — and did.

The `check` job now reads `publisher`, `name`, and `publisher.name` from the manifest once and passes them downstream. There is exactly one source for the identity.

## Open VSX is now optional

Requested explicitly. `OVSX_PAT` does not currently exist in any secret scope, so with the Marketplace PAT fixed it would have become the *next* failure. Now: no token → the step is skipped with a `::notice::`, and the release body names only the registries actually reached. A release that says "Published to Open VSX" when it wasn't is worse than one that omits it.

A `vsce verify-pat <publisher>` preflight also runs before the VSIX is uploaded, so a missing publisher grant fails in seconds with a clear message instead of after a full build.

## Verification

- The rewritten `jq`-built Marketplace query returns `publish=true` for `aet-tum.apollon-vscode`, and `publish=false` for both `aet-tum.iris-thaumantias@0.4.8` and `tumaet.apollon-vscode@0.0.17`. The negative controls prove the absence is real rather than a query I broke.
- `vsce verify-pat` exits `1` on an **unset**, **empty**, and **invalid** PAT, tested in a clean `node:24-slim` container. It never prompts — so the hang I was worried about doesn't exist, and no guard was added for it.
- The release-notes footer renders correctly for `OVSX=true`, `OVSX=false`, and `OVSX=''` (the backfill case where the publish job was skipped).
- Tag is still `apollon-vscode@5.1.1`. Workflow YAML parses; all four jobs and their `needs` wiring check out.
- `pnpm format:check` clean; 44/44 extension unit tests pass.

## Merging this does not publish

By design (#808): the `check` job compares the version against the parent commit, and `5.1.1` is unchanged. Merging is inert. To ship, run **Actions → Release VS Code Extension → Run workflow** on `main` with `dry_run: false` — `workflow_dispatch` skips the version-changed guard, sees `5.1.1` absent from `aet-tum.apollon-vscode`, and publishes, tags, and cuts the Release.

## Out of scope, but you should know

The `vscode-marketplace` environment has **no protection rules at all** (`protection_rules: []`, `deployment_branch_policy: null`), while `npm-publish` has a branch policy. The setup docs instruct you to add "Selected branches → main" and required reviewers; that was never applied. Any branch that can trigger the workflow can currently read `VSCE_PAT`.

The two legacy listings are now dead ends. Worth deprecating both in favour of `aet-tum.apollon-vscode` — I have not verified the exact Marketplace deprecation flow, so I did not write it into the docs.

### DIFF
--- a/.github/workflows/release-vscode-extension.yml
+++ b/.github/workflows/release-vscode-extension.yml
@@ -27,6 +27,9 @@ jobs:
       contents: read
     outputs:
       version: ${{ steps.v.outputs.version }}
+      publisher: ${{ steps.v.outputs.publisher }}
+      name: ${{ steps.v.outputs.name }}
+      extension-id: ${{ steps.v.outputs.extension-id }}
       publish: ${{ steps.v.outputs.publish }}
       release: ${{ steps.v.outputs.release }}
     steps:
@@ -49,6 +52,18 @@ jobs:
           fi
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
+          # The manifest is the only source for the Marketplace identity: `vsce`
+          # and `ovsx` derive what they publish from `publisher` and `name`, so
+          # anything this workflow states about the extension has to be read from
+          # there too. A second copy here would decide which listing gets queried
+          # while the manifest decided which listing gets written.
+          PUBLISHER=$(node -p "require('./vscode-extension/package.json').publisher")
+          NAME=$(node -p "require('./vscode-extension/package.json').name")
+          EXT_ID="${PUBLISHER}.${NAME}"
+          echo "publisher=$PUBLISHER" >> "$GITHUB_OUTPUT"
+          echo "name=$NAME" >> "$GITHUB_OUTPUT"
+          echo "extension-id=$EXT_ID" >> "$GITHUB_OUTPUT"
+
           # A release is a version change, and nothing else. This workflow wakes
           # on any edit to the manifest, which also carries the icon, the
           # commands and the settings schema — so an unrelated edit would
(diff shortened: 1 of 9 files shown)


Claudia-Anthropica (approved):
@FelixTJDietrich The publisher rename and release workflow wiring look consistent, and the dynamic Marketplace/Open VSX handling is scoped to the manifest identity. I did not find any code issues; the extension-specific checks are green, with only broader CI jobs still pending when I reviewed.

PR #813, opened 2026-07-09
fix(vscode): publish as aet-tum.apollon-extension

FelixTJDietrich:
Follow-up to #812. The publisher was right; the **name** is not available.

```
Verify the Marketplace PAT can write this publisher  ✅
  The Personal Access Token verification succeeded for the publisher 'aet-tum'.

Publish to VS Marketplace  ❌
  Publishing 'aet-tum.apollon-vscode v5.1.1'...
  ERROR The extension 'apollon-vscode' already exists in the Marketplace.
        Please use a different 'name' in the package.json file
```

## Why

The Marketplace reserves an extension `name` across **all** publishers, not per publisher. `apollon-vscode` is held by `tumaet.apollon-vscode` — the abandoned 0.0.17 listing from `ls1intum/apollon-vscode`, the archived repo this workspace replaced. Freeing the name would mean deleting an extension 38 people still have installed.

I checked the rule rather than trusting the message: sampling ~980 distinct extension names turned up 15 owned by more than one publisher (`python` belongs to `ms-python`, `tht13` and `stevedower`). Every one of them was created **2016–2019**, before the constraint existed. They're grandfathered; new names are globally unique.

## The fix

`aet-tum.apollon-extension` is the listing this project **already owns**, published from the same archived repo. `vsce` therefore *updates* it instead of *creating* it, so the conflict cannot arise — and its 10 installs receive 5.1.1 instead of being stranded.

`package.json` `name` is simultaneously the pnpm workspace name, the git tag prefix and the Marketplace extension name, so this reaches the root scripts, the changesets `fixed` group, both workflows and the docs. The `apollon-vscode-*` CSS classes and the webview's control ids are **not** the package name and are deliberately untouched.

| | before | after |
| --- | --- | --- |
| Marketplace ID | `aet-tum.apollon-vscode` (rejected) | `aet-tum.apollon-extension` |
| Git tag | `apollon-vscode@5.1.1` | `apollon-extension@5.1.1` |
| pnpm filter | `apollon-vscode` | `apollon-extension` |

The name reads oddly next to a `vscode-extension/` directory, and it can't be improved while the dead listing squats the better one. `project-structure.md` now records exactly that, so nobody "fixes" it and rediscovers this the hard way.

## Verification

- `pnpm -r list` resolves `apollon-extension`; **`pnpm-lock.yaml` is unchanged**, because pnpm keys importers by path, not name.
- `changeset status` accepts the `fixed` group. Falsified: restoring the old name makes it die with `ValidationError: The package … "apollon-vscode" … does not match any package in the project` — so that entry is load-bearing, and a half-done rename fails loudly at release time.
- `pnpm --filter apollon-extension` → lint, typecheck, `build:all` all clean; **44/44** tests pass. `pnpm format:check` clean.
- `vsce package` emits `apollon-extension-5.1.1.vsix`, and the manifest inside reads `aet-tum.apollon-extension@5.1.1`.
- `vsce show aet-tum.apollon-extension` resolves the listing — confirming the update path. It also reports **☆☆☆☆☆ (0)** ratings, which retires the "4.45★" I mistakenly cited earlier: that number is the Marketplace's zero-rating prior.
- The exact Marketplace query the `check` job runs returns `publish=true` for `aet-tum.apollon-extension@5.1.1`.

## Shipping

Merging is inert — the `check` job compares `version` against the parent commit and `5.1.1` is unchanged (#808's guard). To publish: **Actions → Release VS Code Extension → Run workflow** on `main`, `dry_run: false`.

Still outstanding, unrelated to this PR: `OVSX_PAT` is unset (Open VSX is skipped by design since #812), and the `vscode-marketplace` environment has no protection rules while `npm-publish` has a branch policy.


docs/user/getting-started/setup.md
```diff
@@ -15,7 +15,7 @@ Open <https://apollon.aet.cit.tum.de> in a browser. No install, no account. Diag
 
 ## VS Code extension
 
-Install **Apollon** from the [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=aet-tum.apollon-vscode) (or [Open VSX](https://open-vsx.org/extension/aet-tum/apollon-vscode)). Diagrams live next to your code as `.apollon` files.
+Install **Apollon** from the [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=aet-tum.apollon-extension) (or [Open VSX](https://open-vsx.org/extension/aet-tum/apollon-extension)). Diagrams live next to your code as `.apollon` files.
```
Claudia-Anthropica, line 18: @FelixTJDietrich [low] This Open VSX link points at an extension that does not exist (`/api/aet-tum/apollon-extension` returns 404), and the release workflow skips Open VSX when `OVSX_PAT` is unset. Either drop the Open VSX install path for now or point it at the registry entry you actually publish.

In `docs/user/getting-started/setup.md`, the Open VSX install link was changed to `aet-tum/apollon-extension`, but that extension is not published and the workflow can skip Open VSX when `OVSX_PAT` is unset. Fix this by removing the Open VSX link until the new extension is published there, or by pointing the docs and support table at the registry entry that actually exists.

</details>

Claudia-Anthropica (approved):
@FelixTJDietrich The rename looks consistent across the workspace filters, Changesets config, Marketplace release flow, and docs. CI is green; I left one low-priority note about the Open VSX link.

PR #814, opened 2026-07-09
fix(vscode): claim a free display name and fix the Marketplace README links

FelixTJDietrich:
Third server-side rejection, third distinct rule:

```
Verify the Marketplace PAT can write this publisher  ✅
Publishing 'aet-tum.apollon-extension v5.1.1'...
ERROR This extension display name is taken. Please try a different one.
```

Rather than fix it and find out about rule #4 next run, I put three principal engineers on it — one researching the Marketplace's documented validation surface, one auditing the artifact vsce actually produces, one red-teaming the result.

## The blocker

The Marketplace reserves `displayName` across **all** publishers, exactly as it reserves `name`:

> "The display name must be unique to the Marketplace."
> — [Extension Manifest reference](https://code.visualstudio.com/api/references/extension-manifest)

The rule went live 2023-02-01 ([vscode-discussions#433](https://github.com/microsoft/vscode-discussions/discussions/433)) — which is why 29 duplicate display names I sampled from the gallery are all pre-2023 grandfathering.

`Apollon` is held by the abandoned `tumaet.apollon-vscode`. **`Apollon - TUM`** is the display name our `aet-tum.apollon-extension` listing already carries, so the update keeps it, and it matches its sibling `aet-tum.iris-thaumantias` → `Artemis - TUM`.

## The defect nobody would have reported

While auditing the listing, both relative links in the Marketplace README were found to **404 silently**. vsce rewrites relative README links against the repository root and ignores `repository.directory` ([vscode-vsce#980](https://github.com/microsoft/vscode-vsce/issues/980), open), so an extension in a subdirectory ships links one level too high — with **zero warnings** at package time.

Measured with `curl` against the URLs inside the actual VSIX:

| README source | Shipped as | |
| --- | --- | --- |
| `[README.dev.md](./README.dev.md)` | `.../blob/HEAD/README.dev.md` | **404** |
| `[LICENSE](../LICENSE)` | `.../blob/HEAD/../LICENSE` | **404** |

`--baseContentUrl` fixes the first. It **cannot** fix the second: vsce keeps `../` verbatim and GitHub does not normalize `..` in a blob path, so it stays broken even with a correct base. That link now points at `./LICENSE.txt`, which exists in the extension root and already ships inside the VSIX.

The base URLs are declared once, in `package:vsix`, and the release workflow packages through that script instead of calling `vsce` itself — the same duplication that produced the wrong extension ID two PRs ago.

## Also corrected

The README advertised `SFC`; the shipped label is `Sequential function chart`. All 13 diagram types now match `src/shared/diagramTypes.ts` exactly.

## Verified against the artifact that will actually be uploaded

Not the source tree — the VSIX:

| | |
| --- | --- |
| id | `aet-tum.apollon-extension` |
| displayName | `Apollon - TUM` |
| version | `5.1.1` (listing is at 1.0.0) |
| icon | `media/icon.png`, present, 256×256 PNG (the live listing currently has **no icon asset at all**) |
| LICENSE.txt | present → no interactive vsce license prompt to hang CI |
| `src/`, sourcemaps | 0 files |
| every URL in packaged `readme.md` | **200** |

The one non-200 is `npmjs.com` → `403`, which is Cloudflare bot mitigation against `curl`; `registry.npmjs.org` returns `200` for the package.

I also **executed** the changed CI packaging path locally rather than reasoning about it: installed `vsce` into a temp prefix, put it on `PATH`, and ran `pnpm run package:vsix` from `vscode-extension/` — packages cleanly, glob matches exactly one VSIX.

Red team confirmed via vsce v3.9.1 `src/publish.ts` that `--packagePath` reads the manifest **from inside the VSIX**, never from disk — which is what makes the publish job's `.nvmrc`-only sparse checkout safe, and means the base-URL flags can't disagree with what ships.

`lint`, `typecheck`, `format:check` clean; **44/44** tests pass.

## Shipping

Merging is still inert — `version` is unchanged, so #808's push guard correctly skips. Publish with **Actions → Release VS Code Extension → Run workflow** on `main`, `dry_run: false`.

## Residual risk, stated honestly

`vsce package` performs only *local* validation; `name` / `displayName` uniqueness is enforced **only** at the server on publish, and there is no documented dry-run. Every rule I could find a trusted source for is now satisfied, and the two identity fields are values this listing **already holds** — the strongest guarantee available short of publishing. Unverified by any trusted source: whether a reserved-name list exists, and any `description` length cap.

Unrelated and still open: `OVSX_PAT` unset (Open VSX skipped by design), and the `vscode-marketplace` environment has no protection rules.


Claudia-Anthropica (approved):
@FelixTJDietrich Looks good. The Marketplace identity and README link handling are consistent, and the checks are green.

PR #816, opened 2026-07-17
feat(library): add element tags to color a group of class members at once

### LINKED ISSUE #50: Feature Request / Allow coloring of diagram attributes/methods
I have a feature request for Apollon (related ArTEMiS https://github.com/ls1intum/ArTEMiS/pull/539):

## Background

For the interactive instructions of the programming exercises I want to switch from plantUML to the more user friendly Apollon.
The instructions are updated with the build result, indicating in the UML which implementations are successful, are false, etc.:

image

planUML is using markup, so I set up a placeholder for the test color:

class Context {
  -<color:testsColor(testAttributes[Context])>dates: List<Date></color>
  +<color:testsColor(testMethods[Context])>sort()</color>
}

The placeholder is then replaced on parsing with the appropriate color (red if failed, etc.).

## Feature Request

To switch to Apollon I would need the following functionality:

1. I need to be able to address an attribute/method of an element by an id that I can define (that is the name of the test case for me). Multiple elements could have the same id.
2. I need to be able to set the color (green, red, grey) of an attribute/method of an element

Ideally for both I would not have to go through the element, but could just query the diagram:
E.g. 
attributes = diagram('testAttributes[Context]');
diagram.setColor(attributes, 'red')

FelixTJDietrich:
### Summary

Closes #50.

Coloring an individual class attribute or method already worked, and hosts could already tint elements by id with `setElementHighlights`. The missing piece was **addressing**: element ids are per-element UUIDs, so a host wanting to color "everything test case X covers" had to keep its own `test case → [ids]` map outside the diagram and rebuild it whenever an element was deleted and re-created.

This adds an optional `tags: string[]` to element data — on nodes and on each class attribute/method — plus one query: `editor.getElementIdsByTag(tag)`. Many elements may share a tag, so a single lookup returns the whole group, which the host hands to the existing highlight overlay. That closes the loop the original issue asked for: tag the members a programming test covers, then repaint them green/red/grey on every build.

**Tag authoring is opt-in and host-configurable** (added in review): it's off by default and turned on with a `tags` option — `true` for free-form tagging, or an object that supplies a fixed `available` vocabulary and decides whether users may also create their own (`allowCreate`). Hosts can also assign tags without the UI via `editor.setElementTags(id, tags)`. In the editor, tags show inline as compact removable chips beside a tag button that opens a searchable combobox. Color editing was moved into a popover in the same pass, so the edit panel no longer expands inline.

### Release note

Adds element tags — short labels you attach to a class, an attribute, or a method, where many elements can share the same one. Tag every member a programming test covers, then ask the editor for that tag with `editor.getElementIdsByTag("testAttributes[Context]")` and hand the result straight to `setElementHighlights`. That is enough to repaint a sample solution green, red, or grey from each build result, so the same diagram can show a different picture to every student without the saved model ever changing.

Tag authoring is off by default and enabled with the new `tags` option: `true` for free-form tags, or an object to offer a fixed vocabulary (`available`) and decide whether users may create their own (`allowCreate`). A host can also set tags without the UI via `editor.setElementTags(id, tags)`. When enabled, each class, attribute, and method shows its tags as removable chips next to a tag button that opens a searchable combobox. Colors are now edited in a popover too, so the edit panel no longer grows inline. Tags are part of the diagram, so they survive export, import, and copy/paste — and a diagram with no tags is untouched.

### Implementation notes

**Why `tags` and not `references`.** In a UML tool a "reference" reads as a pointer to another model element, which is the wrong mental model — these are opaque, host-defined labels. `tags` matches the prior art ([draw.io tags](https://www.drawio.com/doc/faq/shape-metadata), Kubernetes labels) and is honestly many-to-many.

**Why inline on `data` and not a top-level map.** A `Record<tag, elementId[]>` next to `assessments` reads well but needs cross-cutting cleanup on every delete, duplicate, and paste — the dangling-id problem `assessments` already has. Inline, tags die with their element and clone with it for free, and they merge per element (last-writer-wins) exactly like `fillColor` already does.

**Authored vs. ephemeral — the important line.** Tags are authored content: serialized, collaborated, undoable. Build results are *not*, and stay in `setElementHighlights`, which is never written to the model. Writing per-build colors into `fillColor` would destroy authored styling, leak one student's results into a shared document, and add a Yjs write on every CI run.

**Wire format stays `4.1.0`, deliberately.** `tags` is optional and rides in the open `data` envelope, so nothing needs a migration. A bump to `4.2.0` was considered and rejected: the webapp's version-history `structuralFingerprint` hashes `version`, so a bump would mark every existing diagram as changed and produce one spurious autosave snapshot each. A model without tags is byte-identical through the normalizer (asserted in `versionConverter.test.ts`).

**Opt-in + configuration plumbing.** The `tags` option resolves to a `{ enabled, available, allowCreate }` config in the metadata store (the same home + reactive-setter pattern as `labels`), read by the UI through a `useTagConfig` hook. `allowCreate` defaults to `false` when a vocabulary is given, `true` otherwise.

**Idiomatic shadcn, adapted.** The editor bundle is Tailwind-free, so it can't consume `@tumaet/ui`'s Tailwind components directly. Both new popovers follow the repo's established idiom (the existing `EditorColorPicker`): a Base UI `Popover` structured like the shadcn component, styled Tailwind-free via `data-slot`s + `--apollon-*` tokens in `app.css`. The tag combobox is the shadcn Popover-plus-Command pattern rendered that way.

**One traversal, one write rule.** `taggableElements` is the single walk shared by the query *and* the import normalizer, so a reader and a writer can never disagree about what an element is. `applyTags` / `withTags` is the single "normalize, and omit the key when empty" rule used by every write path — an empty `tags: []` is not equal to an absent key for the store's `deepEqual`, so a stray `[]` would be a permanent Yjs write and undo entry.

### Steps for testing

1. **Storybook — `pnpm storybook`, then *Editor → Element Tags*.**
   - *Authoring: Tag a Class Member* — the story opts tagging on with a small vocabulary. Each class/attribute/method shows its tags as chips beside a tag button; the button opens a combobox to search the vocabulary, toggle a tag (checkmark), or **Create "…"** a new one. Colors open in a popover from the paint-roller.
   - *Coloring: Build Results by Tag* — a tagged class diagram repainted from a fake build result: `Animal.name` / `run()` green, `Dog.breed` green **from the same tag as `Animal.name`** (one test spanning two classes), `Dog.bark()` red, `Vehicle.drive()` grey.
2. **In the editor** (`pnpm dev`, `new ApollonEditor(el, { enablePopups: true, tags: true })`): open a class popover → tag an attribute → export the diagram; the tag is in the JSON. Re-import: still there. Copy/paste the class: the copy keeps the tags. With `tags` omitted, no tag UI appears.
3. **Programmatic:** `editor.setElementTags("<attr-id>", ["x"])` then `editor.getElementIdsByTag("x")` returns that id; `editor.setTags({ available: ["a","b"] })` restricts the picker to that vocabulary.
4. **Round-trip a diagram with no tags** — the JSON should come back unchanged.
5. `pnpm --filter @tumaet/apollon test` — unit coverage for the normalizer, the query, `resolveTagConfig`/`applyElementTags`, and the `<TagPicker>` combobox (gating, vocabulary, create).

### Screenshots / screencasts

**Tag combobox** — inline chips + a tag button per row (no section header); the popover searches the vocabulary, toggles tags, and offers **Create "…"**:

_Screenshot to be attached: `tag-combobox.png`._

**Color popover** — the paint-roller now opens Fill / Line / Text in a popover instead of expanding the panel inline:

_Screenshot to be attached: `color-popover.png`._

**Coloring by build result** — one `getElementIdsByTag` lookup per test case, fed to `setElementHighlights`:

_Screenshot to be attached: `tag-coloring.png`._

### Checklist

- [x] Linked to a related issue (if applicable) — closes #50
- [x] Tests added or updated
- [x] Documentation updated (if applicable) — new [Element tags & group coloring](https://ls1intum.github.io/Apollon/library/api/element-tags) guide, plus the `ApollonEditor` reference and the model contract
- [ ] Screenshots or screencasts attached (if a UI change) — captured, pending upload (see above)

---

**Note for reviewers on test coverage:** the unit suite (1334 tests) and the Storybook interaction + a11y tests for the new stories run green locally; the Playwright spec (`element-tags.spec.ts`) passes in CI. One pre-existing, unrelated Storybook story test (`ClassDiagram → Edit: Class`) fails on `getByText("Class")` matching both the popover title and the class-type select — it fails identically before this branch and is not run in CI, so it's left for a separate fix rather than bundled here.


library/lib/utils/tagUtils.ts
```diff
@@ -0,0 +1,166 @@
...
+  enabled: false,
+  available: [],
+  allowCreate: false,
+}
+
+/** Normalize the `tags` option (absent / boolean / object) to a `TagConfig`. */
+export function resolveTagConfig(input?: boolean | TagOptions): TagConfig {
+  if (!input) return DISABLED_TAG_CONFIG
+  if (input === true) return { enabled: true, available: [], allowCreate: true }
+  const available = normalizeTags(input.available ?? [])
```
Claudia-Anthropica: @FelixTJDietrich [medium] `normalizeTags()` enforces `MAX_TAGS_PER_ELEMENT`, so reusing it here silently removes every host vocabulary entry after the 50th. That limit makes sense for tags stored on one element, but not for the set of tags a host may offer; a diagram with more than 50 test cases cannot select the later entries.

In `library/lib/utils/tagUtils.ts`, `resolveTagConfig()` normalizes `available` with the per-element 50-tag cap. Normalize the vocabulary without applying `MAX_TAGS_PER_ELEMENT`, while retaining trimming, validation, and de-duplication, and add a test with more than 50 available tags.

</details>

library/lib/components/popovers/TagPicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <ul
+                data-slot="tag-picker-list"
+                className="apollon-tag-picker__list"
+                role="listbox"
+                aria-label={t.editTagsFor(subject)}
+              >
+                {options.map((tag) => {
+                  const selected = tags.includes(tag)
+                  return (
+                    <li
```
Claudia-Anthropica: @FelixTJDietrich [medium] These options only handle clicks: the `li` elements are not focusable and the listbox has no active-descendant or arrow-key handling, so keyboard users cannot choose an existing tag. Use native buttons for the choices or implement the complete listbox keyboard pattern, and cover Enter/Space selection in the interaction test.

In `library/lib/components/popovers/TagPicker.tsx`, the vocabulary options are click-only `li` elements with no keyboard behavior. Make every option keyboard-operable using native interactive controls or a complete ARIA listbox implementation, and add a test that selects a tag using the keyboard.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The tag persistence/query path looks solid, but the picker currently drops vocabulary entries after the 50th and its options cannot be operated from the keyboard. Please also attach the three required UI screenshots listed in the PR before this is ready.

library/lib/utils/tagUtils.ts
FelixTJDietrich: Fixed in 5ac1e3ab. `normalizeTags` now takes a `maxCount` (default `MAX_TAGS_PER_ELEMENT`), and `resolveTagConfig` normalizes `available` with `Infinity`, so the vocabulary keeps trimming/validation/de-duplication but no count cap. Added a unit test covering a 50+ vocabulary (`does not cap the vocabulary at the per-element tag limit`).

library/lib/components/popovers/TagPicker.tsx
FelixTJDietrich: Fixed in 5ac1e3ab. Each vocabulary option is now a native `<button aria-pressed>` inside a `role="group"`, so options are Tab-focusable and toggle on Enter/Space with no custom key handling. The Storybook interaction test now selects a tag by keyboard (focus + `{Enter}`), and the unit test asserts the options are buttons carrying `aria-pressed`.

library/lib/utils/tagUtils.ts
Claudia-Anthropica: @FelixTJDietrich Nice fix — the vocabulary keeps its full size while preserving the existing normalization rules.

library/lib/components/popovers/TagPicker.tsx
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. Native toggle buttons cover keyboard operation cleanly.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Both code issues are fixed and the threads are resolved. The three required UI screenshots are still placeholders in the PR description, so please attach them before this is ready.

PR #817, opened 2026-07-17
feat: save with Ctrl+S, duplicate with Ctrl+D, and a keyboard-driven viewport

### LINKED ISSUE #301: More Editor Shortcuts (CTRL + S)
### Is your feature request related to a problem?
It would be pretty nice to have certain shortcuts at hand: For example if working on a diagram one might want to save it to JSON. Create a shortcut like CTRL + S or something else to save as JSON.
Right now I always have to click on File -> Export -> As JSON, which gets pretty annoying if one does this quite often.
There are already shortcuts present for Undo and Redo, so I guess it wouldn't be too much work to create shortcuts for some other often used functionality as well.

### Describe the solution you'd like
Create some more shortcuts.
CTRL + (sth) -> Export as JSON
The next step would be a settings panel to edit own shortcuts, but that I think would be overkill for Apollon since there are not that many things that would be nice to have shortcuts for I think.

### Describe alternatives you've considered
I haven't considered any. I guess if Apollon-Devs are completely against creating another shortcut, maybe it's too difficult with current project setup, it would be possible to create a Tampermonkey-script or something that click the button on shortcut? CTRL + S in app is cleaner though I believe.

### Additional context
Pretty self-explanatory

Thank you very much for consideration and creation of such a nice tool!

FelixTJDietrich:
### Summary

Closes #301, which asked for `Ctrl+S` to export JSON instead of walking File → Export → As JSON every time. That shortcut is here — and since it meant touching the shortcut layer anyway, the rest of it grew up too.

`Ctrl/Cmd+D` now **duplicates** the selection. It used to clear it, which is what `Esc` is for, and what every other canvas tool binds `Ctrl+D` to. This is the one change that will surprise an existing user, so it is called out in the release note.

The keys the editor consumes now live in **one exported registry**. The handler executes it; the "How to use this editor?" sheet, the File menu hint and every `aria-keyshortcuts` attribute read from it. The sheet a user reads can no longer drift from what the editor actually does — a `Record<ApollonShortcutId, …>` in the webapp makes an uncovered shortcut a **build error**, not a stale doc.

| | |
| --- | --- |
| `Ctrl/Cmd+S` | Save as JSON *(#301)* |
| `Ctrl/Cmd+D` | Duplicate the selection |
| `Ctrl/Cmd+=` · `Ctrl/Cmd+-` · `Ctrl/Cmd+0` | Zoom in · out · 100% |
| `Ctrl/Cmd+Shift+1` · `Ctrl/Cmd+Shift+2` | Fit the diagram · fit the selection |

### Release note

Two changesets, one per audience, per [the style guide](https://ls1intum.github.io/Apollon/contributor/development/release-notes/):

- `@tumaet/apollon` (minor) — duplicate + the viewport combos, the read-only/dialog/IME/auto-repeat rules, and the `APOLLON_SHORTCUTS` surface for embedders.
- `@tumaet/webapp` (minor) — `Ctrl/Cmd+S` saves as JSON; the How-to-use sheet prints the keys your own keyboard has.

- `@tumaet/webapp` (minor) — drag a `.json` diagram anywhere onto the app to import it as a new diagram.

### Implementation notes

**React Flow keeps sole ownership of `Delete` and the arrow keys.** The old handler ran its own `Delete` branch *alongside* React Flow's `deleteKeyCode` — two deletion paths racing on one keypress. The registry lists those keys as `canvasHandled` so an embedder still sees every key the editor consumes, and a discriminated union makes wiring an action for one a **compile error** rather than a bug.

**Zoom-to-fit takes a `Mod` where Figma, Excalidraw and tldraw use a bare `Shift+1`.** A shortcut whose keys produce a printable character (`Shift+1` → `!`) fails [WCAG 2.1.4 (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/character-key-shortcuts) unless it can be turned off, remapped, or scoped to focus — speech input dictating `!` would jump the viewport. Given Apollon runs exams in Artemis under BITV 2.0 / EN 301 549, compliance beat parity. A test over the registry enforces the rule, so the next person to reach for `Shift+1` gets a red build and this paragraph.

**`key` vs `code`, chosen per shortcut rather than by habit.** `key` is layout-aware, so `Mod+Z` follows the key the user's layout prints. `code` is physical, so digits survive a layout that moves them. Getting this backwards is invisible on a QWERTY laptop and fatal elsewhere: French AZERTY prints `à` where QWERTY prints `0`, and Dvorak puts `S` on QWERTY's `;`. Both directions have tests.

**One `document` listener, registered once.** The actions close over `nodes`/`edges`, so a naive dependency array tore the listener down and re-added it on every drag frame (~60 Hz). A latest-value ref keeps it stable; a test asserts exactly one registration survives a re-render, and that unmount removes it.

**Embedders get an off switch.** The editor claims 13 combos on `document`, so `keyboardShortcuts: false` hands them all back — needed by a host that binds those keys itself, or mounts more than one editor.

**`Ctrl/Cmd+Shift+S` now actually saves a version.** It used to only open the history panel, leaving the user to click Save. The shortcut can't reach the editor model or the dirty-check, so it bumps a per-diagram nonce the mounted panel watches and runs its *existing* save path against — reusing every guard (skips an unchanged diagram, an empty one, or a preview) rather than duplicating them.

**Every advertised shortcut is now E2E-tested, so the sheet can't lie.** `keyboard-shortcuts.spec.ts` drives each row the "How to use" dialog shows — select-all, Shift+click add/remove, Shift+drag box-select, delete, copy/cut/paste, duplicate (edges included), arrow-nudge, undo/redo, the full zoom family, zoom-to-selection, pan — against a real editor; the version shortcuts (Ctrl/Cmd+Shift+S, Alt+Shift+H) are covered in `local-version-history.spec.ts`. Doing this surfaced one row that was already lying: it claimed "Scroll = zoom", but `panOnScroll` makes a plain wheel **pan** and only Ctrl/Cmd+scroll zoom — copy corrected and both pinned with tests.

**Bugs found while reviewing, fixed here.** Ctrl/Cmd-clicking two connected classes and duplicating them dropped the edge between them. Pasting a container nudged its children 20 px further inside it every time. Copy claimed `Ctrl+C` even with nothing selected, swallowing a native text copy.

**Deleted along the way:** `parentChildRelations` and `timestamp` (written into every clipboard payload, read by nothing), `deleteSelectedElements` (orphaned once React Flow took over `Delete`), and the module-level `navigator` sniff in the old version-shortcut hook.

**Not done, deliberately:** `duplicate` and `zoom-to-selection` have no pointer affordance — a node-toolbar button would churn the How-to-use screenshots, which want a follow-up PR that can regenerate them. Both are keyboard-reachable and listed in the sheet today.

**Drag-and-drop JSON import — the whole window is the drop target.** Drop a `.json` exported from Apollon anywhere on the app and it opens as a *new* diagram (never overwriting the current one) — the same result as File → Import, now reachable by drag. The first cut wrapped each surface in a dropzone element, and it did not work: a browser opens a dropped file unless **both** `dragover` and `drop` are cancelled on the path to the document, so every gap a wrapper left — page margins, the version drawer, anything outside page content — let the browser navigate away to the raw JSON. Per [MDN's file drag-and-drop guidance](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop), `DiagramFileDropzone` now registers its listeners on `window`, is mounted **once** in the root route, and portals a fixed "Drop to import" overlay while a file is over the app (drag-depth counter so it doesn't flicker across children). It ignores non-file drags, so the palette's pointer drag and ordinary text/link drags keep their native behaviour. The read-and-create logic that File → Import and the home button each duplicated now lives in one `useImportDiagramFile` hook the dropzone shares, so all three paths migrate old formats, create the new diagram, and report success/failure identically.

### Steps for testing

1. Open any diagram, select a node, press <kbd>Ctrl/Cmd</kbd>+<kbd>D</kbd> → an offset copy appears, selected and ready to move.
2. <kbd>Ctrl/Cmd</kbd>-click two classes that have an association, then <kbd>Ctrl/Cmd</kbd>+<kbd>D</kbd> → **the association comes with them** (it used to vanish).
3. <kbd>Ctrl/Cmd</kbd>+<kbd>S</kbd> → a `.json` download and a toast; the browser's save-page dialog never appears. It works while renaming the diagram, too — that's intentional, and matches [excalidraw#9281](https://github.com/excalidraw/excalidraw/issues/9281).
4. <kbd>Ctrl/Cmd</kbd>+<kbd>Shift</kbd>+<kbd>S</kbd> → the version panel opens and a version lands in the history right away (no composer). Press it again with no edits → a quiet "No changes to save".
5. <kbd>Ctrl/Cmd</kbd>+<kbd>-</kbd> then <kbd>Ctrl/Cmd</kbd>+<kbd>0</kbd> → 100%. <kbd>Ctrl/Cmd</kbd>+<kbd>Shift</kbd>+<kbd>1</kbd> frames the whole diagram.
6. Hold <kbd>Ctrl/Cmd</kbd>+<kbd>D</kbd> → exactly one duplicate. Hold <kbd>Ctrl/Cmd</kbd>+<kbd>-</kbd> → zoom keeps stepping.
7. Open the PPTX export dialog and press <kbd>Ctrl/Cmd</kbd>+<kbd>A</kbd> → the diagram behind it must not select. Type in the diagram title and press <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>H</kbd> → the version drawer must not open under you.
9. Drag a `.json` diagram (export one first with Ctrl/Cmd+S) onto the app — over the canvas, the home gallery, or the margin **outside** any page content → a "Drop to import" overlay appears, and releasing opens it as a new diagram. The browser must never navigate to the raw JSON. Drop a non-JSON file → it's declined with a toast.
10. Help → **How to use this editor?** → Shortcuts. On a Mac: `⌘D`, `⇧⌘Z`, `⌫`, no `+` separators. Elsewhere: `Ctrl+D`, `Ctrl+Shift+Z`, `Delete`. Open the same dialog from the home page → the File group is gone, because those shortcuts aren't mounted there.

### Screenshots / screencasts

**Not attached — I couldn't produce them, and I'd rather say so than skip the row.** The shortcut sheet is covered by four Storybook stories instead (`HowToUseModal.stories.tsx`): `Shortcuts`, `ShortcutsOnMac` (⌘⇧⌥⌫ with Apple's modifier order), `ShortcutsOnHome` (asserts the File group is absent), and `Default`. A reviewer can see every state with `pnpm storybook`. Happy to attach captures if you'd like them in the PR itself.

### Checklist

- [x] Linked to a related issue (if applicable) — closes #301
- [x] Added a changeset whose summary is written in the user's voice (`pnpm changeset`, [how](https://ls1intum.github.io/Apollon/contributor/development/release-notes/)) — two, one per track
- [x] Tests added or updated
- [x] Documentation updated (if applicable) — `docs/library/api.md` gains a Keyboard shortcuts section; its example is type-checked by `check-doc-snippets`
- [ ] Screenshots or screencasts attached (if a UI change) — **not attached**, see above

---

<details>
<summary>What I ran locally, and the one thing I couldn't</summary>

Green: `pnpm run build` · library 1304 tests · webapp 274 tests · `tsc --noEmit` in both packages · `pnpm knip` · `pnpm run lint` (0 errors, and **no new warnings** — the 68/55 remaining are pre-existing in untouched files) · `pnpm run format:check` · `pnpm run check:doc-snippets` · `sync-library-version --check` · `check-release-taxonomy` · `check-no-perf-hooks` · webapp size budget (1.13/1.3 MB, CSS 35.06/36 kB) · the MUI/Emotion guard.

**Not run:** the new `tests/e2e/keyboard-shortcuts.spec.ts` — Playwright can't boot the standalone workspace in my sandbox. It's the one claim in this PR I can't back myself; please let the PR Health Gate settle before merging.

</details>


library/lib/keyboard.ts
```diff
@@ -0,0 +1,289 @@
...
+  // Mid-composition, an IME is still spelling a character out of these keys.
+  if (event.isComposing) return
+  if (isTypingTarget(event)) return
+  if (isInsideOverlay(event)) return
+
+  for (const shortcut of APOLLON_SHORTCUTS) {
+    if (!shortcut.combos.some((combo) => matchesShortcutCombo(event, combo))) {
+      continue
+    }
+    if (shortcut.canvasHandled) return
```
Claudia-Anthropica, line 284: @FelixTJDietrich [medium] `canvasHandled` returns before any action, but React Flow’s Delete and arrow handlers are independent of this hook. As a result, `keyboardShortcuts: false` still leaves those keys active, and Delete from a dialog’s non-input control can still remove the canvas selection despite the guards above; the React `<Apollon>` wrapper also never accepts or forwards the new option. Wire the enabled state and overlay suppression through the React Flow-owned shortcuts and expose the option on `ApollonProps`.

In `library/lib/keyboard.ts`, React Flow-owned shortcuts bypass the new disable and overlay contract, and the React wrapper does not expose the option. Wire `keyboardShortcuts` through `ApollonProps`, make React Flow deletion and arrow movement honor it, prevent canvas-handled shortcuts originating in overlays, and add integration tests for disabled shortcuts and Delete inside a dialog.

</details>

standalone/webapp/src/components/versioning/VersionDrawer.tsx
```diff
@@ -165,17 +169,20 @@ export const VersionSidebarBody: FC<Props> = ({
     if (!latestSavedVersion) {
       setSavedFingerprint(null)
       setHasChanges(true)
+      setBaselineResolved(true)
```
Claudia-Anthropica: @FelixTJDietrich [medium] `!latestSavedVersion` also describes the window before the initial version list has finished loading. Because both editor pages fire `fetchVersions()` without awaiting it, an immediate Ctrl/Cmd+Shift+S can mount this panel with an empty `versions` array, mark the baseline resolved, and create a duplicate before the existing latest version arrives. Keep the request pending until the initial list load has completed, then distinguish that state from a genuinely empty history.

In `standalone/webapp/src/components/versioning/VersionDrawer.tsx`, an empty `versions` array is treated as a resolved empty history even while the initial fetch is pending. Gate `baselineResolved` on completion of the initial version-list load, preserve the shortcut request while loading, and add a test where an existing version arrives after the shortcut is pressed.

</details>

library/lib/hooks/useKeyboardShortcuts.ts
```diff
@@ -1,151 +1,125 @@
...
+      },
+      cut: () => {
+        if (!hasSelectedElements()) return false
+        pasteCountRef.current = 0
+        void cutSelectedElements()
+      },
+      // Each paste of the same clipboard steps further from the original; a
+      // paste that found nothing to insert doesn't consume a step.
+      paste: () => {
+        const step = pasteCountRef.current + 1
```
Claudia-Anthropica: @FelixTJDietrich [medium] The paste slot is reserved only after the asynchronous clipboard read completes. Two quick Ctrl/Cmd+V presses can therefore both choose step 1, and both callbacks close over the same pre-paste node list, so the later insertion can overwrite the earlier one instead of producing two cascaded copies. Serialize paste operations or reserve the step before awaiting, and apply each insertion against the latest store state.

In `library/lib/hooks/useKeyboardShortcuts.ts`, concurrent paste requests reuse the same offset and stale diagram state. Serialize paste requests or reserve their offsets synchronously, insert each result against the latest store state, safely release failed reservations, and add a delayed-clipboard test for two rapid paste presses.

</details>

standalone/webapp/src/utils/platform.ts
```diff
@@ -26,3 +26,13 @@ export const isIOS = (): boolean => test(/iPhone|iPod/i) || isIpad()
...
+
+// --- Apollon's own, not vendored from Ionic ---
+
+/**
+ * Whether to render shortcuts with macOS key symbols. Display-only — no handler
+ * branches on it (they take Ctrl or Cmd either way), so a wrong guess costs a
+ * glyph, not a shortcut. iPadOS 13+ reports a desktop `Macintosh` UA, which is
+ * what we want here: an iPad with a keyboard has a ⌘ key.
+ */
+export const isMacLike = (): boolean => test(/mac|iphone|ipod/i)
```
Claudia-Anthropica: @FelixTJDietrich [low] This check omits the legacy `iPad` user agent even though `isIpad()` immediately above already detects it. Those iPads with a hardware keyboard will be shown Ctrl shortcuts instead of the Command-key caps they actually have. Reuse `isIOS()` here or include `ipad` explicitly.

In `standalone/webapp/src/utils/platform.ts`, `isMacLike()` fails for user agents containing `iPad`. Make it reuse the existing iOS/iPad detection or include iPad explicitly, and add coverage for both legacy iPad and desktop-mode iPad user agents.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The registry-driven approach and test coverage are strong, but the off-switch and overlay contract do not yet cover React Flow-owned keys, and the version/paste paths each have an async race. I left one smaller iPad key-cap issue inline as well. Please address these before merging.

library/lib/keyboard.ts
FelixTJDietrich: Fixed in 3f11f663. `keyboardShortcuts: false` now also gates React Flow: `deleteKeyCode={keyboardShortcuts ? [...] : []}` and `disableKeyboardA11y={!keyboardShortcuts}` in `App.tsx`, so Delete and arrow-key moving stand down with the rest. `onBeforeDelete` (the one gate every RF deletion funnels through) returns `false` when `document.activeElement` is inside an overlay over the canvas, so Delete from a dialog no longer removes the selection behind it. The option is now on `ApollonProps` → `editor.setKeyboardShortcuts()`. Tests: `isElementInOverlay` + `onBeforeDelete` (dialog blocks, canvas allows) in the library unit suite.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
FelixTJDietrich: Fixed. The quick-save now waits for the initial list load before deciding: it gates on `s.versions[id] !== undefined || s.error[id] != null` (the store writes `versions[id]` only after `fetchVersions` resolves), so a briefly-empty list on mount is no longer read as "no history". The request stays pending until then. New e2e in `local-version-history.spec.ts`: press Ctrl/Cmd+Shift+S right after a reload on an unchanged diagram → still exactly one version, "No changes to save".

library/lib/hooks/useKeyboardShortcuts.ts
FelixTJDietrich: Fixed. The step is now reserved synchronously (`pasteCountRef.current += 1` before the await) and the burst is serialized via a promise chain. The root cause — insertion against stale state — is fixed in `useSelectionForCopyPaste`: `insertClipboardData` and `duplicateSelectedElements` read live state via a new `useDiagramStoreApi().getState()` at insertion time, so a later paste builds on the earlier one rather than overwriting it. A failed paste leaves its slot unused (a harmless gap the next copy resets). New e2e: two distinct rapid pastes → two cascaded copies (this reproduced the overwrite on the old code — 3 nodes instead of 4).

standalone/webapp/src/utils/platform.ts
FelixTJDietrich: Fixed. `isMacLike()` is now `isIOS() || test(/mac/i)`, so it reuses the existing iPad detection (legacy `iPad` UA and the iPadOS 13+ `Macintosh`+coarse-pointer form). Added coverage in `platform.test.ts` for legacy iPad, desktop-mode iPad, iPhone, Mac, Windows, and Android.

library/lib/keyboard.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks — the option reaches the React wrapper and React Flow's delete/a11y paths, and the overlay guard covers the independent delete handler.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
Claudia-Anthropica: @FelixTJDietrich This still has an effect-order race when the list first arrives. That render sees `initialListLoaded === true` while `baselineResolved` is still `true` from the earlier empty-list effect; the baseline effect queues `setBaselineResolved(false)`, but the request effect in the same flush still sees the old `true` and can call `handleCreate()` before `getBody()` resolves. Track readiness for the specific latest version (or avoid resolving the empty baseline until the initial list is loaded), and make the test delay the list/body fetch so it actually holds this window open.

library/lib/hooks/useKeyboardShortcuts.ts
Claudia-Anthropica: @FelixTJDietrich Nice fix. Reserving before the await, serializing the reads, and inserting against `getState()` closes both sides of the rapid-paste race.

standalone/webapp/src/utils/platform.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. Reusing `isIOS()` covers both iPad user-agent forms and the added cases pin it down.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Three of the four previous concerns are addressed, but the version quick-save can still process the pending request in the same effect flush where the list arrives, before the fingerprint baseline is reset. Please tie readiness to the resolved baseline for the current latest version and hold that race open in the test; details are in the remaining thread.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
FelixTJDietrich: Fixed in b4b336f8. You were right — the boolean lagged a render. Readiness is now derived from the specific latest-version id rather than a separate boolean: `const baselineResolved = baselineVersionId === (latestSavedVersion?.id ?? null)`. On the frame the list first arrives, `baselineVersionId` is still `null` (empty branch) while `latestSavedVersion.id` is `v1`, so `baselineResolved` is `false` with no stale state — the slow-path `getBody` sets `baselineVersionId` to that id in `.finally`, and only then does it read ready. The `initialListLoaded` gate still distinguishes an unloaded list from a genuinely empty one. New integration test `VersionDrawer.saveShortcut.test.tsx` holds both the `list` and `getBody` fetches open across that frame with deferred promises and asserts `create` is never called for an unchanged diagram — it fails on the old boolean (create called once) and passes now.

Claudia-Anthropica: @FelixTJDietrich Nice fix — tying readiness to the current version id closes the effect-order window, and the deferred list/body test now holds the exact race open.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
```diff
@@ -188,6 +204,7 @@ export const VersionSidebarBody: FC<Props> = ({
         setSavedFingerprint(null)
         setHasChanges(true)
       })
+      .finally(() => setBaselineVersionId(resolvingVersionId))
```
Claudia-Anthropica: @FelixTJDietrich [medium] An older `getBody()` can still resolve after a newer version body and overwrite both `savedFingerprint` and `baselineVersionId`. This happens when a collaborator creates or restores a version while the first baseline fetch is pending; after the stale `.finally` writes the old id, `baselineResolved` stays false for the current latest version and later quick-saves remain queued. Ignore stale effect results and cover reversed completion order.

In `standalone/webapp/src/components/versioning/VersionDrawer.tsx`, overlapping baseline `getBody()` calls can resolve out of order and let an older request overwrite the fingerprint and `baselineVersionId` for the current latest version. Add an effect cleanup or generation guard so only the request for the current latest version may update baseline state, and add a regression test that resolves the newer body before the older one and verifies quick-save still completes against the newer baseline.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The original effect-order race is fixed, and the focused regression test passes. One async baseline race remains when the latest version changes while an older getBody() request is still pending; see inline.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
FelixTJDietrich: Good catch — fixed in 467a3f43.

The `getBody()` effect now closes over a `let stale = false` flag, and its cleanup sets `stale = true`. Because the effect is keyed on `latestSavedVersion.id`, a newer version arriving tears down the older run first, so a late-resolving `getBody()` can no longer write either `savedFingerprint` or `baselineVersionId`:

```ts
let stale = false
const resolvingVersionId = latestSavedVersion.id
getVersionRepository()
  .getBody(latestSavedVersion.diagramId, resolvingVersionId)
  .then((body) => { if (!stale) setSavedFingerprint(structuralFingerprint(body)) })
  .catch(() => { if (!stale) { setSavedFingerprint(null); setHasChanges(true) } })
  .finally(() => { if (!stale) setBaselineVersionId(resolvingVersionId) })
return () => { stale = true }
```

Covered by a regression test — `VersionDrawer.saveShortcut.test.tsx` → "ignores a stale body that resolves after a newer version's". It resolves the two `getBody()` promises out of order and asserts the quick-save still fires against the newer body. Removing the `stale` guard fails it (`expected vi.fn() to be called once, but got 0 times`).

leonsmnts:
🎉

PR #818, opened 2026-07-17
fix(library): stop resize cursors on borders a node can't resize

### LINKED ISSUE #629: [BUG] False affordance - Resize cursor appears on non-vertically resizable elements
### Describe the bug

Hovering over the top or bottom edge of a Class element changes the cursor to a vertical resize indicator (e.g., ns-resize). Since Class height is dynamically determined by its content (attributes/methods), vertical resizing is disabled. Showing the resize cursor creates a false affordance, confusing users who try to drag it.

### To Reproduce

Open Apollon (Class Diagram).

Add a Class element to the canvas.

Select the element.

Hover the mouse cursor directly over the top or bottom horizontal boundary.

### Expected behavior

Cursor remains the default pointer or move/grab icon.

### Screenshots

https://github.com/user-attachments/assets/a9f589f5-e403-46ac-94c9-c72a4c53ffd2

### OS

_No response_

### Browser

_No response_

### Version

_No response_

### Additional context

_No response_

### Web Console Output

```shell

```

FelixTJDietrich:
### Summary

Hovering the top or bottom border of a class showed a vertical-resize cursor (`ns-resize`) and let you start a drag — but a class's height is fixed by its attributes and methods, so the drag did nothing. Classic false affordance (#629): the pointer promises something the node can't deliver.

The root cause turned out to be a bit of a trap. React Flow's `<NodeResizer>` **always** renders all four edge lines plus four corner handles, and its own stylesheet paints each one with a resize cursor. A previous attempt fixed this with a CSS `cursor` override keyed on a `horizontally-not-resizable` class — but that class was put on the node only through `FeedbackDropzone`, which drops its `className` in every mode except assessment. Resize controls only render while modelling, so the override **never reached the DOM where it mattered**. It was dead from the start, which is why the issue stayed open.

The fix removes the controls that shouldn't be there and keeps the ones that should. A content-sized node no longer draws the edge lines of its pinned axis, **but keeps its four corner handles**, each constrained to the axis that can move with its cursor relabelled to match.

While auditing every node's resize behaviour for this, I found a second, older bug (thanks to the reviewer who spotted it on swimlanes): on any *filled* node — a swimlane, pool, package — the edge resize lines were invisible and un-grabbable, leaving only the corners. React Flow renders the resize controls before the node's content in the DOM, and its 1px edge lines carry no z-index, so the node body painted straight over them. A rule that was meant to fix this long ago never matched (`.react-flow__resize-control line` targets an SVG `<line>`; the controls are `<div>`s). This PR lifts the lines above the content and gives every line a comfortable grab area.

### Release note

Resize handles behave consistently across nodes: borders that can't resize no longer show a resize cursor, content-sized nodes keep their corner handles (resizing only the axis that can change), and on filled nodes like activity swimlanes the resize edge is grabbable instead of hidden behind the node.

### Implementation notes

- New `NodeResizer` wrapper (`library/lib/nodes/wrappers/NodeResizer.tsx`) around React Flow's. On a pinned axis (`min >= max`) it drops that axis's two edge lines and keeps the four corner handles, constraining each to the free axis with React Flow's **`resizeDirection`**; both axes free delegates to React Flow; both pinned renders nothing.
- A pinned node's corners get a marker class (`apollon-resize-corner--x` / `--y`) so app.css can relabel React Flow's hard-coded diagonal corner cursor to the free axis's cursor. The override wins on specificity (four classes vs. three), no `!important`.
- **Edge-line occlusion fix:** `.react-flow__resize-control.line` now gets `z-index: 9` (above the node content, below the corner handles at 20 and connection arcs at 10), and every line gets a widened, hover-gated hit area. This restores draggable edges on all filled nodes.
- **Audit result:** all ~40 resizable node types route through the wrapper; the 12 fixed-geometry shapes (initial/final nodes, BPMN events & gateways, interfaces, Petri place, SFC start/jump/transition) render no resizer, which is intended. The six content-sized nodes (class, object, communication-object, SFC action table, both activity fork bars) are the only ones with a pinned axis.
- A `no-restricted-imports` rule bans importing `NodeResizer` from `@xyflow/react`, so a future node can't silently pick up the un-fixed one; the wrapper is the single sanctioned importer. The `handleStyle={{ width: 8, height: 8 }}` that was copy-pasted across ~30 files is now one default in the wrapper.
- Also removed: the dead cursor-override CSS, its two now-unused class names, and the now-unused `className` prop that threaded them through `DefaultNodeWrapper` → `FeedbackDropzone`.

### Steps for testing

1. Open a class diagram, add a **Class**, select it. Hover the **top/bottom** border → default/grab cursor, no dead drag (was `ns-resize`). Hover a **corner** → `ew-resize`, dragging resizes width; height stays content-driven.
2. Add a **Package** (free both axes) → all four edges and corners resize normally.
3. Open an **activity diagram**, add a **Swimlane**, select it, hover an outer edge → the edge is grabbable and resizes the swimlane (previously only the corners worked).
4. Automated:
   - `cd library && pnpm test` — unit specs assert which controls render per axis, the free-axis corner cursor, and the edge-line hit-area class.
   - `cd standalone/webapp && pnpm exec playwright test --project=chromium resize-cursor-locked-axis resize-edge-reachable` — real-browser checks of the resolved cursor and that a swimlane's top edge is the top-most (grabbable) element. All were mutation-checked — each fails if its fix is reverted.

### Screenshots / screencasts

The change is cursor/affordance behaviour, best seen live — the repro screencast on #629 shows the broken state, and the Playwright specs assert the real resolved cursor and edge reachability in CI.

Note on the "How to use" modal images: two render a *selected* class, so they show its resize controls, which this PR changes slightly (the two top/bottom edge lines go). I ran the repo's **Update Visual Baselines** workflow and it correctly committed nothing — the diff is under the project's `maxDiffPixelRatio: 0.01`, so `--update-snapshots` treats the images as unchanged. No repo tooling will refresh them without forcing `--update-snapshots=all`; flagging transparently, happy to force-refresh in this PR if you'd prefer.

### Checklist

- [x] Linked to a related issue (if applicable) — closes #629
- [x] Tests added or updated — unit + e2e, all mutation-checked
- [ ] Documentation updated (if applicable) — n/a
- [x] Screenshots or screencasts attached (if a UI change) — see note above

### DIFF
--- a/library/lib/nodes/wrappers/NodeResizer.tsx
+++ b/library/lib/nodes/wrappers/NodeResizer.tsx
@@ -1,0 +1,101 @@
+// This module is the one sanctioned importer of React Flow's resizer; every
+// other call site is pointed here by `no-restricted-imports`. The shadowed name
+// is deliberate: nodes should not have to know an axis-aware variant exists, and
+// the lint rule makes reaching past it a build error rather than a convention.
+import {
+  // eslint-disable-next-line no-restricted-imports
+  NodeResizer as ReactFlowNodeResizer,
+  NodeResizeControl,
+  ResizeControlVariant,
+  type NodeResizerProps,
+} from "@xyflow/react"
+
+// React Flow's corner handles default to 5x5; every Apollon node wants 8x8.
+const HANDLE_STYLE = { width: 8, height: 8 }
+
+// A node pins a content-sized dimension by bounding it from both sides — a
+// Class's height is driven by its attribute/method rows. Bounds that cross leave
+// no range to drag either, so they count as pinned; a side left unbounded never
+// does.
+const isAxisLocked = (min?: number, max?: number): boolean =>
+  min !== undefined && max !== undefined && min >= max
+
+/**
+ * React Flow's `<NodeResizer>` always renders four edge lines and four corner
+ * handles, so a pinned axis still shows a resize cursor and accepts a drag that
+ * does nothing. Rendering only the controls that can move is the fix; the cursor
+ * is a signifier, and repainting it would leave the dead control behind.
+ *
+ * `resizeDirection` is not an alternative: it sits on `NodeResizeControl`, and
+ * `ResizeControlLineProps` omits it outright. It constrains a corner's drag
+ * while leaving the diagonal cursor that lies about it, so corners go rather
+ * than get constrained.
+ */
+export function NodeResizer(props: NodeResizerProps) {
+  const {
+    isVisible = true,
+    minWidth,
+    minHeight,
+    maxWidth,
+    maxHeight,
+    nodeId,
+    color,
+    keepAspectRatio,
+    autoScale,
+    shouldResize,
+    onResizeStart,
+    onResize,
+    onResizeEnd,
+    lineStyle,
+    lineClassName,
+  } = props
+
+  // `isVisible` exists only on NodeResizerProps — NodeResizeControl has no such
+  // prop, so this early return is the only thing enforcing it on the locked path.
+  if (!isVisible) return null
+
+  const widthLocked = isAxisLocked(minWidth, maxWidth)
+  const heightLocked = isAxisLocked(minHeight, maxHeight)
+
+  if (!widthLocked && !heightLocked) {
+    return <ReactFlowNodeResizer handleStyle={HANDLE_STYLE} {...props} />
+  }
+
+  // Without this, a fully pinned node reads as height-locked below and gets the
+  // side controls it equally can't honour.
+  if (widthLocked && heightLocked) return null
+
+  const positions = heightLocked
+    ? (["left", "right"] as const)
+    : (["top", "bottom"] as const)
+
+  // `keepAspectRatio` is the one prop that cannot mean anything here — it would
+  // ask the pinned axis to follow the free one. No node passes it.
+  return (
+    <>
+      {positions.map((position) => (
+        <NodeResizeControl
+          key={position}
+          position={position}
+          variant={ResizeControlVariant.Line}
+          nodeId={nodeId}
+          color={color}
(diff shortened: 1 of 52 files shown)


library/lib/nodes/wrappers/NodeResizer.tsx
```diff
@@ -0,0 +1,117 @@
...
+ */
+export function NodeResizer(props: NodeResizerProps) {
+  const {
+    isVisible = true,
+    minWidth,
+    minHeight,
+    maxWidth,
+    maxHeight,
+    handleStyle,
+    handleClassName: _handleClassName,
```
Claudia-Anthropica: @FelixTJDietrich [low] The wrapper accepts React Flow's full `NodeResizerProps`, but on the locked-axis path it silently drops `handleClassName`. Combine it with `cornerClass` so custom handle styling behaves the same whether an axis is locked or free.

In `library/lib/nodes/wrappers/NodeResizer.tsx`, the locked-axis path destructures and discards `handleClassName`. Preserve it and combine it with the required axis marker class on each corner handle, then add a unit test covering the forwarded class.

</details>

Claudia-Anthropica (approved):
@FelixTJDietrich The axis-aware wrapper and its coverage look solid, and the edge hit-area fix is nicely scoped. I left one small prop-forwarding issue inline; approving since it does not affect any current caller.

library/lib/nodes/wrappers/NodeResizer.tsx
FelixTJDietrich: Good catch — fixed in 89f2bdf4. The locked path now merges `handleClassName` with the axis marker on each corner, matching the free path, and a unit test covers the forwarded class.

PR #821, opened 2026-07-17
refactor(webapp): move version history and diagram loading to TanStack Query

### LINKED ISSUE #687: Adopt TanStack Query as the webapp's HTTP data layer
## What ships

Replace every direct `fetch`/`DiagramApiClient`/`VersionApiClient` call inside React with `useQuery` / `useMutation`. Retire three classes of bespoke scaffolding the bare `fetch` style forced on us: ad-hoc `AbortController` plumbing through optional `opts.signal` params, hand-rolled `cancelled` flags inside effects, and the "reset `isLoading` at the top of the effect" workaround that #ee854a31 introduced because the same component instance is reused across diagrams.

**Dependency** (one package, no peer deps beyond React 18 which we already have)
- Add `@tanstack/react-query` `^5.x` to `standalone/webapp/package.json`. Pin to exact per the repo's pinning rule (#9ed33d59).
- Add `@tanstack/react-query-devtools` as a `devDependencies` entry; mount conditionally on `import.meta.env.DEV`.

**Provider wiring**
- New `standalone/webapp/src/queryClient.ts` exporting a singleton `QueryClient` configured with:
  - `defaultOptions.queries.staleTime: 30_000` (global default; per-query overrides below)
  - `defaultOptions.queries.gcTime: 5 * 60_000`
  - `defaultOptions.queries.retry: (failureCount, err) => !(err instanceof ApiError && err.status >= 400 && err.status < 500) && failureCount < 2` — never retry 4xx, retry up to 2× on transport/5xx
  - `defaultOptions.queries.refetchOnWindowFocus: false` — Yjs+WS already drives liveness; window-focus refetch would clobber editor state mid-edit
  - `defaultOptions.mutations.retry: 0`
- Mount `<QueryClientProvider client={queryClient}>` in `AppProviders.tsx` outside `<EditorProvider>` (Query has no opinion on editor lifecycle; editor depends on Query).

**Query catalogue** (every key + its options + its owner)

| Key | Owner hook | `staleTime` | `gcTime` | Notes |
|---|---|---|---|---|
| `['diagram', diagramId]` | `useDiagram(id)` | `Infinity` | `0` | Body is one-shot seed; Yjs owns post-mount truth. `gcTime: 0` so cache is dropped on unmount — no risk of re-using stale body on remount of a re-joined room. |
| `['diagram', diagramId, 'versions', { limit }]` | `useVersionsInfinite(id)` | `30_000` | `5 min` | `useInfiniteQuery` with `getNextPageParam` reading server's `nextCursor`. |
| `['diagram', diagramId, 'version', versionId, 'body']` | `useVersionBody(id, versionId, { enabled })` | `Infinity` | `5 min` | Version bodies are immutable; never refetch. `enabled` is the visibility gate (IntersectionObserver in `VersionThumbnail`). |

**Mutation catalogue** + invalidation map

| Hook | Server call | Effect on cache |
|---|---|---|
| `useAutosave(id)` | `PUT /api/diagrams/:id` w/ `If-Match` | On success: `queryClient.setQueryData(['diagram', id, 'meta'], { headRev })`; on `REVISION_MISMATCH`, rebase from `err.meta.currentHeadRev` exactly as today — no cache invalidation (Yjs owns body) |
| `useCreateDiagram()` | `POST /api/diagrams` | None (new entity; consumer navigates) |
| `useDeleteDiagram()` | `DELETE /api/diagrams/:id` | `queryClient.removeQueries({ queryKey: ['diagram', id] })` |
| `useCreateVersion(id)` | `POST .../versions` | Optimistic prepend into `['diagram', id, 'versions', ...]`, plus `invalidateQueries({ queryKey: ['diagram', id, 'versions'] })` on settle |
| `useRestoreVersion(id)` | `POST .../versions/:vid/restore` | Invalidate both `['diagram', id]` and `['diagram', id, 'versions']` (server creates an auto-snapshot) |
| `useRenameVersion(id)` | `PATCH .../versions/:vid` | `setQueryData` over the matching versions page entry; no refetch |
| `useDeleteVersion(id)` | `DELETE .../versions/:vid` | Optimistic remove from `['diagram', id, 'versions']`; invalidate on settle |

**WebSocket → cache invalidation bridge**
- New `useVersionEventBridge(id)` hook subscribed via `wsManager.onControl`. On each control event:
  - `VERSION_CREATED`: `invalidateQueries({ queryKey: ['diagram', id, 'versions'] })`
  - `VERSION_DELETED`: `invalidateQueries({ queryKey: ['diagram', id, 'versions'] })` + `removeQueries({ queryKey: ['diagram', id, 'version', event.versionId] })`
  - `VERSION_RENAMED`: same cache patch as `useRenameVersion`'s optimistic path
  - `VERSION_RESTORED`: invalidate `['diagram', id, 'versions']`; the body refetch is gated by `useVersionStore.pendingRestoreFromId` (existing flag) to avoid clobbering peer-local intent
- This collapses the inline VERSION_RESTORED branch in `ApollonWithConnection.tsx:191-224` into one line: `useVersionEventBridge(diagramId)`.

**Yjs ↔ Query contract** (the single load-bearing decision)
- `useDiagram(id)` is the **seed** loader. The first successful response is what gets fed into `new ApollonEditor({ model })`. After that point, **Yjs is the source of truth**; the cache entry is intentionally `staleTime: Infinity` and `gcTime: 0` — never refetched while mounted, dropped on unmount.
- The only post-mount writes to `['diagram', id]` are: (a) post-restore refetch via `invalidateQueries` in `useRestoreVersion.onSuccess`, gated by `pendingRestoreFromId`; (b) preview-exit refetch in `ApollonWithConnection.tsx:431` (`restoredDuringPreviewRef` branch). Both are explicit, both are documented.
- Document this contract in a top-of-file comment on `useDiagram.ts` so the next contributor doesn't add a `refetchInterval`.

**Component changes**
- `ApollonWithConnection.tsx`: replace the `setIsLoading` + reset block + manual `AbortController` + `fetchDiagram` call with `const { data: diagram, isPending, error } = useDiagram(diagramId)`. The reset block deletes entirely — Query keys give us per-diagram isolation for free.
- `VersionDrawer.tsx`: replace `useVersionStore.versions` + `fetchVersions` orchestration with `useVersionsInfinite(diagramId)`.
- `VersionThumbnail.tsx:72-99`: replace the `cancelled` flag + `useState`/`useEffect` with `useVersionBody(diagramId, versionId, { enabled: isVisible })`. The hand-rolled cancellation pattern is exactly what `enabled` covers.
- `NewDiagramModal.tsx`, `ShareModal.tsx`, `LoadDiagramModal.tsx`, `NewDiagramFromTemplateModal.tsx`: switch from direct API client calls to `useCreateDiagram` / etc. The `close-before-navigate` ordering stays (route-unmount race is orthogonal to Query); the comment in `NewDiagramModal.tsx:61` is retained.

**`DiagramApiClient` / `VersionApiClient` shape**
- Keep both as **bare async functions** (delete the `opts: { signal?: AbortSignal }` parameters added in this branch — TanStack Query wires `signal` into `queryFn`'s arg automatically). Service layer becomes purely "HTTP transport"; nothing in it knows about React.
- Delete `useVersionStore.fetchVersions` and the `versions`/`nextCursor`/`total`/`isLoading`/`hasLoaded` fields. These were server state masquerading as client state — Query owns them now. Keep `useVersionStore` for **client-only** state: `preview`, `pendingRestoreFromId`, `undoRestore`, `enterPreview`/`exitPreview`. Add ESLint rule (or TODO) that any field touching network data does not land in this store.

**Autosave**
- Replace the `setInterval`-driven autosave at `ApollonWithConnection.tsx:283-316` with a `useAutosave(diagramId)` hook that internally runs a 5 s tick and calls `mutation.mutate` from a ref. The `If-Match`/`REVISION_MISMATCH` rebase logic is preserved verbatim inside `onError`. Mutation lifecycle is automatic: on unmount, in-flight `fetch` aborts via the mutation's `AbortSignal`.
- `useFlushOnUnload.ts` cannot use Query (mutations can't fire on `pagehide`/`beforeunload` with cancellation semantics). Keep it as today — it imports `DiagramApiClient.sendDiagramUpdate` directly and uses `navigator.sendBeacon`-grade timing. Document in the hook's header.

**Tests**
- Existing `ApollonWithConnection.test.tsx`: drop the bespoke `fetchHoisted` mock that simulates `signal.addEventListener('abort', reject, …)`. Replace with a real `QueryClient` whose `queryFn` is mocked at the service-layer boundary (`vi.mock("@/services/DiagramApiClient", …)`), wrapped in `QueryClientProvider` inside `mountAt`. Test count and revert-proofs (Share-again red, unmount-races-fetch red, collab-prompt-each-diagram red) carry over.
- New `useDiagram.test.tsx`, `useVersionsInfinite.test.tsx`, `useAutosave.test.tsx` — service-layer mocked, hook behavior asserted via `renderHook` + `QueryClientProvider`.

## Why

Five concrete problems collapse into one architectural decision:

1. **The "reset block at the top of the effect"** (`ApollonWithConnection.tsx:88-96`) exists *only* because the same component instance is reused across diagrams (route `/:diagramId`). Query keys are per-diagram by construction; the reset block disappears.
2. **The optional `signal?: AbortSignal` parameter on `DiagramApiClient.fetchDiagram`** introduced in #ee854a31 is asymmetric (1 of 4 diagram methods, 0 of 6 version methods). TanStack Query threads `signal` into every `queryFn` for free; the API surface goes back to clean async functions.
3. **`VersionThumbnail.tsx`'s hand-rolled `cancelled` flag** is `useQuery({ enabled })` with extra steps and a race window.
4. **The VERSION_RESTORED branch at `ApollonWithConnection.tsx:191-224`** mixes WS control dispatch, conditional refetch, and a toast. The toast fires unconditionally before `.then` resolves — if unmount races, the toast appears for a diagram the user already left (Loop 2 latent finding, deferred from #ee854a31). The Query bridge moves this into `useVersionEventBridge` + mutation `onSuccess`, both of which are unmount-aware.
5. **`useVersionStore` conflates server state and client state.** `versions[]`, `nextCursor`, `total`, `isLoading`, `hasLoaded` are server cache; `preview`, `pendingRestoreFromId`, `undoRestore` are client intent. Query splits this cleanly; the resulting Zustand store is ~70 % smaller and only holds client state — easier to reason about during preview/restore races.

This is a follow-up to #ee854a31 (the "blank screen on Share-again / New-diagram" fix). That PR resolved the symptom with empirical test coverage; this issue resolves the underlying class.

## Non-goals

- **No Suspense mode.** `useQuery` with `data ?? <Skeleton />` matches the current loading-overlay UX; Suspense + Error Boundary plumbing is a separate decision.
- **No `persistQueryClient`** (IndexedDB / localStorage persistence). Diagrams arrive via WS; cross-tab persistence is the wrong layer.
- **No replacement of `WebSocketManager`.** WS is not request/response; it's a long-lived collaborative pipe. Query is for HTTP. Boundary: `WebSocketManager` dispatches control events; the bridge hook translates those into cache invalidations.
- **No backend changes.** Endpoints, status codes, `If-Match` semantics stay identical. Server contract is frozen for this PR.
- **No change to `useFlushOnUnload`.** Mutations can't survive `pagehide`; the hook stays direct-call. Documented.
- **No change to `ApollonLocal.tsx`.** Local mode reads from `usePersistenceModelStore` (IndexedDB / localStorage layer), not the HTTP client; nothing to migrate.

## Migration sequence

Six atomic commits, each leaves `npm test` green:

1. **Add dep + `QueryClient` + provider mount.** No call sites converted yet. Verifies setup, devtools wiring, test infra (`renderWithQuery` helper in `tests/setup.ts`).
2. **`useDiagram(id)` + ApollonWithConnection migration.** Reset block deleted, manual `AbortController` deleted, `setIsLoading` deleted. `DiagramApiClient.fetchDiagram` reverts to no-signal signature (Query passes `signal` to `queryFn` arg). The 5 existing regression tests must stay green with the same revert-proofs.
3. **`useAutosave(id)` migration.** `setInterval` moves into the hook; component just calls `useAutosave(diagramId)`. `useFlushOnUnload` untouched.
4. **`useVersionsInfinite` + `useVersionBody` + `useVersionEventBridge`.** `VersionDrawer`, `VersionThumbnail`, and the WS branch in `ApollonWithConnection` migrate together (they share cache keys). `useVersionStore` gets pruned to client-only state in the same commit.
5. **Mutation hooks for create/restore/edit/delete version + create diagram.** Modal call sites switch. Optimistic-update tests added.
6. **Delete `DiagramApiClient.fetchDiagram`'s `opts.signal` param + the matching `request()` plumbing audit.** Pure cleanup; lockstep with commit 2.

Each commit is independently revertable. PR is opened as commits 1–6 squashed for review but kept as a stacked branch on the author's side until merge — reviewers can checkout any commit to verify the suite passes.

## Acceptance criteria

- [ ] `npm test` green; all 5 existing regression tests in `ApollonWithConnection.test.tsx` pass with the same revert-proofs (commit message lists each: `setIsLoading` revert / `AbortError` filter revert / `hasPromptedRef` reset revert — each line still turns exactly one test red)
- [ ] `npx tsc -b --noEmit` clean
- [ ] **Net deletion** ≥ 100 LOC across `src/pages/ApollonWithConnection.tsx`, `src/stores/useVersionStore.ts`, `src/components/versioning/VersionThumbnail.tsx`, `src/components/versioning/VersionDrawer.tsx`, `src/services/DiagramApiClient.ts`. PR description quotes the `git diff --stat` output.
- [ ] `grep -rn "AbortController\|cancelled" src/` returns **zero** matches inside React component/hook files (allowed only inside services/, tests/, and `WebSocketManager.ts`)
- [ ] `grep -rn "signal?: AbortSignal" src/services/DiagramApiClient.ts` returns **zero** matches (Query owns cancellation)
- [ ] `useVersionStore` interface is reduced to client-only fields. PR description quotes the before/after type signature.
- [ ] New tests: `useDiagram.test.tsx`, `useVersionsInfinite.test.tsx`, `useVersionBody.test.tsx`, `useAutosave.test.tsx`, `useVersionEventBridge.test.tsx`. Each asserts at least one behavior whose revert turns it red (commit body documents the revert).
- [ ] React DevTools profiler shows ≤ current render count for the editor mount path (informational baseline + post-PR result in PR description; **not gated** on numeric improvement — the goal is removing scaffolding, not a perf win).
- [ ] Yjs/Query contract documented in `useDiagram.ts` header comment (≥ 4 lines): `staleTime: Infinity` + `gcTime: 0` + why + the two legitimate post-mount writes.
- [ ] Manual smoke (steps written in PR description, executed by author):
  - Initial load of a collaborative diagram: overlay shows then disappears; WS connects; peer cursors appear.
  - Share-again from inside a diagram: overlay re-shows on the new URL; no blank screen; no stale model.
  - New diagram from inside a collaboration: same.
  - Restore a version locally: overlay does not flicker; canvas swaps to historical state.
  - Restore a version on peer A while peer B is mid-edit: peer B sees toast + canvas updates after the WS event (no `setQueryData` race with concurrent Yjs ops).
  - Close tab during autosave: server receives the final PUT via `useFlushOnUnload` (DevTools network tab confirms).
  - Open Version Drawer, scroll through 100+ versions, hover thumbnails: all bodies load on visibility, none load off-screen, no spinner flicker.

## Tests to write

- `useDiagram.test.tsx` — revert-proof: `staleTime: 0` instead of `Infinity` triggers a refetch on remount; pinning `Infinity` keeps the cache. Test calls `queryClient.getQueryState` to verify the seed body is the one editor mounted with.
- `useVersionsInfinite.test.tsx` — pagination cursor passed through; `fetchNextPage` returns the next page; `hasNextPage` collapses when server returns `nextCursor: undefined`.
- `useVersionBody.test.tsx` — `enabled: false` does not call the service; flipping to `true` fires exactly one request.
- `useAutosave.test.tsx` — `If-Match` rebase on `REVISION_MISMATCH` preserves `lastObservedHeadRev` when meta absent (the exact contract preserved from `ApollonWithConnection.tsx:301-310`); unmount mid-flight aborts.
- `useVersionEventBridge.test.tsx` — dispatch a `VERSION_DELETED` event; `removeQueries({ queryKey: ['diagram', id, 'version', vid] })` was called; `invalidateQueries({ queryKey: ['diagram', id, 'versions'] })` was called.
- `ApollonWithConnection.test.tsx` — port the existing 5 tests onto `QueryClientProvider`. Replace `fetchHoisted` with a `queryClient` whose service mock controls when the fetch resolves. Maintain revert-proofs.

## Implementation notes

- **`gcTime: 0` on `['diagram', id]` is deliberate.** A re-join of the same collaborative room (close → reopen tab on the same diagram) MUST re-fetch — peers may have committed updates that are now in Yjs's authoritative state on the server side. A non-zero `gcTime` here would risk seeding the editor with stale body. The single body fetch on remount is dirt cheap (one GET); the cost of staleness is correctness.
- **`refetchOnWindowFocus: false` is global, not just for `['diagram', id]`.** Defending: the versions list COULD benefit from focus refetch, but the WS bridge (`useVersionEventBridge`) is the source of truth for that list — peers create/delete/rename versions through the WS, not through HTTP that this client missed. Focus refetch would race the bridge.
- **`setQueryData` from inside a Yjs subscription is forbidden.** Yjs and Query each have their own batching/scheduling; coupling them via shared cache reads inside a subscription causes priority-inversion bugs. The bridge hook reads control events (a narrow Apollon-defined channel), not Yjs ops. Document this rule in the bridge hook's header.
- **`useMutation`'s `signal` for autosave.** TanStack Query v5's `useMutation` does NOT pass `signal` to `mutationFn` by default the way `useQuery` does. We wire it manually inside `useAutosave` via `AbortController` scoped to the hook's `useEffect` cleanup; component unmount → `controller.abort()` → mutation's in-flight fetch aborts. The autosave hook becomes the single owner of mutation cancellation, mirroring the previous bespoke pattern with the same semantics.
- **`useInfiniteQuery` vs paginated `useQuery`.** Picked `useInfiniteQuery` because the version drawer is append-only scroll; `getNextPageParam` reads `nextCursor` from the previous page. The alternative (`useQuery` + manual cursor state) recreates the wheel. v5's `useInfiniteQuery` API requires `initialPageParam` — wire `undefined`.
- **Rejected: `select` for fine-grained re-render control.** Tempting on `['diagram', id]` to memoize subsets, but the editor mount only reads `data` once; subsequent reads come from Yjs. `select` would add cognitive overhead for zero render savings. Revisit only if the versions drawer's render profile shows a problem.
- **Rejected: `placeholderData: keepPreviousData` on version pagination.** Page transitions are not user-visible — the drawer already has a stable scroll position and we display partial-loaded states. `keepPreviousData` would cause one ghost-render of the prior page during transition.
- **Rejected: `useSyncExternalStore` adapter to bridge Yjs into Query.** The editor IS the Yjs subscriber; the data flow is `Yjs → editor.model → React state inside library/`. Pulling it back through Query would require Yjs↔Query bidirectional sync. Out of scope and architecturally wrong (Query is for request/response, not pub/sub).
- **`QueryClient` is a singleton, not a context-provided instance.** Tests that need isolation create their own client; production uses one. Per TanStack docs and the v5 migration guide, a per-test client is the canonical way to isolate cache state across `it()` blocks.
- **DevTools mount.** `<ReactQueryDevtools initialIsOpen={false} />` lives inside `AppProviders` behind `import.meta.env.DEV`. Tree-shaking strips it from prod bundles. Confirmed via `vite build --report` (PR description includes bundle delta — expected: 0 prod bytes added beyond core Query runtime).
- **ESLint rule:** add `@tanstack/eslint-plugin-query` with `recommended` preset. Catches the common mistakes (`exhaustive-deps`, stale closures over `queryClient`, mutation `useEffect` antipattern).
- **No `useQuery` inside conditionals or loops.** Standard React-hook rule; called out because the migration touches 8 files and reviewers should verify.
- **`request()` in `DiagramApiClient` keeps its `signal` parameter.** That's the HTTP boundary — Query forwards `signal` from `queryFn` → service → `request()` → `fetch`. We delete the *re-exported* `opts: { signal? }` on the *named methods* (`fetchDiagram`, etc.), not the underlying `request()` plumbing.
- **Manual cache eviction on logout.** Not relevant today (no auth wall on the webapp), but if/when added, `queryClient.clear()` on session end is the one-liner. Noted for future.

## Risk

| Risk | Mitigation |
|---|---|
| Yjs↔Query cache divergence (Query holds stale body, editor edits → autosave commits a divergent state) | Contract: `staleTime: Infinity` + `gcTime: 0` + only two documented post-mount writers (restore, preview-exit). `useDiagram` header comment enforces the rule for the next contributor. |
| Autosave mutation aborted mid-flight when user navigates → unsaved edits lost | `useFlushOnUnload` stays untouched; it runs on `pagehide` with `navigator.sendBeacon`-equivalent timing. Mutation abort is for *route* unmounts, not *tab close*. The two paths are independent. |
| WS bridge invalidates cache while VersionDrawer is mid-render → tearing | Query's `invalidateQueries` schedules a refetch; it does not synchronously mutate `data`. React batches the resulting re-render normally. No tearing risk. |
| `useInfiniteQuery` `getNextPageParam` returning `undefined` confused with "load more" | v5 contract: `undefined` → `hasNextPage: false`. Tested in `useVersionsInfinite.test.tsx`. |
| Adding `@tanstack/eslint-plugin-query` triggers churn on unrelated files | Configure plugin to apply only to `src/hooks/use*` and `src/pages/*` initially; widen after stabilization. Documented in the eslint config diff. |

## Dependencies

Follows #ee854a31 (blank-screen fix; current branch under review). No required-by issues at this time.

FelixTJDietrich:
### Summary

The web app fetched HTTP data (version history, version snapshots, shared diagrams) with hand-written code: `fetch` calls wired through optional `AbortSignal` parameters, `cancelled` flags inside effects, and a slice of server data living in a Zustand store next to genuine client state. This PR moves **version history** onto [TanStack Query](https://tanstack.com/query) — the list, the immutable snapshot bodies, and the create/rename/delete/restore mutations — so pagination, optimistic updates with rollback, and request deduplication come from one well-tested library instead of bespoke scaffolding.

It is **behaviour-preserving by design** — the editor looks and works exactly as before (matching the non-goals in #687).

**On size, stated honestly:** `useVersionStore` drops from **563 to ~170 lines** and now holds only client state, and the five files the issue calls out shrink by a net ~336. But the app's *shipped* code **grows by roughly 670 lines**, because the ~730-line `src/queries/` layer is new. This PR does not make the codebase smaller. What it buys is a single shared cache for version bodies (fewer requests — thumbnails, preview entry and the dirty-check baseline now share one fetch), real cursor pagination, and a rollback primitive replacing hand-written splice-restore logic. Bundle cost is **+10 kB gzipped** against a 1300 kB budget.

One framing in the original issue does not survive contact with the code, and is worth correcting: the "delete duplicated fetch boilerplate" argument is thin here. `origin/main` had exactly one hand-rolled `cancelled` flag (`VersionThumbnail`) and two `AbortController`s (`ApollonShared`); loading/error handling was already centralised in one store rather than duplicated across call sites.

The collaborative document itself is untouched: Yjs over the WebSocket stays the single source of truth for the live canvas. TanStack Query owns request/response HTTP only, and that boundary is documented at the top of `queryClient.ts`. The debounced autosaver and the `pagehide` flush are deliberately left out of Query (a `keepalive` beacon can't run through a cancellable mutation) — see the implementation notes.

Closes #687.

### Release note

No visible change to the editor. Under the hood, the web app now loads and caches version history, version snapshots, and shared diagrams through one shared data layer instead of hand-written fetching, so the version panel refreshes and reconciles more consistently across tabs and collaborators.

### Implementation notes

A few decisions are load-bearing and worth a reviewer's attention. Each is documented next to the code it governs.

- **Query hooks sit on the existing `VersionRepository` port, not on the REST client.** Version history runs against two backends — REST for shared diagrams, IndexedDB for local ones — behind one port. The hooks resolve the adapter by a `kind` (`"local" | "remote"`) that the **editor route** declares via a small React context. Modals and toast bodies render at the app root, *outside* that route subtree, so they take `kind` as a prop instead. (An earlier revision used a mutable module global for this; a review found it could re-key the outgoing page's queries onto the wrong backend during a `/local → /shared` navigation, so it was replaced.)
- **The editor seed is deliberately NOT a query.** The body the editor mounts with is an initialisation input, not server state: Yjs owns the document from mount onwards, so it must never be refetched or replayed from a cache. Expressing that through Query meant disabling caching in both directions (`staleTime: Infinity, gcTime: 0`) — a plain fetch wearing a cache's clothes — so it is now `useDiagramSeed`, one abortable fetch per diagram. The two imperative "give me the latest HEAD" reads (peer restore, preview exit) each own an `AbortController`, so one caller's teardown cannot abort another's request.

- **No optimistic rows are forged into the cache.** The version-create mutation returns its settle-time invalidation promise, so it stays "pending" until the refreshed list lands; the drawer renders the in-flight row straight from that pending state. (An `isSuccess`-derived row turned out to be sticky and could resurrect a just-deleted version.)
- **Autosave and flush-on-unload stay outside Query.** They keep their own retry/rebase and `keepalive` semantics, which mutations can't express; this is stated in `queryClient.ts` so the boundary is explicit.

While migrating, the review pass also surfaced and fixed a handful of edge cases in the same code: an undo-restore toast that read its backend from a context it renders outside of (it now takes the value as a prop); a peer deleting the version you're previewing no longer leaves a stale snapshot on the canvas; and a failed history load now shows a "couldn't load" surface instead of the "no versions yet" empty state. Each has a regression test that fails when the fix is reverted.

**Not done here / follow-ups (non-blocking):**
- The version-list query uses `refetchOnWindowFocus`, and v5 refetches *all* loaded pages on focus. For very long histories a `maxPages` bound would cap that; left out until a profile shows it matters.
- A peer's restore can still momentarily clobber an active local preview overlay. This predates the PR and isn't touched.

### Steps for testing

Everything below is behaviour that should be **unchanged** — the point is to confirm the migration didn't regress anything.

1. **Collaborative load & share-again.** Open a shared diagram — the loading overlay shows then clears, the canvas mounts, peers connect. From inside it, "Share again" to a new URL: the overlay re-shows, no blank screen, no stale model.
2. **Version panel.** Open the version drawer, save a version (the "Saving…" row should hand over to the real row with no gap), rename one, delete one — the deleted row must not reappear. Scroll a long history and hover thumbnails: only visible rows fetch (check the Network tab).
3. **Undo restore.** Restore a version in a collaborative session, then click **Undo** on the toast — it must complete without error. *(This path threw before the fix in this PR; the regression test is `UndoRestoreToast.test.tsx`.)*
4. **Peer delete during preview.** While previewing a version, have a second client delete that version — your canvas must exit preview rather than keep showing the deleted snapshot.
5. **Local (offline) mode.** Repeat the version-panel steps at `/local/:id` to confirm the IndexedDB backend shares the same code path.

**On collaborative coverage, stated plainly:** the webapp's e2e stack serves the bundled frontend only — no API server, no Redis, and no existing spec touches `/shared` — so a genuine two-browser collaboration test cannot run in CI without adding server infrastructure to every PR. Instead, `ApollonShared.collab.test.tsx` drives the exact `ControlEvent` sequence the server publishes to a peer through the **real** page, store, query cache and event bridge, doubling only the WebSocket, HTTP and editor. That covers this client's reaction to a collaborator — where every defect found while building this layer lived — and not the server's behaviour, which the server suite owns. Steps 3–4 above are still worth doing by hand once.

Automated, on this branch:
- `pnpm --filter @tumaet/webapp test` → **291 passing** (44 files; 287 on `main`). New: the query layer, the collaborator-actions spec, and a regression test for each defect fixed here.
- `pnpm --filter @tumaet/webapp exec tsc -b --noEmit`, `pnpm lint`, `pnpm build` → clean.

### Screenshots / screencasts

No UI change — this is behaviour-preserving plumbing, so there is nothing new to show.

### Checklist

- [x] Linked to a related issue — closes #687
- [x] Added a changeset in the user's voice (`@tumaet/webapp`, patch)
- [x] PR title's Conventional Commit type (`refactor`) matches the kind of change
- [x] Tests added or updated — 5 new query-layer test files + regression tests for each fix
- [x] Ran `pnpm lint && pnpm build && pnpm test` locally — green (`pnpm format:check` is clean for all committed files)
- [ ] Documentation updated — n/a (no public API or docs change)
- [ ] Screenshots or screencasts — n/a (no UI change)


standalone/webapp/src/components/versioning/VersionDrawer.tsx
```diff
@@ -239,32 +279,32 @@ export const VersionSidebarBody: FC<Props> = ({
...
-        toast.error(t.failureToCreate)
+    // Call-site callbacks run before the option-level `onSettled` awaits the
+    // refetch, so `lastLocalSaveIdRef` is set while the fresh row is still
+    // landing — otherwise the baseline effect re-fetches a body we have.
+    createMutation.mutate(
+      { body: editor.model, name, description: description || undefined },
+      {
+        onSuccess: (summary) => {
+          lastLocalSaveIdRef.current = summary.id
+          onVersionSaved?.(summary.headRev)
```
Claudia-Anthropica: @FelixTJDietrich [medium] This success handler is observer-scoped, so TanStack Query drops it if `VersionSidebarBody` unmounts while the POST is in flight (closing the rail or sheet does exactly that). The create endpoint bumps `headRev`, but `onVersionSaved` then never updates and settles the page autosaver, leaving it dirty with a stale revision and forcing a revision-mismatch/re-PUT on the next save; await `mutateAsync` here as before, or move this autosaver update to mutation-level logic that survives unmount.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `standalone/webapp/src/components/versioning/VersionDrawer.tsx`, `onVersionSaved(summary.headRev)` is inside a per-call `mutate` callback that is skipped when the sidebar unmounts. Refactor `handleCreate` to ensure the HEAD-revision/autosaver update survives unmount while preserving the pending-row and list-invalidation behavior, and add a regression test that unmounts the sidebar before the create request resolves.\n\n</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The query migration is thoughtfully structured and well tested overall. One mutation-lifecycle regression can leave the collaborative autosaver with a stale HEAD revision when the version panel closes during a save; please see the inline comment.

standalone/webapp/src/components/versioning/VersionDrawer.tsx
FelixTJDietrich: Good catch — fixed in edd71ee6, and you were right about the mechanism.

`useCreateVersionMutation` now takes an `onCommitted` option and calls it from the mutation's own `onSuccess`, which runs regardless of the caller's lifecycle. The panel keeps only genuinely panel-local UX (clearing the composer) at the `mutate()` call site — the same split the module header already documents, which I had violated here.

Regression test: `versionMutations.test.tsx` → "fires even when the caller unmounts before the create resolves" unmounts the hook's consumer before the create settles and asserts the callback still fires with the committed `headRev`. I verified it is revert-proof — moving the handoff back to the call site turns it red.

Claudia-Anthropica: @FelixTJDietrich Nice fix — the mutation-level callback and unmount regression test cover the lifecycle issue. Thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich The mutation lifecycle issue is fixed and covered by the unmount regression test. All feedback addressed, nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich The devtools opt-in change is clean, and the TanStack Query migration remains well structured and thoroughly covered. Approving the new head.

Claudia-Anthropica (approved):
@FelixTJDietrich The README now makes the opt-in Query devtools flag discoverable and accurately documents why it stays disabled by default. The TanStack Query migration remains clean and well covered; approving the new head.

Claudia-Anthropica (approved):
@FelixTJDietrich The focus-refetch opt-in restores the previous open-panel-only behavior and the added tests cover both observer modes. The unrelated legal-page changes are cleanly removed; approving the new head.

standalone/webapp/src/stores/versionStoreBootstrap.ts
```diff
@@ -39,33 +48,40 @@ export function ensureVersionStoreBootstrapped(): void {
...
+        const previewing = store.preview
         if (!previewing || previewing.diagramId !== msg.diagramId) return
-        const stillThere = (after.versions[msg.diagramId] ?? []).some(
-          (v) => v.id === previewing.versionId
-        )
-        if (!stillThere) after.exitPreview()
+        const versions = getCachedVersions(queryClient, "local", msg.diagramId)
+        // Only act on a loaded list — an empty cache proves nothing.
+        if (versions && !versions.some((v) => v.id === previewing.versionId)) {
+          store.exitPreview()
```
Claudia-Anthropica: @FelixTJDietrich [medium] This clears only Zustand after the list refetch, but `?version=` remains the source of truth. `useVersionPreviewUrlSync` immediately re-enters, and because `fetchVersionBody` now has `staleTime: Infinity` and this path never removes the deleted body key, it serves the cached snapshot without touching IndexedDB; a version deleted in another tab therefore stays previewed indefinitely. Evict the missing body query before exiting (so the retry 404 clears the URL), or move this reconciliation to a route-aware owner that calls `closePreview()`; extend the test through the URL/body-cache path.

In `standalone/webapp/src/stores/versionStoreBootstrap.ts`, cross-tab deletion clears only the preview store while the URL and immutable body cache immediately restore the deleted preview. Reconcile the deleted preview by clearing the URL through a route-aware owner or evicting the missing body query before exiting, and add a regression test that covers the URL sync plus cached-body path.

</details>

standalone/webapp/README.md
```diff
@@ -6,7 +6,24 @@ Part of the Apollon monorepo — run it from the repo root with `pnpm dev`, not
 
 ## Stack
 
-React 19, TypeScript, Vite, the shadcn-style [`@tumaet/ui`](../../packages/ui) design system (Base UI primitives + Tailwind v4), Storybook, Vitest, Playwright (visual + e2e).
+React 19, TypeScript, Vite, the shadcn-style [`@tumaet/ui`](../../packages/ui) design system (Base UI primitives + Tailwind v4), [TanStack Query](https://tanstack.com/query) for server state, Storybook, Vitest, Playwright (visual + e2e).
+
+## Debugging server state
+
+HTTP data (diagram loads, version history, version snapshots) goes through
```
Claudia-Anthropica: @FelixTJDietrich [low] This is no longer accurate after `4840de76`: editor seed loads use the plain `useDiagramSeed` fetch, not TanStack Query. The changeset similarly claims shared diagrams are loaded and cached through the shared data layer. Please limit both descriptions to version history/snapshot bodies and Query-backed mutations, or document the deliberate seed exception.

In `standalone/webapp/README.md` and `.changeset/tanstack-query-webapp.md`, the documentation incorrectly says diagram/shared-diagram loads go through TanStack Query after the editor seed moved to `useDiagramSeed`. Update both descriptions to accurately distinguish the plain editor-seed fetch from Query-managed version history, snapshot bodies, and mutations.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The Query migration remains thoughtfully structured and well tested, but the local cross-tab deletion path can keep a deleted snapshot preview alive through the immutable body cache. Please fix that reconciliation issue; I also noted two stale descriptions from the latest seed refactor.

standalone/webapp/src/stores/versionStoreBootstrap.ts
FelixTJDietrich: Fixed in 495c5a5d — and you had the mechanism exactly right (immutable body cache + URL-as-source-of-truth = deleted snapshot served forever).

Took the "evict the missing body query before exiting" option, since the bootstrap is a non-React singleton that can't reach the router: the path now `removeQueries` the deleted body key before `exitPreview()`, so the URL sync's re-entry misses the cache, reads IndexedDB, 404s, and its existing error path strips `?version=`.

The bootstrap test is extended through the body-cache path — it warms the body, then asserts both that the key is evicted and that a re-read rejects with 404. Reverting the eviction turns it red.

standalone/webapp/README.md
FelixTJDietrich: Fixed in 495c5a5d. The README and the changeset now scope Query to version history (list, snapshot bodies, and the create/rename/delete/restore mutations) and call out the deliberate exception: the editor seed is a one-shot `useDiagramSeed` fetch, not a query, pointing at the hook.

standalone/webapp/src/stores/versionStoreBootstrap.ts
Claudia-Anthropica: @FelixTJDietrich Nice fix — evicting the immutable body before exiting makes the URL-sync retry hit IndexedDB and clear the dangling parameter. The warmed-cache regression test covers the failure mode well.

standalone/webapp/README.md
Claudia-Anthropica: @FelixTJDietrich Looks good now — both descriptions accurately scope Query to version history and call out the one-shot editor seed exception.

Claudia-Anthropica (approved):
@FelixTJDietrich Both requested changes are fixed and covered by the updated tests. All feedback addressed, nice work.

PR #822, opened 2026-07-17
feat(library): place palette elements with a tap or keyboard, not only drag

FelixTJDietrich:
### Summary

The node palette only supported **drag-to-place**. A quick click or tap was interpreted as a zero-distance drag, so the new node was created at the pointer — which, for a press on the palette, is **underneath the palette itself**, where the user can't see it. This surprised people who expected "click the shape, get the shape."

This PR keeps drag exactly as it was and adds **tap / click / keyboard placement**:

- **Click or tap** a palette element → it's placed in the **centre of the visible canvas** and selected.
- **Repeated** placements **cascade diagonally** (the same 20 px step as paste) so they never stack on top of each other.
- Palette entries are now **keyboard- and screen-reader-operable**: focus one and press **Enter** or **Space**.
- **Dragging is unchanged** — it still drops the element exactly where you release it, with the same nesting behaviour.

### Release note

Click or tap a palette element to drop it in the centre of the canvas and select it — no more nodes landing hidden under the palette; dragging still places elements exactly where you release them.

### Implementation notes

**Where the logic lives.** Node creation moved out of the `DraggableGhost` component into a `usePalettePlacement` hook (matching how `useSelectionForCopyPaste` / `useMultiSelectionMode` structure interaction logic). The component is now gesture + ghost only; the hook owns creation. Pure, framework-free helpers (`buildPaletteNode`, `resolveTapPosition`, …) live in `utils/paletteNode.ts` and are unit-tested in isolation.

**One nesting path for both gestures.** Drag and tap both build nodes through `buildPaletteNode` and resolve their parent through the same `canDropIntoParent`-checked hit-test, so a tap can never create a nesting that a drag forbids (e.g. dropping into an incompatible BPMN container).

**Gesture classification.** A press is a drag once the pointer travels past a slop threshold *at any point* during the gesture (not just measured start-to-end), so an out-and-back drag can't be misread as a tap. The threshold is looser for touch (16 px) than mouse (8 px), because finger-roll on an intended tap routinely exceeds a mouse-tight value. Placement is routed through `click` (covering mouse, touch, and assistive-tech-synthesised activation) plus an `Enter`/`Space` key handler; the click that trails a real drag is suppressed based on whether a node was *actually* placed.

**Cascade.** Consecutive placements offset off the previously-placed, still-selected node — but only while the resulting node stays on screen, otherwise it falls back to centre, so a placed node is never dropped out of view. State updates use a functional `setNodes`, so rapid placements can't lose each other.

**Accessibility.** Each entry exposes a unique, human-readable accessible name (e.g. "Add element: Interface" — previously the four `class`-family entries all announced the identical machine name), a visible focus ring, and ignores key auto-repeat.

**Deliberately deferred (transparency).** Two accessibility improvements are *not* in this PR because they are palette-wide and would risk the palette's existing view-switch controls — they belong in a focused follow-up, and neither is a WCAG Level-A failure:
- **Roving tabindex / `role="toolbar"`.** Every palette cell is currently its own tab stop. It's operable and labelled, but the [WAI-ARIA APG](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/) prefers one tab stop + arrow-key traversal for a group of controls.
- **`aria-live` status announcement** on placement (WCAG 4.1.3) so a screen-reader user hears "added", rather than relying on the selection change.

### Steps for testing

1. Open any diagram in the editor.
2. **Click** (don't drag) a palette element → it appears in the middle of the canvas, selected — not hidden behind the palette.
3. Click the same element **two or three more times** → the nodes step diagonally instead of stacking; only the most recent is selected.
4. **Drag** an element onto the canvas → it still lands exactly where you release it (unchanged).
5. Drag an element and release it **back over the palette** → nothing is dropped there, and it's instead placed in the centre (the tap you meant).
6. **Keyboard:** `Tab` to a palette element (visible focus ring) and press **Enter** or **Space** → it's placed and selected.
7. **Screen reader:** the four Class-diagram entries announce as distinct names (Class / Abstract / Enumeration / Interface).

### Screenshots / screencasts

Interaction change (no static visual to show). It's covered end-to-end by a new Playwright test — `standalone/webapp/tests/e2e/palette-tap-to-place.spec.ts` — which asserts the tap lands inside the canvas, is selected, and that repeats cascade. A screencast can be added on request; I can't attach binary media from this environment.

### Checklist

- [ ] Linked to a related issue (if applicable) — no tracking issue linked
- [x] Tests added or updated — 3 unit specs + 1 e2e (`paletteNode`, `paletteTapPlacement`, `draggableGhostTheme` mocks, `palette-tap-to-place`)
- [x] Ran locally — green (see note)
- [ ] Documentation updated (if applicable) — no docs affected
- [ ] Screenshots or screencasts attached (if a UI change) — interaction change; covered by e2e, see above

> **What I ran locally (transparency):** `pnpm format:check` (repo-wide, clean), `pnpm lint` (0 errors), `pnpm test` (`@tumaet/apollon` — 1314 pass / 1 skip), `pnpm -w run build:lib` (green), and the tap + drag + swimlane Playwright specs (green). Changes are library-scoped; I did not run the full `server`/`webapp`/`vscode` build locally — CI covers those.


library/lib/hooks/usePalettePlacement.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    const center = screenToFlowPosition({
+      x: rect.left + rect.width / 2,
+      y: rect.top + rect.height / 2,
+    })
+    const topLeft = screenToFlowPosition({ x: rect.left, y: rect.top })
+    const bottomRight = screenToFlowPosition({ x: rect.right, y: rect.bottom })
+
+    // Cascade off a single selected node (an edge selection yields no anchor).
+    const selected =
```
Claudia-Anthropica: @FelixTJDietrich [medium] This treats any single selected node as the cascade anchor, so the first palette activation after selecting an existing node is placed beside that node instead of at the viewport centre. Track the last tap-placed node explicitly and only cascade while that exact node remains the sole selection.

In `library/lib/hooks/usePalettePlacement.ts`, `placeAtViewportCenter` uses any sole selected node as a cascade anchor, including nodes not created by the preceding palette activation. Track the last tap-placed node across palette controls and only use it as the anchor when it is still the sole selection; otherwise place at the viewport centre, and add a regression test for an unrelated pre-selected node.

</details>

library/lib/hooks/usePalettePlacement.ts
```diff
@@ -0,0 +1,263 @@
...
+    (
+      build: (prev: Node[]) => Node,
+      parentId: string | undefined,
+      select: boolean
+    ) => {
+      setNodes((prev) => {
+        const newNode = build(prev)
+        const next = [...prev, newNode]
+        if (parentId) resizeAllParents(newNode, next)
+        return select
```
Claudia-Anthropica, line 117: @FelixTJDietrich [medium] The sole-selection branch only clears `node.selected`. If an edge was selected beforehand, it stays visually selected even after `selectedElementIds` is replaced with the new node ID, leaving the rendered selection out of sync. Clear selected edges as part of the same placement flow too.

In `library/lib/hooks/usePalettePlacement.ts`, tap placement deselects existing nodes but leaves any selected edges marked `selected`. Update the placement flow to clear selected edges when the new palette node becomes the sole selection, and add a test starting from a selected edge.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The shared placement path is a solid direction, but the tap-selection state has two regressions: an unrelated selected node can become the first cascade anchor, and a selected edge is not visually cleared. See the inline comments for the affected paths.

library/lib/hooks/usePalettePlacement.ts
FelixTJDietrich: Fixed in 2c070380. The cascade now anchors on a new `lastPlacedElementId` in the diagram store (set on every tap placement) and only offsets while that exact node is still the sole selection — otherwise it centres. So selecting an unrelated node and then tapping a palette entry now places at the viewport centre, not beside the selection. Added a regression test (`centres (not beside) when an unrelated node is selected, not the last placed one`) in `paletteTapPlacement.test.tsx`.

library/lib/hooks/usePalettePlacement.ts
FelixTJDietrich: Fixed in 2c070380. When a tap becomes the sole selection, the placement flow now also clears `selected` on any selected edge (mirrors paste), so the rendered edge selection no longer lingers out of sync with `selectedElementIds`. Added a regression test starting from a selected edge (`clears a previously selected edge so the rendered selection stays in sync`).

PR #823, opened 2026-07-18
docs: add editor hero, live-demo link, social card, and "Why Apollon"

FelixTJDietrich:
Closes #750

For a visual editor, the repo's shop window had no visuals: the READMEs showed only badges, the live demo was easy to miss, and repo links unfurled without an image. (The feature-request template and issue-chooser config from the issue checklist already landed earlier, so this PR covers the rest.)

## What changed

**Generated, never hand-made assets.** A new `readme-assets` Playwright project generates every public-facing image straight from the live editor — the same one-source-of-truth pattern as `howto-assets`. It renders a curated showcase fixture (`readme-hero.json`, a class diagram of Apollon's own domain model) and writes to `docs/static/img/`:

- `apollon-editor-light.png` / `apollon-editor-dark.png` — README hero, 1440×810 @2×
- `apollon-social-card.png` — GitHub social preview, exactly 1280×640 (2:1, well under the 1 MB limit), composed in-page with the repo's bundled Inter fonts and logo so it renders identically on any machine

The per-PR visual job diff-gates all three, so the marketing images can never silently drift from the real product. `update-visual-baselines.yml` stages the new paths, so the canonical-container refresh flow covers them too.

**README (repo front page).** Centered header with a **▶ Try the live demo** link row, a dark/light `<picture>` hero linked to the demo, and a "Why Apollon / who it's for" section for people arriving from outside Artemis.

**npm page.** `library/README.md` now embeds the hero via an absolute `raw.githubusercontent.com` URL (relative paths don't resolve on npmjs.com), with a patch changeset so it ships on the next publish.

**Docs site.** `themeConfig.image` now points at the generated social card — closing the existing config TODO ("a dedicated 1200×630 card would render better in social embeds") — so docs links unfurl with the card as well. `visual-tests.md` documents the generated-assets projects.

## Preview

| Hero (light) | Social card |
| --- | --- |
| hero light | social card |

## Manual step for a maintainer (no GitHub API exists for this)

Upload `docs/static/img/apollon-social-card.png` under **Settings → General → Social preview**. Re-upload whenever the card regenerates.

## Verification

- `pnpm exec playwright test --project readme-assets` passes both with and without `--update-snapshots` (stable diff).
- Webapp + docs lint clean, Prettier clean, docs site builds with the new `og:image` (`https://ls1intum.github.io/Apollon/img/apollon-social-card.png` in the built HTML).
[...]

PR #824, opened 2026-07-18
fix(library): stop mutating the host page's color-scheme via style.css

FelixTJDietrich:
## Problem

After #823 went live, https://ls1intum.github.io/Apollon/ looked broken for dark-preference visitors — dark canvas and dark-flipped colors bleeding into the light-themed page. The trigger is pre-existing (not introduced by #823, which only refreshed text/assets), but it is very visible on the docs homepage.

## Root cause

`packages/ui/src/styles/tokens.css` declared:

```css
:root {
  color-scheme: light dark;
}
```

These tokens ship inside `@tumaet/apollon/style.css`, so **every host page that imports the editor CSS gets its document root's `color-scheme` rewritten** — exactly the "host mutating `:root`" anti-pattern the file's own scoped-theming comment warns against. Two user-visible consequences on hosts that leave `html`/`body` backgrounds transparent (Docusaurus does):

1. **Dark canvas under light pages.** With every background transparent down to `<html>` (verified via `elementsFromPoint` — nothing paints a background), the browser resolves the page canvas from `color-scheme` + visitor preference: dark-preference visitors get a dark canvas behind light-theme text.
2. **Tokens follow the OS, fighting the host's theme toggle.** lightningcss downlevels the declaration into a global `@media (prefers-color-scheme: dark)` helper-variable flip (visible in the deployed `styles.e10633a2.css`), so token-driven colors switch with the OS scheme even while Docusaurus's `data-theme` stays `light`.

## Fix

Scope the declaration instead of mutating the host root:

- `.apollon-editor` and `[data-theme="light"]` subtrees declare `color-scheme: light` (native controls/scrollbars inside the editor still adapt);
- the existing `:root[data-theme="dark"], [data-theme="dark"]` block keeps `color-scheme: dark` (opt-in via attribute — host-controlled);
- the webapp already sets the document-root `color-scheme` at runtime (`useThemeStore` → `applyThemeToDocument`), so it needs nothing.

## Verification

- Rebuilt `library/dist` CSS now contains exactly one scoped `color-scheme: light` (`.apollon-editor, [data-theme=light]`) and one `color-scheme: dark` — no bare `:root` declaration.
- Rebuilt docs bundle contains **zero** `prefers-color-scheme` blocks (the old deployed bundle had the global var-flip block).
- Headless probe of the rebuilt docs site: `html` computed `color-scheme` is now `light`/`dark` following Docusaurus's own `--ifm-color-scheme` (was `light dark` before), `data-theme` toggles correctly per preference, and the full-page renders light-on-light / dark-on-dark.
- Webapp behavior unchanged (root `color-scheme` set at runtime by the theme store); the per-PR visual suite gates the rest.

A patch changeset is included so embedders get the fix on the next release — this bug affects **any** host that imports `@tumaet/apollon/style.css`, not just the docs site.

### DIFF
--- a/.changeset/light-hosts-stay-light.md
+++ b/.changeset/light-hosts-stay-light.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Stop mutating the host page's color scheme. `style.css` declared `color-scheme: light dark` on `:root`, so any host that imports the editor CSS and leaves its own `html`/`body` backgrounds transparent (the Docusaurus docs site does) rendered a browser-picked dark canvas behind light-themed text for dark-preference visitors. The declaration is now scoped to the editor mount and `[data-theme]` subtrees; host pages keep full control of their document root.
--- a/packages/ui/src/styles/tokens.css
+++ b/packages/ui/src/styles/tokens.css
@@ -12,9 +12,20 @@
    Static CSS = correct first paint, no runtime <style> injection.
    ────────────────────────────────────────────────────────────────────────── */
 
-:root {
-  color-scheme: light dark;
+/* color-scheme is deliberately NOT set on bare :root. This stylesheet ships
+   inside @tumaet/apollon/style.css, so a `:root { color-scheme: light dark }`
+   would mutate every HOST page that imports the editor CSS: a host that leaves
+   html/body transparent (Docusaurus does) then gets a browser-picked DARK
+   canvas behind its light-themed text for dark-preference visitors. Instead,
+   scope it: the editor mount (and any [data-theme] subtree) declares its own
+   scheme so native controls/scrollbars adapt, and the webapp sets the
+   document-root color-scheme at runtime (useThemeStore.applyThemeToDocument). */
+.apollon-editor,
+[data-theme="light"] {
+  color-scheme: light;
+}
 
+:root {
   /* ── Layer 0: device safe area ─────────────────────────────────────────── */
 
   /* The SINGLE declaration of the safe-area insets — every consumer (the editor's


packages/ui/src/styles/tokens.css
```diff
@@ -12,9 +12,20 @@
...
-  color-scheme: light dark;
+/* color-scheme is deliberately NOT set on bare :root. This stylesheet ships
+   inside @tumaet/apollon/style.css, so a `:root { color-scheme: light dark }`
+   would mutate every HOST page that imports the editor CSS: a host that leaves
+   html/body transparent (Docusaurus does) then gets a browser-picked DARK
+   canvas behind its light-themed text for dark-preference visitors. Instead,
+   scope it: the editor mount (and any [data-theme] subtree) declares its own
+   scheme so native controls/scrollbars adapt, and the webapp sets the
+   document-root color-scheme at runtime (useThemeStore.applyThemeToDocument). */
+.apollon-editor,
```
Claudia-Anthropica, line 23: @FelixTJDietrich [medium] `.apollon-editor` overrides the inherited scheme from a dark outer mount. Both public APIs put `data-theme="dark"` on the mount and render `.apollon-editor` below it, so the tokens stay dark while native controls and scrollbars compute `color-scheme: light`; make the nested editor retain the dark scheme and cover this DOM shape with a regression test.

In `packages/ui/src/styles/tokens.css`, the unconditional `.apollon-editor` light scheme overrides the inherited dark scheme when `data-theme="dark"` is on an ancestor mount. Adjust the selectors so default editors are light but editors inside a dark-themed mount compute `color-scheme: dark`, and add a regression test covering the nested DOM shape used by both Apollon APIs.

</details>

.changeset/light-hosts-stay-light.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Stop mutating the host page's color scheme. `style.css` declared `color-scheme: light dark` on `:root`, so any host that imports the editor CSS and leaves its own `html`/`body` backgrounds transparent (the Docusaurus docs site does) rendered a browser-picked dark canvas behind light-themed text for dark-preference visitors. The declaration is now scoped to the editor mount and `[data-theme]` subtrees; host pages keep full control of their document root.
```
Claudia-Anthropica, line 5: @FelixTJDietrich [low] This changes visible page and editor theming, but the PR has no before/after screenshots. Please attach real screenshots of the rebuilt docs page under light and dark preference, plus the dark editor state, as required for visible UI changes.

The visible behavior documented in `.changeset/light-hosts-stay-light.md` lacks the required screenshots. Follow the repository screenshot procedure to capture authentic before/after renders of the rebuilt docs page and dark editor, then attach them to PR #824; do not use mock or synthetic images.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The host-page scoping is the right direction, but the new light rule overrides dark editor mounts, so native controls and scrollbars no longer follow the dark theme. Please fix that cascade and attach the required visual evidence; details inline.

PR #826, opened 2026-07-18
feat(library): improve automatic edge routing

FelixTJDietrich:
### Summary

Automatic edges now coordinate endpoint sides, connection points, and orthogonal paths to produce cleaner diagrams with fewer avoidable bends, overlaps, and crossings. Pinned endpoints and hand-placed bends remain authoritative, while nearby automatic edges can still rebalance around them.

The change also makes routing responsive and consistent across the product:

- the first solve and diagrams with fewer than 32 edges remain synchronous; later dense solves use a backpressured Web Worker that retains only the newest unserialized snapshot and adapts its preview cadence to observed throughput;
- drag previews project or reuse accepted geometry while the latest exact result is computed, and stale Worker results cannot replace settled geometry;
- line jumps follow the routes actually on screen, so bridge arcs stay attached to their edge during a drag and at the preview-to-settled handoff instead of floating over routes that already moved;
- replacing the model increments a routing epoch that discards any in-flight solve for the previous model, so a fresh diagram can never briefly flash the old one's routes;
- on load and on model replacement, edges stay unmounted until the first exact generation is measured, removing the brief straight-then-relayout flash;
- releasing a drag prioritizes the newest exact snapshot, then transitions from the preview to the accepted route unless reduced motion is requested;
- SVG export waits for accepted routing geometry instead of a fixed delay;
- required-interface sockets, package bounds, shared junctions, and legacy diagram imports use the same geometry contracts;
- node and edge action toolbars use the existing React Flow toolbar primitives with accessible, full-size controls;
- the five bundled design-pattern templates start from reviewed automatic layouts;
- server-side consumers can normalize diagrams through the DOM-free `@tumaet/apollon/model` entry.

### Release note

**Library:** Create cleaner diagrams with automatic edges that balance connection points, reduce collisions with nodes and nearby edges, and stay responsive while you drag, without losing pinned endpoints, hand-placed bends, aligned component interfaces, or compatibility with older diagrams; embedding apps can also normalize diagrams in server-side code through the new `@tumaet/apollon/model` entry.

**Web app:** Start every bundled design-pattern template in a clean, visually balanced editing state, with stable shared hierarchy trunks and automatic routing preserved for associations.

### Implementation notes

Routing is a bounded, deterministic pipeline rather than one global optimization:

| Stage | Responsibility |
| --- | --- |
| Node-local coordination | Propose sides and balanced connection-point ordering |
| Joint endpoint/path search | Choose endpoint candidates and an obstacle/edge-aware orthogonal path |
| Route-set repair | Try bounded ordering variants for interacting edges and keep strict improvements |
| Rendering | Preserve manual topology, project live previews, and publish accepted geometry |

Routing authority remains explicit:

- no custom anchors or points: automatic routing owns the edge;
- `sourceAnchor` or `targetAnchor`: that endpoint remains pinned while generated geometry adapts;
- non-empty points: the authored bend topology remains authoritative.

Search work is capped to protect interaction latency. If a joint search exceeds its bound, the solver retries the selected endpoint pair on a smaller graph and then uses deterministic obstacle-derived fallbacks ranked by hard crossings first. This is a deliberate degradation path, not a claim that every pathological diagram has a globally optimal route.

For dense interactive diagrams, the main thread owns one latest snapshot and the Worker owns at most one in-flight solve. Serialization is deferred to a later task and occurs only when the Worker can accept work, avoiding both an obsolete request queue and main-thread cloning on every pointer frame. Preview cadence adapts between 40 and 160 ms from measured round-trip time. Independent input, dispatch, and acceptance revisions verify that release settles the latest authored state.

The `/model` export has its own documentation, Node 22 smoke test, complete runtime-graph check, declaration isolation check, and 8 kB size budget. Its current normalization chunk is 4.73 kB compressed and does not pull React or XYFlow into server code.

Bundle budgets pass, but the main entry and routing Worker are intentionally close to their limits: 116.97/118 kB and 38.38/40 kB respectively. The export entry is 413.41/700 kB.

### Steps for testing

1. Create several automatic edges on one node side and confirm their connection points spread into balanced gaps.
2. Pin one endpoint, then move its node and nearby nodes; the pin should stay fixed while the generated route adapts.
3. Add and drag a manual bend; its topology should remain authoritative without a jump on release.
4. Exercise a diagram with at least 32 edges and drag continuously; incident edges should follow immediately, other routes should update during the gesture, and the newest exact route should settle after release.
5. Repeat the dense drag with reduced motion enabled; the accepted route should appear without a settlement animation.
6. Open component, deployment, and package diagrams; verify interface sockets and package connections meet the visible shapes correctly.
7. Import representative v3 and early-v4 diagrams, create each bundled design-pattern template, and export a diagram as SVG.
8. In Node, import `importDiagram` from `@tumaet/apollon/model` and normalize a saved model without browser globals.

Validation completed on `60bfaf92`:

- `pnpm lint && pnpm format:check && pnpm build && pnpm test`
- library tests: 1,724 passed, 1 skipped
- production-built edge-routing performance suite: Chromium 4/4 and Firefox 4/4
- repeated continuous-drag freshness run: Chromium 3/3
- focused Worker and execution-equivalence tests: 25/25
- `pnpm knip`
- `pnpm dlx publint` from `library/`
- `pnpm --filter @tumaet/apollon size`
- `node scripts/check-model-entry.mjs`
- `node scripts/check-doc-snippets.mjs` (39 checked examples)

### Screenshots / screencasts

Flowchart routing:

Flowchart routing

Component routing with required-interface sockets:

Component routing

Bridge template:

Bridge template

### Checklist

- [x] Linked to a related issue — not applicable; no issue was identified
- [x] Added user-voice changesets for the library and web app
- [x] PR title's `feat` type matches the primary user-visible change
- [x] Tests added or updated
- [x] Documentation updated
- [x] Screenshots attached


library/lib/hooks/useConnect.ts
```diff
@@ -169,6 +183,9 @@ export const useConnect = () => {
 
   const onConnect = useCallback(
     (connection: Connection) => {
+      // Withdraw the preview edge as the real one lands, so the solver never holds
+      // both for a frame (which would fan them apart, then collapse — a flicker).
+      setPendingConnectionEdge(null)
```
Claudia-Anthropica, line 191: @FelixTJDietrich [medium] The preview ID is minted specifically to preserve lane ordering, but this valid `onConnect` path still builds the committed edge with a fresh `generateUUID()` below. Parallel native-handle connections can therefore re-lane as soon as the ghost is released; use `pendingConnectionId.current` here just like the freeform commit path.

In `library/lib/hooks/useConnect.ts`, the valid `onConnect` path discards the ID minted for the pending preview and generates a different committed edge ID. Commit with `pendingConnectionId.current` when available, with `generateUUID()` only as a fallback, and add a test that the preview and committed edge share an ID.

</details>

library/lib/components/ReconnectConnectionLine.tsx
```diff
@@ -100,66 +132,178 @@ export const ReconnectConnectionLine = ({
...
+  // WITHIN one node does not thrash the solver.
+  const isNewConnection = !previewEdge || !reconnectPreviewHandleType
+  const pendingEdge = useMemo<Edge | null>(() => {
+    if (!isNewConnection || !fromNodeId || !newConnection.targetId) return null
+    return {
+      id: pendingEdgeId,
+      source: fromNodeId,
+      target: newConnection.targetId,
+      type: previewEdgeType,
+      data: { points: [] },
```
Claudia-Anthropica: @FelixTJDietrich [medium] Straight-hook edges resolve geometry from `sourceHandle`/`targetHandle`, but this transient edge drops both while the valid commit keeps the actual connection handles. In use-case, syntax-tree, and Petri-net diagrams the solver ghost can therefore attach at different points from the edge that appears on release.

In `library/lib/components/ReconnectConnectionLine.tsx`, the pending preview edge omits handle IDs even though straight-hook routing depends on them. Populate its source and target handles from the connection-line props or the same resolved handles used by commit, and add a straight-hook preview-versus-commit geometry test.

</details>

library/lib/utils/geometry/edgeGeometrySolver.ts
```diff
@@ -0,0 +1,1019 @@
...
+const cellKey = (cx: number, cy: number): string => `${cx},${cy}`
+
+/** Bucket `edgeId` under every cell its polyline has a vertex in. */
+function indexRoutePolyline(
+  grid: NeighborGrid,
+  edgeId: string,
+  polyline: readonly IPoint[]
+): void {
+  let lastKey: string | null = null
+  for (const p of polyline) {
```
Claudia-Anthropica: @FelixTJDietrich [high] Indexing only vertices makes a long segment invisible whenever it crosses the query box with both endpoints outside it; the later `polyline.some(...)` filter has the same flaw. A later edge can then overlap or cross that segment without ever receiving it as a neighbor.

In `library/lib/utils/geometry/edgeGeometrySolver.ts`, `indexRoutePolyline` and `collectNeighbors` detect only vertices rather than segments. Index every grid cell traversed by each orthogonal segment and use segment-versus-box intersection for the exact range check, with a regression test for a two-point segment spanning the entire query box.

</details>

library/lib/utils/geometry/obstacles.ts
```diff
@@ -0,0 +1,265 @@
...
+  let h = nodes.length
+  for (const node of nodes) {
+    const { width, height } = nodeSize(node)
+    h =
+      (Math.imul(h, 31) +
+        (node.position.x | 0) * 7 +
+        (node.position.y | 0) * 13 +
+        (width ?? 0) * 17 +
+        (height ?? 0) * 19 +
+        (node.hidden ? 1 : 0)) |
```
Claudia-Anthropica: @FelixTJDietrich [medium] Despite the comment above, this key omits node IDs, types, and `parentId`, then compresses geometry into a collision-prone 32-bit weighted sum. `indexNodes` can consequently return a previous global `byId`/softness/ancestor index after a model switch, reparent/type change, or hash collision, so routing uses stale obstacles.

In `library/lib/utils/geometry/obstacles.ts`, `geometryFingerprint` does not uniquely cover every field consumed by `indexNodes`. Replace the lossy numeric hash with an exact stable key or verify equality across IDs, types, parent IDs, hidden state, positions, and measured sizes before reusing the cache, and add cache-invalidation tests.

</details>

library/lib/utils/geometry/edgeGeometrySolver.ts
```diff
@@ -0,0 +1,1019 @@
...
+  targetCustom: FreeformEdgeAnchor | undefined,
+  enableStraightPath: boolean,
+  thirdPartyObstacles: readonly ObstacleRect[],
+  neighborEdges: readonly IPoint[][]
+): string {
+  const e = endpoints
+  const anchor = (a: FreeformEdgeAnchor | undefined) =>
+    a ? `${a.side},${a.ratio}` : "-"
+  return [
+    enableStraightPath ? "1" : "0",
```
Claudia-Anthropica, line 762: @FelixTJDietrich [medium] This cache key omits routing-relevant edge type/marker padding, while `routeInputSignature` also omits the chosen endpoints themselves. Changing a component/deployment edge between provided and required interface styles can therefore hit both caches and reuse the old polyline even though the required-interface socket changes target padding.

In `library/lib/utils/geometry/edgeGeometrySolver.ts`, auto-route cache signatures omit marker padding and adjusted endpoints. Include all routing-relevant edge-style inputs in the anchor signature and serialize the resolved endpoint coordinates, positions, and padding in `routeInputSignature`; add a test that changing an edge type with different marker padding invalidates the cache.

</details>

library/package.json
```diff
@@ -94,6 +94,7 @@
   ],
   "peerDependencies": {
     "@xyflow/react": "^12.9.0",
+    "@xyflow/system": ">=0.0.78",
```
Claudia-Anthropica: @FelixTJDietrich [medium] The exact enforced command `pnpm --filter @tumaet/apollon size` now fails: the main entry is 114.28 kB against the 87 kB limit. `@xyflow/system` is also declared as a peer here but is absent from the Vite external list, so the published entry bundles its own copy instead of using that peer.

In `library/package.json` and `library/vite.config.ts`, make the published main entry pass its size budget. Externalize `@xyflow/system` consistently with its peer declaration, then reduce the remaining bundle growth or deliberately update the reviewed budget so `pnpm --filter @tumaet/apollon size` passes.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The central routing approach and coverage are ambitious, but the current integration can change routes between preview and commit, and several caches can reuse stale geometry. The published bundle also exceeds the enforced size limit; details are inline.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich This remains vertex-only in the current head. Long segments that cross a query box with both endpoints outside it are still absent from the neighbor set.

library/package.json
Claudia-Anthropica: @FelixTJDietrich Raising the budget addresses the failing size gate, but the peer dependency issue remains: `@xyflow/system` is still absent from Vite externals and is present inside `dist/index.js`.

library/lib/utils/geometry/obstacles.ts
Claudia-Anthropica: @FelixTJDietrich This is still unchanged: the 32-bit fingerprint still omits IDs, types, and parent IDs while `indexNodes` consumes those fields.

library/lib/hooks/useConnect.ts
Claudia-Anthropica: @FelixTJDietrich This is still unchanged in the current head: `onConnect` commits a fresh UUID instead of `pendingConnectionId.current`, so preview and commit can still re-lane.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich This is still open: the anchor signature still omits marker padding, and route reuse can still ignore changed resolved endpoint geometry.

library/lib/components/ReconnectConnectionLine.tsx
Claudia-Anthropica: @FelixTJDietrich This is still open: the pending preview edge still omits `sourceHandle` and `targetHandle`, so straight-hook preview geometry can differ from the committed edge.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The new commit only raises the size budget; the five correctness issues and the `@xyflow/system` externalization concern remain open in the existing threads. Requesting changes until those are addressed.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The shaped-node drop anchoring now looks consistent, but the six existing review threads are still unresolved in this head. The preview/commit identity and handle mismatches, route-neighbor indexing, cache invalidation, and `@xyflow/system` externalization still need to be addressed.

FelixTJDietrich:
Thanks for the thorough pass — all six threads are addressed. Summary, newest commits at the top:

**1. `@xyflow/system` bundled + size gate** (`d4224857`) — It was a declared peer but missing from the Vite externals, so the entry bundled a private copy. Externalized it (the solver imports `getEdgePosition` from it, and it shares module state with the external `@xyflow/react`). That drops the main entry **114 → 101 kB** brotli, so I lowered the stopgap budget from 118 to **105 kB** (new baseline + headroom). `pnpm --filter @tumaet/apollon size` passes.

**2. Neighbour detection was vertex-only** (`dc0a1d98`) — Confirmed: a long run crossing a query box with both ends outside it was missed by both the grid and the exact filter. Now every cell a *segment* traverses is bucketed, and the precise filter does segment-vs-box (exact for orthogonal routes). Regression tests added for a two-point segment spanning the box with both ends outside, incl. a mid-polyline run.

**3. Obstacle cache fingerprint omitted identity** (`228403fa`) — Confirmed: `geometryFingerprint` didn't cover id/type/parentId, all of which `indexNodes` reads (byId, the soft flag, absolute positions + ancestors). Folded them in; added a test that a type flip with unchanged geometry re-indexes.

**4. `onConnect` re-mints the edge id** (`5f7dcf06`) — Confirmed: the valid handle-drop path used a fresh UUID instead of `pendingConnectionId.current`, so parallel native-handle connections could re-lane on release. Now reuses the preview id.

**5. Pending preview edge dropped its handles** (`5f7dcf06`) — Confirmed: straight-hook types resolve endpoints through `getEdgePosition`, which reads the handles. The preview now carries the same source/target handles the commit assigns.

**6. Auto-route cache key omitted the endpoints / marker padding** (`dc0a1d98`) — Two parts, and I want to be precise:
   - *Soundness:* `routeInputSignature` now keys on the resolved endpoints (adjusted coords, exit positions, padding), matching `routeSignature`, so the key covers every router input rather than leaving the endpoints implicit in the reach corridor.
   - *Observable staleness:* I could not reproduce a stale **route** from a provided→required interface switch. Empirically the solved polyline is padding-independent — the socket gap is applied as a render-time marker offset, not baked into the route (verified: both types produce an identical polyline, pinned or auto). So the previously-cached polyline was not visibly stale; the change is a key-soundness hardening, not a fix for a wrong drawing. Happy to add the offset to the anchor signature too if you'd prefer, but the anchor *pick* is padding-independent by design, so I left it out to avoid needless re-picks.

Full suite green locally: lint, format, build, 1503 unit tests, and the connection/shape-aware e2e.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks. Segments are indexed across every traversed cell and the exact range check is segment-based.

library/lib/components/ReconnectConnectionLine.tsx
Claudia-Anthropica: @FelixTJDietrich Nice fix — the pending edge now carries the same source and target handles as the commit.

library/lib/utils/geometry/obstacles.ts
Claudia-Anthropica: @FelixTJDietrich This still uses a single 32-bit hash as proof of equality, so the collision issue remains. For example, Java-style polynomial hashes collide for IDs `Aa` and `BB`; with otherwise identical nodes, this can reuse the old `byId` map. Please use an exact key or verify all fields before returning the cached index.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich The route key now covers adjusted endpoints, but `autoAnchorSignature` still omits the edge type/marker padding. A style change can therefore reroute the previously cached anchors without re-running anchor selection, even though padding participates in candidate scoring. Please include the routing style or resolved padding in the anchor-selection signature as well.

library/package.json
Claudia-Anthropica: @FelixTJDietrich Nice, `@xyflow/system` is externalized now and the updated size gate is green.

library/lib/hooks/useConnect.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks — the native-handle commit reuses the preview ID.

Claudia-Anthropica (changes requested):
@FelixTJDietrich Four original threads are resolved, but two cache-invalidation issues remain. The node index still trusts a collision-prone 32-bit hash, and the auto-anchor signature still omits routing style/marker padding, so requesting changes until those are addressed.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commits improve short-edge handles and interface layout, but they do not address the two unresolved cache-invalidation threads. The node-index fingerprint and auto-anchor signature still need exact invalidation before this is ready.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The new endpoint-handle overlap fix looks good, but the two existing cache-invalidation threads remain open. Codacy also reports two stylesheet violations on this head, so changes are still requested.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich This also remains unchanged: `autoAnchorSignature` still omits `endpoints.padding`, even though `resolveEdgeEndpoints` derives it from the edge marker style and anchor selection routes/scorers consume it. Please include the resolved padding in this signature.

library/lib/utils/geometry/obstacles.ts
Claudia-Anthropica: @FelixTJDietrich This is still present in the current head: `indexNodes` returns the cached maps solely when the 32-bit fingerprint matches, so collisions such as `Aa`/`BB` can still reuse another node set. The recent commits do not touch this implementation.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest routing and handle fixes do not address the two unresolved cache-invalidation threads: node-index fingerprint collisions and missing marker padding in the anchor cache key. Codacy also remains failing on this head, so changes are still requested.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two latest commits only adjust endpoint-handle rendering and edge cleanup, so both cache-invalidation threads remain unchanged: the node index still trusts a collision-prone 32-bit fingerprint, and auto-anchor selection still omits marker padding from its signature. Codacy is also currently failing; the remaining checks passed.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two existing cache-invalidation findings are still unresolved, and Codacy is currently failing on this head. Requesting changes until the open threads are addressed and the failing check is cleared.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest handle-overlap fixes do not address the two unresolved cache-invalidation threads: the node-index fingerprint remains collision-prone, and auto-anchor selection still omits marker padding. Codacy is also failing on this head, so changes are still requested.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest corner-routing commit does not address the two open cache-invalidation threads: the node index still trusts a collision-prone 32-bit fingerprint, and auto-anchor selection still omits resolved marker padding. Codacy is also failing on this head, so changes are still requested.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit only reverts the corner-routing experiment, so the two cache-invalidation threads remain open: the node-index fingerprint is still collision-prone, and auto-anchor selection still omits resolved marker padding. Codacy is also currently failing on this head.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit only changes corner-overshoot behavior, so the two existing cache-invalidation threads remain unresolved. Codacy also fails on the current head; requesting changes until the open threads are addressed and the failing check clears.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest commit only changes corner-overshoot behavior, so the two cache-invalidation threads remain unresolved: the node index still relies on a collision-prone 32-bit fingerprint, and auto-anchor selection still omits resolved marker padding. Codacy is also failing on this head, so changes are still requested.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The newer corner-anchoring work looks focused, but the two cache-correctness issues from the prior review are still open: obstacle-index reuse still relies on a collision-prone 32-bit fingerprint, and anchor selection still does not key marker padding. Please address those existing threads before this can be approved.

library/lib/utils/geometry/obstacles.ts
FelixTJDietrich: Fixed in c8b2f7c4 — but by making the key exact rather than by widening the hash. The deterministic Aa/BB-style collision you note can't arise here (node IDs are UUIDs, not attacker- or user-controlled short strings), and a random 32-bit collision between two frame-adjacent node sets self-corrects on the next frame — so this was low-severity. That said, a digest is never proof of equality, and there's a strictly-better option: both callers (getEdgeObstacles, getContainerBorderPolylines) receive the SAME nodes array reference for every edge in a solver pass, and the store publishes a new array on any change (React/Zustand never mutate in place). So the key is now the array reference itself — exact, no collision surface, and cheaper than the per-node digest (no hashing loop). A new-but-identical array on a later pass just rebuilds once, which is always correct.

library/lib/utils/geometry/edgeGeometrySolver.ts
FelixTJDietrich: Confirmed and fixed in c8b2f7c4. You're right that selection consumes padding: selectEdgeAnchors resolves each candidate via resolveEdgeEndpoints, which sets an auto endpoint back by endpoints.padding (from getEdgeMarkerStyles(edge.type)) before the candidate route is scored — so two edge types with identical node rects but different markers could pick different anchors while colliding on the same autoAnchorSignature entry. Padding is now folded into the key, matching routeInputSignature, which already included it. Note the impact was bounded (padding shifts an auto endpoint by a few px, so it only occasionally flips a pick, and only until another field invalidated the entry) but the omission was a genuine soundness gap, so this is the correct fix.

Claudia-Anthropica: @FelixTJDietrich Nice, the resolved padding is now part of `autoAnchorSignature`, so that cache follows the scorer input. Looks good.

library/lib/utils/geometry/obstacles.ts
Claudia-Anthropica: @FelixTJDietrich This is still unsafe for the main solver caller. `EdgeGeometrySolver` reads React Flow `s.nodes` and explicitly documents that React Flow keeps that array reference stable while mutating measured geometry in place; `nodeGeometryKey` then reruns the layout effect, but `indexNodes` now returns the pre-measurement index solely because the reference is unchanged. The preceding cache comment in `obstacles.ts` describes this exact failure. Please use an exact content key/equality check and add a same-array in-place measurement regression test.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The marker-padding cache key is fixed, but the obstacle-cache thread remains open: React Flow can mutate measured node geometry while keeping the `nodes` array reference stable, so reference-only reuse can serve the pre-measurement index. Please use exact content invalidation and cover that same-array mutation path.

library/lib/utils/geometry/obstacles.ts
Claudia-Anthropica: @FelixTJDietrich Nice fix — the cache is now solve-scoped, and the same-array measurement regression passes locally. Looks good.

Claudia-Anthropica (changes requested):
@FelixTJDietrich All previous review feedback is addressed, and the focused obstacle-cache regression passes locally. Codacy Static Code Analysis is still failing on this head, so I’m keeping the review at changes requested until that check is green.

FelixTJDietrich:
@Claudia-Anthropica Codacy and the complete hosted suite are now green on `4617db5f` (24 successful checks, zero failures). The final follow-up also fixed the CI production-probe check at its root by restricting counter-name discovery to the `PerfCounters` type, and addressed Codacy's sole CSS finding. Re-review requested when convenient.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed, and the final Codacy/CI fixes are green. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the new template-preparation and legacy-import compatibility changes are covered by green CI. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused edge-geometry solver suite passes on the shared-junction follow-up. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback is addressed, and the focused routing regressions plus bundle-size gates pass locally. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the refreshed template-picker baselines look consistent across both themes. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused required-interface routing and marker suite passes on the latest follow-up. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused interface geometry regressions pass with the refreshed component/deployment baselines. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused straight-routing and required-interface marker regressions pass on this reconnect follow-up. Nice work.

library/lib/components/ReconnectConnectionLine.tsx
Claudia-Anthropica: @FelixTJDietrich This invariant regressed in the replacement `ConnectionPreviewLine`: `ConnectionLineComponentProps` exposes the exact native `toHandle`, but the component ignores it and derives `targetHandle` only from `toPosition`. A valid drop onto a non-central handle such as `top-left` therefore previews under `top` and commits under `top-left`, so straight-hook edges can jump on release. Please use `toNode`/`toHandle` for native targets, falling back to freeform resolution only when they are null, and cover a non-central target handle.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The original cache and bundle feedback remains resolved, and the focused routing tests, library build, and size gates all pass. However, the replacement connection preview regresses the exact native target-handle handoff for non-central handles, so straight-hook edges can still jump on commit; I reopened the existing thread with details.

library/lib/components/ReconnectConnectionLine.tsx
Claudia-Anthropica: @FelixTJDietrich The exact native handle is carried now, but this thread is still open because `toNode`/`toHandle` can describe an invalid React Flow target candidate. Treating every such pair as native bypasses the continuous freeform anchor for use-case ovals; the required E2E fails identically on both attempts with the committed endpoint 8.54 px from the snap circle. Gate the native handoff on a valid/connectable target and keep the freeform fallback for invalid candidates.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The exact native handle is now preserved, but the new native-target gate also claims invalid target candidates and breaks the continuous use-case fallback. The required E2E fails deterministically on both attempts because the committed endpoint is 8.54 px from the snap circle, so native handoff needs validity/connectability gating before this is ready.

library/lib/components/ReconnectConnectionLine.tsx
Claudia-Anthropica: @FelixTJDietrich Nice fix — native handoff is now gated on a valid connection, and the focused use-case ghost/commit regression passes locally. Looks good.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback is addressed, and the focused use-case preview/commit plus package geometry regressions pass locally. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the shared toolbar follow-up builds cleanly with all seven focused interaction tests passing. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused worker execution, DOM-free model entry, size gates, and reduced-motion release regression pass on this head. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused routing-generation store suites pass on the first-exact-generation follow-up. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the three cleanup/fix commits leave those invariants intact with green CI. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the focused tag-picker test passes with the restored paint and tag trigger tooltips. Nice work.

Claudia-Anthropica (approved):
@FelixTJDietrich All previous feedback remains addressed, and the Worker-settlement E2E passes with the gentler polling. The remaining Node 22 failure is an unrelated 504 while downloading libvips. Nice work.

PR #827, opened 2026-07-19
fix(library): pin editor chrome and stop repainting host borders

FelixTJDietrich:
### Summary

Two bugs that only appear when Apollon is embedded in someone else's page. Our own docs homepage was the first host to hit both, which is how they surfaced — but they ship to every consumer of `@tumaet/apollon`.

**1. The zoom / undo / minimap controls drifted into the middle of the diagram.**

`useKeyboardInset` measured the distance from the canvas's bottom edge to the bottom of the window and treated it as on-screen-keyboard overlap. That distance is positive for *any* editor mounted below the fold — a docs page, a long form, anything the reader scrolls to — so the editor reserved that entire gap as padding. `visualViewport` emits no `scroll` event for ordinary layout-viewport scrolling ([csswg-drafts#8205](https://github.com/w3c/csswg-drafts/pull/8205)), so the bogus value measured at mount never cleared, even once the editor was fully in view.

The overlap is now clamped to the visible viewport, so it can never exceed the keyboard's own height.

**2. Importing the editor's CSS repainted the host page's borders.**

`components.css` applied a `border-color` reset to `*`. That rule ships inside the published `style.css`, so it reached host DOM — where `--color-border` is undefined. An unresolvable `var()` is invalid at computed-value time, which makes the property behave as `unset`; `border-color` doesn't inherit, so it falls back to its initial `currentColor`. Host inputs, selects, buttons and fieldsets take their border colour from the UA stylesheet, and *any* author rule outranks the UA sheet — so they lost their neutral grey and picked up the surrounding text colour.

Scoped to our own `[data-slot]` primitives, matching the `box-sizing` floor directly beside it. Same class of bug as #824.

Alongside those, the docs homepage gets what it needed to actually demonstrate the product: the live embed is seeded with a real diagram instead of an empty canvas, and the hero is tightened so the editor sits nearer the fold.

### Release note

Two entries, both `@tumaet/apollon` patches:

- Keep the zoom, undo and minimap controls pinned to the bottom of the canvas in embedded editors.
- Stop repainting the host page's borders when the editor's `style.css` is imported.

### Implementation notes

**Why a clamp rather than an "is the keyboard open?" check.** Gating on "the visual viewport is shrunk" looks equivalent and isn't: with a keyboard actually open on a below-the-fold editor it still reserves the whole overhang (800px for a 300px keyboard), and sub-pixel rounding between `clientHeight` and `visualViewport.height` reopens the original bug. Clamping `host.bottom` to the visible viewport bounds the result by construction. The unit test asserts both cases and was checked against both incorrect formulas.

**A dead end worth recording, so nobody re-treads it.** The stranded chrome looks exactly like a CSS Grid `fr` failure, and an earlier attempt at this fix "corrected" grid track sizing in JavaScript. That was wrong: an absolutely-positioned `inset: 0` grid has a *definite* height ([css-sizing-3 §4.1](https://www.w3.org/TR/css-sizing-3/)) and `fr` resolves correctly here. Neutralising the keyboard inset alone makes the centre track fill properly with no grid changes at all. The grid is untouched by this PR.

**CI gap closed.** `packages/**` was missing from every paths-filter, so changes to `@tumaet/ui` triggered no checks at all — even though api-extractor bundles it into the published library (`bundledPackages: ["@tumaet/ui"]`). Without this, the new leak guard could not fire on a PR that only touched the package it guards. It's now in the `code`, `library` **and** `vscode` filters; the last of those came out of review — `vscode-extension-checks` runs `pnpm run build:lib` and typechecks the extension against the declarations api-extractor inlines `@tumaet/ui` into, so a design-system-only PR must not skip it.

**Tests.** Both new guards were mutation-tested rather than assumed: each was re-run against the exact code it replaces and confirmed to fail. The CSS guard is an allowlist (every selector must anchor to `[data-slot]` or `.apollon-`) rather than a denylist of known-bad selectors, because a denylist waves through `input`, `button` and `::selection` — the very selectors that would reproduce this bug.

**Something I removed rather than shipped.** An earlier revision of this branch added a `docs/.browserslistrc` to stop `@csstools/postcss-cascade-layers` rewriting unlayered rules with `:not(#\#)`, which lets Infima's `:root` outrank `custom.css` and renders the brand blue as `#3578e5`. On checking the *production* bundle it turned out that polyfill only runs under `docusaurus start`: the built site already emits zero `:not(#\#)` and renders `#3e8acc` correctly. The file changed production browser targets (including SWC's JS output) to fix a dev-server-only cosmetic difference, so it's gone. The dev/prod divergence is real but harmless and out of scope here; `configurePostCss` is the surgical lever if we ever want to close it.

**Deliberately not done.** The `heroModel.json` fixture is intentionally not shared with the webapp's visual-regression fixture; they are independent artefacts and should be free to diverge, and reaching from docs source into another workspace's `tests/` would be worse. The pre-existing pinch-zoom (`visualViewport.scale`) edge case is unchanged.

**Known limitation.** The keyboard path is covered by unit tests and by the spec, but I have not verified it on physical iOS or Android hardware. That's the one thing I'd want a device (or a second pair of eyes) on before release.

### Steps for testing

1. `pnpm install && pnpm run build:docs && pnpm --filter @tumaet/docs exec docusaurus serve`
2. Open the homepage and scroll the live editor fully into view. The zoom, undo and minimap controls sit flush against the bottom edge of the canvas, and the editor shows a class diagram rather than an empty grid.
3. Scroll to **What you get** — the cards have a blue top accent and light grey sides, not near-black.
4. Confirm the guards actually bite:
   - Re-widen the selector in `packages/ui/src/styles/components.css` back to `*` → `pnpm --filter @tumaet/ui exec vitest run tests/hostLeak` fails.
   - Remove the clamp in `library/lib/overlay/OverlayLayer.tsx` → `pnpm --filter @tumaet/apollon exec vitest run tests/unit/overlayApi` fails.
5. `pnpm lint && pnpm format:check && pnpm build && pnpm test`

### Screenshots / screencasts

**Editor chrome.** The controls sat stranded in the middle of the canvas over an empty diagram; they now sit flush against the bottom edge, with the embed seeded so it actually shows a diagram.

| Before | After |
| --- | --- |
| Docs homepage embed with the zoom, undo and minimap controls floating in the middle of an empty canvas | Same embed with the controls flush to the bottom edge and a seeded class diagram |

**Host borders.** The same regression inside our own docs: the cards took the editor's `*` border reset and went near-black. They now show the intended blue top accent and light grey sides.

| Before | After |
| --- | --- |
| What you get cards outlined in near-black on all four sides | Same cards with a blue top accent and light grey sides |

<sub>Images live on the <a href="https://github.com/ls1intum/Apollon/tree/pr-827-assets"><code>pr-827-assets</code></a> branch so binaries stay out of the code history; delete it once this is merged.</sub>

### Checklist

- [x] Tests added or updated

---

### DIFF
--- a/.github/workflows/pr-health-checks.yml
+++ b/.github/workflows/pr-health-checks.yml
@@ -37,6 +37,7 @@ jobs:
           filters: |
             code:
               - 'library/**'
+              - 'packages/**'
               - 'standalone/**'
               - 'vscode-extension/**'
               - 'docs/**'
@@ -51,6 +52,9 @@ jobs:
               - '.github/actions/**'
             library:
               - 'library/**'
+              # api-extractor bundles @tumaet/ui into the published library, so
+              # a design-system change ships to consumers as a library change.
+              - 'packages/**'
               - 'pnpm-lock.yaml'
               - 'pnpm-workspace.yaml'
               - '.github/workflows/pr-health-checks.yml'
--- a/.github/workflows/pr-health-checks.yml
+++ b/.github/workflows/pr-health-checks.yml
@@ -61,9 +61,12 @@ jobs:
               - '.github/actions/**'
             vscode:
               # The extension bundles the library and typechecks against its
-              # emitted declarations, so a library change can break it.
+              # emitted declarations, so a library change can break it — and
+              # api-extractor inlines @tumaet/ui into those declarations, so a
+              # design-system change reaches the extension the same way.
               - 'vscode-extension/**'
               - 'library/**'
+              - 'packages/**'
               - 'pnpm-lock.yaml'
               - 'pnpm-workspace.yaml'
               - '.github/workflows/pr-health-checks.yml'
--- a/.changeset/chrome-stays-at-the-canvas-edge.md
+++ b/.changeset/chrome-stays-at-the-canvas-edge.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Keep the zoom, undo and minimap controls pinned to the bottom of the canvas in embedded editors. An editor mounted below the fold mistook the gap between its bottom edge and the window for an on-screen keyboard and reserved that whole gap as padding, stranding the controls mid-diagram — and scrolling the editor into view never cleared it. The editor now reserves only the part of the canvas a keyboard actually covers.
--- a/.changeset/host-borders-stay-host-borders.md
+++ b/.changeset/host-borders-stay-host-borders.md
@@ -1,0 +1,5 @@
+---
+"@tumaet/apollon": patch
+---
+
+Stop repainting the host page's borders. Importing the editor's `style.css` reset the border colour of every element on the page, so the host's own inputs, selects, buttons and fieldsets lost the browser's neutral grey and took on the surrounding text colour — near-black on a light page, stark white on a dark one. The reset now applies only to the editor's own controls.
--- a/docs/src/components/ApollonEmbed/heroModel.json
+++ b/docs/src/components/ApollonEmbed/heroModel.json
@@ -1,0 +1,256 @@
+{
+  "id": "fixture-readme-hero-001",
+  "version": "4.0.0",
+  "title": "Apollon",
+  "type": "ClassDiagram",
+  "nodes": [
+    {
+      "id": "9f1d2a30-0000-4000-8000-000000000001",
+      "width": 220,
+      "height": 130,
+      "type": "class",
+      "position": {
+        "x": 40,
+        "y": 55
+      },
+      "data": {
+        "name": "Diagram",
+        "attributes": [
+          {
+            "id": "hero-a1",
+            "name": "- title: string"
+          },
+          {
+            "id": "hero-a2",
+            "name": "- type: DiagramType"
+          }
+        ],
+        "methods": [
+          {
+            "id": "hero-m1",
+            "name": "+ exportAsSVG(): SVG"
+          }
+        ]
+      },
+      "measured": {
+        "width": 220,
+        "height": 130
+      }
+    },
+    {
+      "id": "9f1d2a30-0000-4000-8000-000000000002",
+      "width": 200,
+      "height": 170,
+      "type": "class",
+      "position": {
+        "x": 50,
+        "y": 330
+      },
+      "data": {
+        "name": "DiagramType",
+        "stereotype": "Enumeration",
+        "attributes": [
+          {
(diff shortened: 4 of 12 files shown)


.github/workflows/pr-health-checks.yml
```diff
@@ -51,6 +52,9 @@ jobs:
               - '.github/actions/**'
             library:
               - 'library/**'
+              # api-extractor bundles @tumaet/ui into the published library, so
+              # a design-system change ships to consumers as a library change.
+              - 'packages/**'
```
Claudia-Anthropica, line 57: @FelixTJDietrich [medium] This closes the library-job gap, but a `packages/ui`-only PR still leaves the `vscode` output false, so `vscode-extension-checks` is skipped even though that job builds the library and consumes it in the extension. Add `packages/**` (or specifically `packages/ui/**`) to the `vscode` filter too.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `.github/workflows/pr-health-checks.yml`, package-only changes set `library` but not `vscode`, so `vscode-extension-checks` skips changes bundled into the extension. Add `packages/**` or `packages/ui/**` to the `vscode` path filter as well.\n\n</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The editor fixes and regression tests look solid, but the new package path coverage still skips the VS Code extension job for package-only changes. Please close that remaining filter gap; see the inline comment.

.github/workflows/pr-health-checks.yml
FelixTJDietrich: Good catch — confirmed and fixed in db751f8.

I verified the mechanism rather than just adding the path: `vscode-extension-checks` gates on `outputs.vscode`, but its first step is `pnpm run build:lib`, and `build:lib` builds `@tumaet/ui` before the library — which api-extractor then inlines into the `.d.ts` the extension typechecks against. So a `packages/ui`-only PR would have compiled that exact code in a job it never ran.

`packages/**` is now in the `vscode` filter alongside `library/**`, with the reason recorded in the comment there. I also swept the remaining gates for the same class of gap: `code` and `library` already cover it, and `version-sync-check` is deliberately unfiltered — so all three filters are now consistent.

Claudia-Anthropica: @FelixTJDietrich Nice fix, that closes the VS Code filter gap.

Claudia-Anthropica (approved):

PR #829, opened 2026-07-24
fix(library): keep custom-colored elements visible in exports

### LINKED ISSUE #828: Colored Components/Diagram elements are not correctly exported
### Before you start

- [x] I searched the existing issues and discussions and did not find a duplicate.
- [x] I can reproduce this on the live demo (https://apollon.aet.cit.tum.de) or the latest `@tumaet/apollon`.

### Describe the bug

When I set a different line color for diagram elements (any) and export them to any format, the elements are not visible.

### Steps to reproduce

1. Open e.g. a component diagram
2. Add a component 
3. Change its line color
4. Export as e.g. svg
5. See that only the text not the component itself is visible

### Expected behaviour

Should be render normally in the color that was selected.

### Area

Standalone web app

### Screenshots or recording

_No response_

### Browser

_No response_

### Operating system

_No response_

### Apollon version

_No response_

### Browser console output

```shell

```

FelixTJDietrich:
### Summary

Setting a custom line, fill, or text color on any diagram element and then exporting it (SVG, PNG, PDF, …) dropped the color — the shape vanished and only its default-black text remained. This closes #828.

The color picker stores a chosen palette swatch as a bare CSS variable, `var(--apollon-swatch-red)`, with no inline fallback (so it can re-theme live between light and dark). External/headless renderers can't resolve CSS variables, so the export pipeline pre-resolves them against a static fallback map — and that map had no entry for the swatch tokens. An unresolved variable with no fallback resolves to an empty string, so `stroke`/`fill` became `""` and the outline disappeared. The default-black text survived because its token *was* in the map, which is exactly why only the text showed.

The fix adds the swatch palette to that fallback map. Because it's the token resolution layer, every export format and every diagram type (nodes and edges, stroke and fill) is covered by the same change.

### Release note

Custom element colors (line, fill, text) now show up correctly when you export a diagram, instead of the colored shape disappearing.

### Implementation notes

- **Root cause:** swatches are stored fallback-less (`ColorButtons.tsx`); the compat export path (`exportUtils.ts` → `replaceCSSVariables` → `resolveCSSVariable`) resolved the missing tokens to `""`.
- **Fix:** add the nine `--apollon-swatch-*` light-theme hexes to `CSS_VARIABLE_FALLBACKS` (`library/lib/constants.ts`). This mirrors how the existing collaboration-cursor and assessment palettes already live in that map.
- **Why light values:** the compat export has no theme context and already pins foreground→`#000` / background→`#fff`, so it's uniformly light. A swatch exported from a dark-mode editor renders as its light shade — consistent with the rest of the pipeline, and the element is now visible, which is the reported bug. Theme-faithful exports would be a separate, cross-cutting feature.
- **Drift guard:** the hex is duplicated from `packages/ui/src/styles/tokens.css`. A contract test (`cssVariableContract.test.ts`) now asserts each fallback equals the light `--primitive-swatch-*` it resolves to, so retuning the palette can't silently ship a stale export color. Verified it fails on a one-digit mutation.
- **Out of scope (pre-existing, by design):** web-mode SVG (never used by a shipped export button) intentionally keeps variables unresolved for host theming.

### Steps for testing

1. Open a component diagram (or any diagram type).
2. Add an element, open its style editor (paint-roller), and pick a non-default **line color** swatch.
3. Export as SVG (also worth trying PNG / PDF).
4. Open the exported file — the outline renders in the chosen color. **Before this PR only the text was visible.**
5. Repeat with a custom **fill** color and with an **edge** line color, and try another diagram type, to confirm parity.

Automated: `pnpm --filter @tumaet/apollon test` — see `exportUtils.test.ts` ("swatch color export") and `cssVariableContract.test.ts` ("export must never blank or drift a swatch").

### Screenshots / screencasts

_No visual UI change; the difference is in exported files. A before/after SVG export of a red-outlined component can be attached if helpful._

### Checklist

- [x] Linked to a related issue (if applicable)
- [x] Tests added or updated
- [x] Ran lint / format / typecheck / affected tests locally — green
- [ ] Documentation updated (if applicable) — n/a, swatch tokens already documented in `THEMING.md`
- [ ] Screenshots or screencasts attached (if a UI change) — n/a, no UI change


Claudia-Anthropica (approved):
@FelixTJDietrich The fallback additions match the light-theme token definitions, preserve export-time overrides, and are covered by both the palette contract and DOM export path. Looks good, nice fix.

Issue #282, opened 2022-12-26
`Syntax tree diagram:` Break the lines


Santia-go:
### Is your feature request related to a problem?

The lines overlap the nodes in the syntax tree diagram.
<img width="567" alt="Screenshot 2022-12-26 at 20 05 36" src="https://user-images.githubusercontent.com/88991954/209581542-b4713190-7218-478a-aefa-7346c7132869.png">

### Describe the solution you'd like

Once the line joins two elements, should be possible to generate 2 lines with a joint in the line in order to accommodate these lines.

![image](https://user-images.githubusercontent.com/88991954/209581625-326fc9bc-5552-4e48-b904-fc04b38156db.png)

FelixTJDietrich:
Related to #392

FelixTJDietrich:
Completed through #834, this will be in the next 5.3 minor release. I'm experimenting with auto-routing of edges but will probably defer that part a bit until it really works well.

Issue #128, opened 2020-08-14
Add Auto-Layout Functionality


Garfield96:
### Is your feature request related to a problem?
Large graphs can quickly become unorganized and confusing. Restructuring takes a lot of effort, especially if the graph is still changing. Thus, some editors support the automatic restructuring of graphs, which makes the graph planar (if possible) and minimizes edge bends.

### Describe the solution you'd like
Implement different auto-layout functions. An overview of graph drawing algorithms can be found here:
[http://cs.brown.edu/people/rtamassi/papers/gd-tutorial/gd-constraints.pdf](http://cs.brown.edu/people/rtamassi/papers/gd-tutorial/gd-constraints.pdf)

An example of such functionality is shown here:
[https://www.visual-paradigm.com/support/documents/vpuserguide/1283/28/6047_automaticdia.html](https://www.visual-paradigm.com/support/documents/vpuserguide/1283/28/6047_automaticdia.html)

PR #830, opened 2026-07-24
feat(library): customizable straight edges — waypoints, tidy-tree layout & obstacle routing

### LINKED ISSUE #392: More Flexibility for adjusting paths between elements
### Is your feature request related to a problem?

When I link two diagram elements directly, I don't have any flexibility to reshape my path. 
No breakpoints are created if there is a straight link between elements.
Users should be able to create better paths between elements

### Describe the solution you'd like

Create more breakpoints on connection paths between elements so I can make better paths

### Additional context

<table>
<tr>
<td>
No breakpoint for a straight link
</td>

<td>
How an indirect link gives me breakpoints to reshape
</td>
<td>
How draw.io gives breakpoints for straight links
</td>
</tr>
<tr>
<td>

Screenshot 2024-11-08 at 09 50 07

</td>
<td>

Screenshot 2024-11-08 at 09 50 17

</td>

<td>

Screenshot 2024-11-08 at 09 56 34

</td>
</tr>
</table>

FelixTJDietrich:
### Summary

Brings the **straight-hook** edge regime (UML use-case associations/include/extend/generalization, syntax-tree links, petri-net arcs) to parity with the orthogonal step-edge engine from #826. These edges were dumb 2-point diagonal lines with no bend handles, no auto-layout, and no obstacle avoidance. They now support hand-editing, automatic node avoidance, and — for syntax trees — a one-click tidy layout.

Closes #392. Addresses #282 (both the literal "add a joint" ask and its root cause) and lands a first concrete slice of #128.

Straight-hook edges are diagonal **by UML convention** — this is deliberately *not* "make them orthogonal". It is a parallel straight-polyline regime that shares the engine's contracts (memoryless, deterministic, preview==commit) and cost vocabulary, but not the orthogonal grid-A\* search or the H/V jog machinery.

### Track A — manual waypoints ("bend handles to pin it")

- Draggable free-2D waypoints on every straight-hook edge. A faint **ghost** handle at each segment midpoint materialises a bend once dragged past a threshold; a bend drags in 2D; remove it by double-click or by dragging it back onto the line (continuous near-collinear pruning).
- **Model:** `data.points` holds **interior waypoints only** (JointJS "vertices" model); the route is `[source, ...points, target]` connected by plain diagonal segments. Documented divergence from step edges (which store the full route).
- **Solver (S1):** both manual-points call sites (the port-coordination reservation *and* the main loop) now use a regime-aware diagonal passthrough — no orthogonal reprojection, which would silently break preview==commit.
- **Rendering (S2):** grip/marker orientation follows the terminal segment, not the endpoint chord, so a bent edge's arrowhead and reconnect grips point the right way. Touch-friendly (no hover-only affordance; generous hit targets).
- **Migration:** the v3 `path` (full polyline incl. endpoints, previously inert for these types) is cleared on import so old diagrams don't sprout spurious/double bends.

### Track B — syntax-tree tidy layout (#282)

- A user-triggered **"Tidy tree layout"** reconstructs the hierarchy from `SyntaxTreeLink` edges and repositions nodes so parent→child links no longer overlap siblings — the root-cause fix for #282 (a node-separation problem, not routing).
- Hand-written integer **Buchheim/Walker/Reingold–Tilford** tidy-tree with **variable-width** sibling separation (long labels push siblings apart). Malformed input (forests, cycles, multi-parent) degrades gracefully — never corrupts the diagram.
- One Yjs transaction / one undo step; a no-op on already-tidy or non-syntax-tree diagrams. Exposed as `ApollonEditor.layoutSyntaxTree()` and a SyntaxTree-only builtin control.

### Track D — automatic obstacle-avoiding straight router

- Straight edges now **auto-bend around any node** sitting directly in their path; an unobstructed chord stays straight. User waypoints are honoured as mandatory **checkpoints** (libavoid model).
- Deterministic reduced/tangent **visibility graph + integer A\***. Every decision is integer — `isqrt` (exact floor) for length, integer dot/cross **turn buckets** for the angle penalty; **no `hypot`/`atan2`/float on any decision path**. Canonical vertex/heap keys keep the search **byte-identical across Yjs peers and reloads**. Reuses `polylineConflictCost` and the obstacle corridor.

### Design & bundle

Budgets were **raised deliberately** (they are a guardrail, not a physical ceiling) with headroom, rather than distorting the architecture to fit an arbitrary number: main entry 118→126 kB (measured 122.06), routing worker 40→44 kB (measured 39.95). `/model` and `/export` unchanged.

### Implementation notes

New modules: `utils/geometry/freeWaypoints.ts`, `utils/geometry/integerGeometry.ts`, `utils/geometry/straightPolylineRouter.ts`, `utils/tidyTree/{tidyTree,buildForest,applyTidyLayout}.ts`. The design was pressure-tested from four independent angles (codebase map, OSS/industry survey, principal-engineer review, product devil's-advocate) before implementation; a cost-model attachment-selection track was deliberately cut as subsumed by Track D and determinism-unsafe on the use-case ellipse.

### Steps for testing

1. In a use-case/petri-net/syntax-tree diagram, hover an edge, drag a ghost midpoint → a bend appears and persists; drag it; double-click or drag-to-collinear to remove; reload → identical.
2. Place a node between two connected use cases → the straight edge auto-bends around it; move the node → it re-routes; a clear path stays straight.
3. Pin a waypoint, then let the router run → the pin is respected as a checkpoint.
4. Build a syntax tree with overlapping siblings → **Tidy tree layout** → no link overlaps a node; one undo reverts.
5. Import a v3 use-case/syntax-tree diagram → no spurious bends.

### Tests

Unit: `freeWaypoints`, `integerGeometry`, `straightPolylineRouter` (incl. shuffled-input determinism), `tidyTree`, `tidyForest`, `straightHookRouting` (solver S1 passthrough + Track D obstacle bend + cold-solve determinism), and a v3-migration case. Full suite: **1783 passed / 1 skipped**. `lint`, `format:check`, `build` (all packages), `size`, `knip` all green.

### Checklist

- [x] Added a user-voice changeset for the library
- [x] `feat` type matches the primary user-visible change
- [x] Tests added
- [x] `pnpm lint && pnpm format:check && pnpm build && pnpm test` — green
- [ ] Interactive-gesture E2E coverage (recommended follow-up; the gestures are thin glue over unit-tested pure functions + the solver integration test)

### DIFF
--- a/library/lib/edges/GenericEdge.tsx
+++ b/library/lib/edges/GenericEdge.tsx
@@ -21,6 +21,7 @@ import type { DiagramEdgeType } from "./types"
 import { Assessment } from "@/typings"
 import type { BendHandle } from "@/utils/geometry/bendHandles"
 import { getSegmentGhostHandles } from "@/utils/geometry/freeWaypoints"
+import { isqrt } from "@/utils/geometry/integerGeometry"
 import { isFreeformEdgeAnchor } from "@/utils/edgeUtils"
 import { CANVAS, EDGES } from "@/constants"
 
@@ -39,6 +40,8 @@ export const getHandleScreenScale = (zoom: number): number => {
   )
   return 1 / Math.min(safeZoom, 1)
 }
+
+const isqrtLength = (dx: number, dy: number): number => isqrt(dx * dx + dy * dy)
 
 const useHandleScreenScale = (): number =>
   getHandleScreenScale(useStore((state) => state.transform[2]))
@@ -511,12 +514,12 @@ export const EdgeBendHandle = ({
 }
 
 /**
- * Draggable free-2D waypoint editing for straight (diagonal) edges. Renders a faint
- * "ghost" handle at every route segment midpoint (dragging one past a threshold
- * materialises a bend) and a solid handle on every existing interior waypoint
- * (drag to move, double-click to remove). Every handle carries a generous invisible
- * hit target so it stays grabbable when zoomed out and on touch. This is the
- * straight-edge analogue of `EdgeBendHandle`, which is orthogonal-only.
+ * Bend editing for straight (diagonal) edges — the counterpart to `EdgeBendHandle`,
+ * and deliberately the SAME affordance: an identically styled `edge-bend-handle`
+ * pill that lies along the segment, hidden until the edge is hovered or selected.
+ * A step edge drags a segment along its one free axis; a straight edge has two, so
+ * these handles carry a `move` cursor and drop a waypoint where they are released.
+ * Existing waypoints get the same pill so the whole row reads as one control set.
  */
 export const EdgeWaypointHandles = ({
   route,
@@ -542,58 +545,80 @@ export const EdgeWaypointHandles = ({
   ) => void
 }) => {
   const screenScale = useHandleScreenScale()
-  const ghosts = getSegmentGhostHandles(route)
+  const segments = getSegmentGhostHandles(route)
   const hit = EDGES.WAYPOINT_HIT_TARGET_PX * screenScale
-  const waypointRadius = EDGES.WAYPOINT_HANDLE_RADIUS_PX * screenScale
-  const ghostRadius = EDGES.WAYPOINT_GHOST_RADIUS_PX * screenScale
+  const shortAxis = 10 * screenScale
+  const radius = 6 * screenScale
+
+  /** A step-edge bend pill, rotated to lie along an arbitrary segment. */
+  const pill = (
+    center: IPoint,
+    longAxis: number,
+    angleDeg: number,
+    className: string
+  ) => (
+    <rect
+      className={className}
+      x={center.x - longAxis / 2}
+      y={center.y - shortAxis / 2}
+      width={longAxis}
+      height={shortAxis}
+      rx={radius}
+      ry={radius}
+      transform={`rotate(${angleDeg} ${center.x} ${center.y})`}
+      pointerEvents="none"
+    />
+  )
 
   return (
     <>
-      {ghosts.map((ghost) => (
-        <g
-          key={`ghost-${ghost.segmentIndex}`}
-          className="edge-waypoint-ghost-group"
-        >
-          <circle
-            className="edge-circle edge-waypoint-ghost"
-            cx={ghost.position.x}
-            cy={ghost.position.y}
-            r={ghostRadius}
-            pointerEvents="none"
-          />
-          <rect
-            className="edge-waypoint-ghost-hit"
-            x={ghost.position.x - hit / 2}
-            y={ghost.position.y - hit / 2}
-            width={hit}
-            height={hit}
-            rx={hit / 2}
-            ry={hit / 2}
-            pointerEvents="all"
-            style={{ cursor: "crosshair", fill: "transparent" }}
-            onPointerDown={(event) =>
-              onGhostPointerDown(event, ghost.segmentIndex)
-            }
-          />
-        </g>
-      ))}
+      {segments.map((segment) => {
+        const a = route[segment.segmentIndex]
+        const b = route[segment.segmentIndex + 1]
+        // Degrees are presentation only — never a routing decision — so the
+        // trigonometry here cannot affect committed geometry.
+        const angle = (Math.atan2(b.y - a.y, b.x - a.x) * 180) / Math.PI
+        const room = isqrtLength(b.x - a.x, b.y - a.y) - 2 * radius
+        const longAxis = Math.min(
+          Math.max(room, EDGES.BEND_HANDLE_MIN_SCREEN_LENGTH_PX * screenScale),
+          EDGES.BEND_HANDLE_SCREEN_LENGTH_PX * screenScale
+        )
+        return (
+          <g key={`segment-${segment.segmentIndex}`}>
+            {pill(
+              segment.position,
+              longAxis,
+              angle,
+              "edge-circle edge-bend-handle"
+            )}
+            <rect
+              x={segment.position.x - hit / 2}
+              y={segment.position.y - hit / 2}
+              width={hit}
+              height={hit}
+              rx={hit / 2}
+              ry={hit / 2}
+              pointerEvents="all"
+              style={{ cursor: "move", fill: "transparent", zIndex: 9999 }}
+              onPointerDown={(event) =>
+                onGhostPointerDown(event, segment.segmentIndex)
+              }
+            />
+          </g>
+        )
+      })}
       {interior.map((point, index) => (
-        <g key={`waypoint-${index}`} className="edge-waypoint-group">
-          <circle
-            className={
-              "edge-circle edge-waypoint-handle" +
+        <g key={`waypoint-${index}`}>
+          {pill(
+            point,
+            shortAxis,
+            0,
+            "edge-circle edge-bend-handle" +
               (selectedWaypointIndex === index
-                ? " edge-waypoint-handle--selected"
+                ? " edge-bend-handle--active"
                 : "")
-            }
-            cx={point.x}
-            cy={point.y}
-            r={waypointRadius}
-            style={{ strokeWidth: FREEFORM_ENDPOINT_GRIP_STROKE * screenScale }}
-            pointerEvents="none"
-          />
+          )}
           <rect
-            className="edge-waypoint-hit"
             x={point.x - hit / 2}
             y={point.y - hit / 2}
             width={hit}
@@ -601,7 +626,7 @@ export const EdgeWaypointHandles = ({
             rx={hit / 2}
             ry={hit / 2}
             pointerEvents="all"
-            style={{ cursor: "move", fill: "transparent" }}
+            style={{ cursor: "move", fill: "transparent", zIndex: 9999 }}
             onPointerDown={(event) => onWaypointPointerDown(event, index)}
             onDoubleClick={() => onWaypointDoubleClick(index)}
           />
--- a/library/lib/edges/GenericEdge.tsx
+++ b/library/lib/edges/GenericEdge.tsx
@@ -20,6 +20,7 @@ import {
 import type { DiagramEdgeType } from "./types"
 import { Assessment } from "@/typings"
 import type { BendHandle } from "@/utils/geometry/bendHandles"
+import { getSegmentGhostHandles } from "@/utils/geometry/freeWaypoints"
 import { isFreeformEdgeAnchor } from "@/utils/edgeUtils"
 import { CANVAS, EDGES } from "@/constants"
 
@@ -278,6 +279,8 @@ export const EdgeEndpointMarkers = ({
   onEndpointPointerDown,
   straight = false,
   bendHandles,
+  sourceNeighbor,
+  targetNeighbor,
 }: {
   sourcePoint: IPoint
   targetPoint: IPoint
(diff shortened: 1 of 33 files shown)


library/lib/utils/versionConverter.ts
```diff
@@ -745,6 +746,13 @@ function convertV3RelationshipToV4Edge(
       y: point.y + relationship.bounds.y,
     }))
   }
+  // Straight-hook edges (use-case, syntax-tree, petri-net) render as diagonal lines
+  // through INTERIOR waypoints only. The legacy v3 `path` is the full rendered
+  // polyline including endpoints and was inert for these types, so importing it as
+  // waypoints would sprout spurious (double-endpoint) bends. Clear it.
+  if (STRAIGHT_HOOK_EDGE_TYPES.has(edgeType as string)) {
```
Claudia-Anthropica, line 753: @FelixTJDietrich [high] This only fixes direct v3 imports. Models already converted and saved by current releases are 4.x and can still carry the old full-route `data.points`; this branch still exports `4.1.0`, so those points are indistinguishable from new interior waypoints and will render as bogus bends (and can then be resaved as user data). Version this schema change and normalize older 4.x models too.

In `library/lib/utils/versionConverter.ts`, legacy full-route points on straight-hook edges are cleared only during v3 conversion, while pre-feature 4.x models pass through unchanged. Add a versioned normalization for pre-feature 4.x models and bump the exported model version so new interior waypoints remain distinguishable and preserved.

</details>

library/lib/utils/geometry/edgeGeometrySolver.ts
```diff
@@ -1337,6 +1347,33 @@ function computeAllEdgeGeometryPass(
...
+    // Straight-hook edges (use-case, syntax-tree, petri-net) stay diagonal lines
+    // through their authored interior waypoints. They auto-bend ONLY to clear an
+    // intervening node body (Track D); an unobstructed chord stays straight. The
+    // waypoints are honoured as mandatory checkpoints. The resulting polyline enters
+    // the neighbour map so step edges route around it.
+    if (straightHookTypes.has(edge.type ?? "")) {
+      const line = routeStraightPolyline({
+        source: endpoints.adjustedSource,
+        target: endpoints.adjustedTarget,
+        checkpoints: getStraightHookInterior(edge),
```
Claudia-Anthropica: @FelixTJDietrich [high] `candidateBounds` only spans the source and target nodes, but these checkpoints can be dragged arbitrarily far outside that box. `getEdgeObstacles` then omits nodes around the distant waypoint legs, so the straight router sees no blocker and routes directly through them. Expand the obstacle-query bounds to include every checkpoint and add an integration test with an out-of-bounds waypoint.

In `library/lib/utils/geometry/edgeGeometrySolver.ts`, straight-hook obstacle discovery uses endpoint-only candidate bounds even when manual checkpoints lie outside them. Expand the bounds passed to `getEdgeObstacles` to cover the source, target, and all straight-hook checkpoints, and add a regression test proving a distant checkpoint leg avoids a nearby node.

</details>

library/lib/edges/GenericEdge.tsx
```diff
@@ -493,6 +510,107 @@ export const EdgeBendHandle = ({
...
+    index: number
+  ) => void
+  onWaypointDoubleClick: (index: number) => void
+  onGhostPointerDown: (
+    event: ReactPointerEvent<SVGRectElement>,
+    segmentIndex: number
+  ) => void
+}) => {
+  const screenScale = useHandleScreenScale()
+  const ghosts = getSegmentGhostHandles(route)
```
Claudia-Anthropica: @FelixTJDietrich [medium] `route` is the solver route, not necessarily `[source, ...interior, target]`: automatic obstacle bends add segments that have no entry in `interior`. Feeding one of those segment indices to `insertWaypoint` indexes the authored array incorrectly, so dragging a ghost before an existing checkpoint can insert the new checkpoint after it and make the route double back. Preserve a solver-segment-to-checkpoint insertion mapping, or generate insertion ghosts from the authored base route.

In `library/lib/edges/GenericEdge.tsx`, ghost handles use segment indices from an auto-routed polyline even though insertion indexes the authored interior-waypoint array. Change the ghost data flow to provide the correct authored insertion index for every displayed segment, and cover a route with automatic bends before an existing checkpoint.

</details>

library/lib/hooks/useStraightPathEdge.ts
```diff
@@ -654,6 +699,199 @@ export const useStraightPathEdge = ({
...
+      if (!pointerTarget.hasPointerCapture(pointerId))
+        pointerTarget.setPointerCapture(pointerId)
+      const ownerDocument = pointerTarget.ownerDocument
+      dragInteriorRef.current = startInterior
+      dragMovedRef.current = false
+
+      // Drive the preview through state; the existing layout effect republishes it
+      // as an authoritative live override so neighbouring step edges reflow around
+      // the dragged diagonal, and clears it when the drag ends.
+      const publish = (interior: IPoint[]) => {
```
Claudia-Anthropica, line 778: @FelixTJDietrich [medium] This updates only the route preview; `EdgeWaypointHandles` still receives the committed `interiorPoints`. During an existing-point drag the solid handle stays at its old coordinate, and during a ghost drag no solid handle materializes until pointer-up, even though the bend itself follows the pointer. Keep preview interior points in render state and use them for the handle layer.

In `library/lib/hooks/useStraightPathEdge.ts`, waypoint dragging previews the route but continues rendering handles from committed `data.points`. Track the gesture's interior points in render state, return them to `EdgeWaypointHandles` during the drag, and test both an existing waypoint and a newly materialized ghost.

</details>

library/lib/hooks/useStraightPathEdge.ts
```diff
@@ -376,25 +401,35 @@ export const useStraightPathEdge = ({
...
       [renderPoints]
     )
 
+  // A bent edge (interior waypoints or a bridged crossing) draws its polyline via
+  // buildEdgePath; a plain 2-point edge keeps calculateStraightPath so the marker
+  // padding / include-extend gap on its single segment is byte-identical to before.
+  const isPolyline = renderPoints.length > 2 || lineJumps.length > 0
   const currentPath = useMemo(() => {
-    if (lineJumps.length > 0) return buildEdgePath(renderPoints, lineJumps)
+    if (isPolyline) return buildEdgePath(renderPoints, lineJumps)
```
Claudia-Anthropica, line 412: @FelixTJDietrich [medium] `buildEdgePath` draws a continuous polyline, so bending an include/extend edge removes the midpoint gap that `calculateStraightPath` creates for the stereotype text. The straight-edge label components also still position from the endpoint chord, which makes use-case and Petri labels float away from a bent route (often onto the node that caused the detour). Derive the label anchor/orientation from the local arc-mid segment and preserve the include/extend gap on that segment.

In `library/lib/hooks/useStraightPathEdge.ts`, bent straight-hook paths bypass the include/extend text gap and expose only endpoint-chord geometry to label renderers. Build a polyline path that carves the stereotype gap on the chosen local segment, pass that segment's endpoints to straight-edge labels, and add rendering-geometry tests for bent use-case and Petri edges.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The core routing approach and pure-helper coverage are substantial, but the new waypoint schema is not safe for already-saved 4.x models, and several obstacle/gesture/label integrations break once routes contain both authored and automatic bends. See the inline comments for the concrete cases that need fixing.

library/lib/utils/versionConverter.ts
Claudia-Anthropica: @FelixTJDietrich This is still open: the current import path still accepts every 4.x model unchanged, and the exported schema remains 4.1.0, so legacy full-route points are still interpreted as new interior waypoints.

library/lib/utils/geometry/edgeGeometrySolver.ts
Claudia-Anthropica: @FelixTJDietrich This is still open: candidateBounds is still computed solely from the endpoint nodes and passed unchanged to getEdgeObstacles, so nodes near a distant manual checkpoint remain outside the query.

library/lib/edges/GenericEdge.tsx
Claudia-Anthropica: @FelixTJDietrich This is still open: ghosts still come from every segment of the solver route and pass that segment index directly to insertWaypoint, so automatic bends can still shift the authored insertion position.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @FelixTJDietrich This is still open: the publish callback updates only dragPreviewPoints, while the hook still returns the committed interiorPoints; the solid handle layer therefore still renders committed positions during the drag.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @FelixTJDietrich This is still open: bent routes still use buildEdgePath without the include/extend gap, and label geometry still exposes only the route endpoints plus the generic midpoint instead of the local label segment.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The facing-side and balanced-port work is a useful addition, but it does not address the five existing routing, migration, gesture, and label findings. Those threads remain open, so I’m requesting changes again on this SHA.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest routing adjustments do not address the five existing migration, obstacle-query, waypoint-gesture, and label-geometry requests. Those threads remain open, so I’m requesting changes again on this SHA.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest routing-cost changes do not address the five existing migration, obstacle-query, waypoint-gesture, and label-geometry requests. Those threads remain open, so I’m requesting changes again on this SHA.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest port-coordination commit does not address the five existing migration, obstacle-query, waypoint-gesture, and label-geometry requests. Those threads remain open, and CI now has four related straight-edge E2E regressions, so I’m requesting changes again on this SHA.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest detour-nesting and stability changes do not address the five existing migration, obstacle-query, waypoint-gesture, and label-geometry requests. Those threads remain open, so I’m requesting changes again on this SHA.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The symmetric port-seat fix is focused, but this commit does not address the five open migration, obstacle-query, waypoint-gesture, and label-geometry requests. Those threads remain unresolved, so I’m requesting changes again on this SHA.

library/lib/edges/GenericEdge.tsx
Claudia-Anthropica: @FelixTJDietrich Nice, using the rendered interior vertices makes the ghost segment index match the array that gets persisted.

library/lib/hooks/useStraightPathEdge.ts
Claudia-Anthropica: @FelixTJDietrich Looks good now — the handle layer follows the rendered drag preview, including newly materialized points.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The rendered-route waypoint and live handle-preview fixes address two of the five requests. The 4.x migration, distant-checkpoint obstacle query, and bent include/extend and label geometry threads are still open, so I’m requesting changes on this SHA.

PR #833, opened 2026-07-27
fix(release): preserve signed OTA recovery artifacts

FelixTJDietrich:
## Summary

- upload the signed OTA zip and manifest as a seven-day artifact before entering the deployment transport
- allow direct-SSH recovery when the deployment gateway is unavailable
- keep the private signing key inside the protected Production workflow

## Context

Production run [30259204116](https://github.com/ls1intum/Apollon/actions/runs/30259204116) failed before copying any files because `gateway.aet.cit.tum.de:2010` timed out. The app-only 5.2.0 deployment has been recovered directly and is healthy; this artifact is needed to publish its matching signed OTA bundle safely.

## Validation

- `actionlint .github/workflows/ios-live-update.yml`
- Prettier check
- documentation markdown lint

No changeset: release recovery and documentation only.

### DIFF
--- a/.github/workflows/ios-live-update.yml
+++ b/.github/workflows/ios-live-update.yml
@@ -88,6 +88,17 @@ jobs:
           CAPGO_MIN_NATIVE_VERSION: ${{ vars.CAPGO_MIN_NATIVE_VERSION }}
         run: pnpm run live-update:build
 
+      # Preserve the public, signed artifacts before entering the deployment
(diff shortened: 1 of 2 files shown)


Claudia-Anthropica (approved):
@FelixTJDietrich This cleanly preserves the signed recovery bundle before the gateway-dependent deployment steps, and the documentation accurately describes the recovery path. Approving as-is.

PR #835, opened 2026-07-28
feat(library): route straight edges and tidy syntax trees

FelixTJDietrich:
### Summary

This PR adds automatic routing for straight-edge diagram families and a tidy
layout for syntax trees:

- Clear connections stay straight. A connection reroutes only when it actually
  meets a node; connector crossings keep the existing line-jump presentation.
- Necessary detours keep breathing room around nodes, favor short and gentle
  paths, and never replace a readable crossing with a disproportionate trip
  around the diagram.
- Connections choose facing sides and spread sibling endpoints across shared
  borders instead of stacking on one point.
- Hand-placed routes remain exactly as drawn. Automatic bends are derived and
  never written into the diagram model.
- Syntax trees gain a **Tidy tree layout** toolbar action and the additive
  `ApollonEditor.layoutSyntaxTree()` API.

The branch is rebased directly onto the current `main`, which includes the
straight-edge editing work from #834.

Follow-up to the node-overlap report in #282, which was closed by the manual
straight-edge editing work in #834. Long-text wrapping is outside this PR and is
not claimed as fixed.

### Why this is intentionally conservative

Automatic routing should help without making a diagram feel restless:

- The direct line has a 1 px collision guard, while a generated detour receives
  the full 10 px minimum clearance. A nearby node therefore does not trigger a
  route change just because it enters the preferred clearance halo.
- Other connections do not become routing obstacles. Crossings therefore remain
  direct and predictable, while the existing line jumps preserve readability.
- A 5% deterministic endpoint-side deadband prevents small movements near a
  node corner from flipping the connection to another side.
- While a node is moving, its valid straight-edge topology is kept for the full
  gesture. It changes immediately only if keeping it would cross a node, then
  settles to the exact route on release.
- Node drags involving straight connections use the routing worker even in small
  diagrams. The display-only handoff preserves diagonal polylines instead of
  briefly turning them into step edges.

### Compatibility and failure behavior

- Existing hand-authored straight routes bypass automatic obstacle routing.
- Existing orthogonal routing is unchanged.
- Native-handle previews and commits use the same validated handle point, even
  when the pointer is off-center or nodes overlap.
- Tidy layout is one undoable operation. Valid trees and forests are arranged;
  cyclic, multi-parent, and otherwise malformed components stay where the user
  placed them.
- The toolbar waits for controlled node positions to reach React Flow and then
  performs the same inset-aware fit as the public API.
- `layoutSyntaxTree()` is additive and is a no-op for other diagram types.

### Release note

Keep straight connections clear, stable, and responsive automatically. They now
choose facing sides, spread sibling links across shared borders, and route around
nodes without reorganizing when other connections cross them; clear shots remain
straight and hand-placed routes stay exactly as drawn. Syntax trees also gain a
one-click tidy layout that arranges valid parent-child forests while leaving
malformed portions untouched.

### Implementation notes

- Routing uses a deterministic, integer-based visibility graph and A* search,
  including deterministic tie-breaking.
- The visibility graph starts with obstacles that block the direct route, adds
  newly relevant node bodies locally, and has a bounded full-field fallback.
- Third-party node bodies are never exempted when an endpoint starts inside
  their clearance halo.
- Coordinated endpoint seats keep straight-edge fans evenly distributed.
- In a new crossing-heavy 12-edge Chromium regression fixture, local profiling
  reduced the worst solve from 2,629.7 ms to 20.5 ms. This is a regression
  fixture result, not a general-purpose benchmark.
- Bundle cost remains explicit and within the existing budgets:
  - main entry: 125.72 kB / 126 kB
  - routing worker: 41.45 kB / 44 kB
  - model chunk: 5.08 kB / 8 kB
  - export chunk: 413.41 kB / 700 kB

### Steps for testing

1. Open a crowded use-case, syntax-tree, or Petri-net diagram and confirm that
   straight connections avoid intervening nodes.
2. Move a node close to a clear connection without touching it; the edge should
   remain straight. Move it into the line; the edge should take a compact,
   well-cleared detour.
3. Move a node repeatedly near an endpoint-side boundary and confirm the edge
   does not jump sides on small movements.
4. Create sibling connections and confirm their endpoints fan out. Cross two
   connections and confirm they remain direct and use the existing line jump.
5. Add or move a manual bend, then place a node over that route. The authored
   route should remain exactly as drawn.
6. Drop a connection off-center within a native handle and confirm the committed
   endpoint exactly matches the preview.
7. Apply **Tidy tree layout** to a valid forest and confirm one-step undo and an
   inset-aware fit. Repeat with cyclic or multi-parent content and confirm the
   malformed components are not rearranged.
8. Call `ApollonEditor.layoutSyntaxTree()` on a syntax tree and on another
   diagram type; confirm layout in the former and a no-op in the latter.

### Verification

- `pnpm lint && pnpm format:check && pnpm build && pnpm test`
  - 95 test files passed
  - 1,860 tests passed; 1 skipped
- Fresh production-build Chromium interaction run:
  - 17/17 targeted connection and straight-edge UX tests passed
- Fresh production-build crossing-heavy performance regression:
  - passed 5/5 Chromium runs and 5/5 Firefox runs at the 34 ms p95 frame budget
- `pnpm --filter @tumaet/apollon size`
  - all four published bundle budgets passed

### Screenshots / screencasts

A syntax tree after applying the tidy tree layout, with a balanced hierarchy and clear straight connections

Captured from this branch's production build after choosing **Tidy tree layout**.
Deterministic geometry fixtures cover route quality and failure cases beyond this
representative diagram.

### Checklist

- [x] Linked the applicable portion of #282
- [x] Added a user-facing changeset
- [x] Used a Conventional Commit PR title matching the user-visible feature
- [x] Added unit and browser interaction coverage
- [x] Passed the complete local pre-PR gate
- [x] Kept published bundles within their documented budgets
- [x] Documented the additive public API inline
- [x] Included a representative screenshot

### DIFF
--- a/library/lib/chrome/builtins/ZoomControls.tsx
+++ b/library/lib/chrome/builtins/ZoomControls.tsx
@@ -28,6 +28,21 @@ export interface ZoomControlsProps {
 }
 
 /**
+ * Apply controlled node positions, then wait two paint boundaries before fitting.
+ * The first lets React publish the new node array; the second lets React Flow
+ * measure and index it. Kept injectable for a deterministic toolbar regression
+ * test.
+ */
+export const runTidyLayoutAndFit = (
+  layout: () => void,
+  fit: () => void,
+  schedule: (callback: FrameRequestCallback) => number = requestAnimationFrame
+): void => {
+  layout()
+  schedule(() => schedule(() => fit()))
+}
+
+/**
  * The canvas cluster: a [zoom-out][%-reset][zoom-in][fit][multi-select] island and
  * a separate [undo][redo] history island. The fit button reserves the current insets
  * so content frames clear of the chrome. `history: false` drops the history island.
@@ -137,7 +152,11 @@ export function ZoomControls({ history = true }: ZoomControlsProps) {
             <button
               type="button"
               className="apollon-chrome-iconbtn"
-              onClick={() => layoutSyntaxTree()}
+              onClick={() =>
+                runTidyLayoutAndFit(layoutSyntaxTree, () =>
+                  insetAwareFitView(rf, insets, safeArea, { duration: 200 })
+                )
+              }
               aria-label={t.tidyLayout}
             >
               <ListTree width={18} height={18} aria-hidden="true" />
--- a/library/lib/chrome/builtins/ZoomControls.tsx
+++ b/library/lib/chrome/builtins/ZoomControls.tsx
@@ -1,6 +1,7 @@
 import { useReactFlow, useStore } from "@xyflow/react"
 import { useShallow } from "zustand/shallow"
 import {
+  ListTree,
   Maximize,
   Redo2,
   SquareMousePointer,
@@ -13,6 +14,8 @@ import {
   useMetadataStore,
   useOverlayStore,
 } from "@/store/context"
+import { useDiagramModifiable } from "@/hooks/useDiagramModifiable"
+import { UMLDiagramType } from "@/types"
 import { insetAwareFitView } from "@/overlay/fitView"
 import { ariaKeyshortcuts } from "@/keyboard"
 import { Tooltip } from "@/components/ui"
@@ -48,12 +51,17 @@ export function ZoomControls({ history = true }: ZoomControlsProps) {
     }))
   )
 
-  const { multiSelectionMode, setMultiSelectionMode } = useMetadataStore(
-    useShallow((state) => ({
-      multiSelectionMode: state.multiSelectionMode,
-      setMultiSelectionMode: state.setMultiSelectionMode,
-    }))
-  )
+  const { multiSelectionMode, setMultiSelectionMode, isSyntaxTree } =
+    useMetadataStore(
+      useShallow((state) => ({
+        multiSelectionMode: state.multiSelectionMode,
+        setMultiSelectionMode: state.setMultiSelectionMode,
+        isSyntaxTree: state.diagramType === UMLDiagramType.SyntaxTree,
+      }))
+    )
+  const layoutSyntaxTree = useDiagramStore((state) => state.layoutSyntaxTree)
+  const isModifiable = useDiagramModifiable()
+  const showTidyLayout = isSyntaxTree && isModifiable
 
   const { ref: toolbarRef, onKeyDown: onToolbarKeyDown } =
     useRovingToolbar<HTMLDivElement>()
@@ -124,6 +132,18 @@ export function ZoomControls({ history = true }: ZoomControlsProps) {
             <SquareMousePointer width={18} height={18} aria-hidden="true" />
           </button>
         </Tooltip>
+        {showTidyLayout && (
+          <Tooltip title={t.tidyLayoutHint}>
+            <button
+              type="button"
+              className="apollon-chrome-iconbtn"
+              onClick={() => layoutSyntaxTree()}
+              aria-label={t.tidyLayout}
+            >
+              <ListTree width={18} height={18} aria-hidden="true" />
+            </button>
+          </Tooltip>
+        )}
       </div>
 
       {history && undoManagerExist && (
--- a/library/lib/hooks/useConnect.ts
+++ b/library/lib/hooks/useConnect.ts
@@ -226,11 +226,16 @@ export const useConnect = () => {
           // central solver can immediately move the endpoint to an automatic seat.
           const edgeId = pendingConnectionId.current
           if (edgeId && connectionState.toNode) {
-            // `FinalConnectionState.to` is the exact validated handle centre in
-            // screen coordinates; convert that point, rather than the release
-            // event's potentially offset pointer, into the node's flow-space rect.
+            // Continuous outlines intentionally follow the aimed drop point that
+            // their snap circle previews. Other native targets use the validated
+            // handle centre, so an off-centre release inside a handle cannot shift
+            // the committed endpoint. In both cases `toNode` remains authoritative:
+            // overlapping scene nodes never replace React Flow's validated target.
+            const targetPoint = dropAnchorIsAimed(connectionState.toNode.type)
+              ? getDropPosition(event)
+              : screenToFlowPosition(connectionState.to)
             const anchor = getNativeConnectionAnchor({
-              to: screenToFlowPosition(connectionState.to),
+              to: targetPoint,
               toNode: connectionState.toNode,
             })
             if (anchor) {
--- a/library/lib/hooks/useConnect.ts
+++ b/library/lib/hooks/useConnect.ts
@@ -17,6 +17,7 @@ import {
 import {
   dropAnchorIsAimed,
   getEdgeAnchorFromPoint,
+  getNativeConnectionAnchor,
 } from "@/utils/connectionModes"
 import { HandleId } from "@/nodes/wrappers"
 import { useDiagramStore, useMetadataStore } from "@/store/context"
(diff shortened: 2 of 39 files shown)


library/lib/utils/geometry/straightPolylineRouter.ts
```diff
@@ -0,0 +1,1109 @@
...
+export function chordClearsObstacles(
+  a: IPoint,
+  b: IPoint,
+  hardObstacles: StraightRouteObstacle[],
+  minClearancePx: number
+): boolean {
+  for (const o of hardObstacles) {
+    if (o.soft) continue
+    const inflated = inflate(rectOf(o), minClearancePx)
+    if (inclusiveContains(inflated, a) || inclusiveContains(inflated, b))
```
Claudia-Anthropica: @FelixTJDietrich [high] This exempts the entire obstacle whenever either endpoint lies inside its inflated clearance rect. The caller already removes the actual source/target nodes, so a different node only 10px from an endpoint is treated as an “own node”; a chord can then run straight through its body. Keep the near-end clearance relaxation if needed, but never exempt the third-party node’s actual rectangle.

In `library/lib/utils/geometry/straightPolylineRouter.ts`, an endpoint inside a third-party obstacle's inflated clearance rect causes the whole obstacle to be skipped. Change the collision and visibility checks so a near endpoint may leave an unsatisfied clearance halo but no segment may enter the obstacle's actual body, and add a regression test with a source beside a blocker whose chord crosses that blocker.

</details>

library/lib/utils/geometry/straightPolylineRouter.ts
```diff
@@ -0,0 +1,1109 @@
...
+  const allSiblingRoutes = siblingRoutes ?? []
+  // Only unrelated neighbors contribute detour vertices. Siblings arrive through
+  // the greedy route walk one at a time; materialising vertices from only the
+  // already-routed half makes a symmetric fan resolve asymmetrically. Their
+  // crossings/overlaps are still priced by the search, while coordinated ports
+  // and obstacle rings provide the order-independent sibling separation.
+  const conflictingRoutes = neighborRoutes.filter((route) =>
+    conflictsStructurally([route])
+  )
+  if (straightClears && conflictingRoutes.length === 0) return direct
```
Claudia-Anthropica: @FelixTJDietrich [medium] The direct-route gate only checks `neighborRoutes`, so a chord that crosses or overlaps a route in `siblingRoutes` returns here without ever pricing that conflict. That contradicts the sibling contract above (“never” exempt crossings/overlaps) and leaves crossings between connectors that share a node. Include structural sibling conflicts in the gate while still excluding sibling crowding.

In `library/lib/utils/geometry/straightPolylineRouter.ts`, the direct-route gate ignores structural conflicts from `siblingRoutes`. Make sibling crossings and overlaps open the visibility search without applying sibling crowding, and add a regression test where an otherwise-clear chord crosses only a sibling route.

</details>

library/lib/hooks/useConnect.ts
```diff
@@ -219,7 +219,40 @@ export const useConnect = () => {
   const onConnectEnd: OnConnectEnd = useCallback(
     (event, connectionState) => {
       try {
-        if (!connectionState.isValid) {
+        if (connectionState.isValid) {
+          // A valid native handle is an exact attachment point, not merely a side
+          // hint. Persist the preview's resolved anchor on commit; otherwise the
+          // central solver can immediately move the endpoint to an automatic seat.
+          const edgeId = pendingConnectionId.current
+          const dropPosition = getDropPosition(event)
```
Claudia-Anthropica: @FelixTJDietrich [medium] A valid native drop already provides the exact `connectionState.to` point and `toNode`, but this recomputes the anchor from the release pointer and an independent freeform hit-test. Releasing anywhere inside a handle can therefore commit a different ratio than the preview, and overlapping nodes can even supply the wrong rectangle for the real target. Derive the anchor from the native connection state instead.

In `library/lib/hooks/useConnect.ts`, valid native-handle commits derive their anchor from the pointer event and `resolveDropTarget` instead of the exact native target. Use `connectionState.to` plus `connectionState.toNode` geometry (or the equivalent exact handle point) to persist the anchor, and test pointer-offset and overlapping-node drops for preview/commit equality.

</details>

library/lib/chrome/builtins/ZoomControls.tsx
```diff
@@ -124,6 +132,18 @@ export function ZoomControls({ history = true }: ZoomControlsProps) {
             <SquareMousePointer width={18} height={18} aria-hidden="true" />
           </button>
         </Tooltip>
+        {showTidyLayout && (
+          <Tooltip title={t.tidyLayoutHint}>
+            <button
+              type="button"
+              className="apollon-chrome-iconbtn"
+              onClick={() => layoutSyntaxTree()}
```
Claudia-Anthropica: @FelixTJDietrich [low] The public `layoutSyntaxTree()` method fits the result, and the PR's test flow promises the same for the toolbar action, but this button only moves nodes. Large trees can be laid out partly off-screen until the user separately clicks Fit. Schedule the same inset-aware fit after React Flow receives the new positions.

In `library/lib/chrome/builtins/ZoomControls.tsx`, the tidy-layout toolbar button does not fit the newly positioned tree into view. After `layoutSyntaxTree()`, schedule `insetAwareFitView(rf, insets, safeArea)` after the controlled node positions reach React Flow, and cover the toolbar action in a test.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The deterministic routing and tidy-tree work is thoughtfully structured, but the new straight router can still pass through a nearby node and can miss sibling-only crossings. Native-handle commits can also diverge from the preview; I left the concrete fixes inline.

library/lib/utils/geometry/straightPolylineRouter.ts
FelixTJDietrich: Fixed in 20db1b61. The near-end relaxation now applies only to the inflated clearance halo; the direct gate, visibility graph, and simplifier still reject every segment that enters the actual obstacle body. I added the requested source-beside-blocker regression whose chord crosses that body.

library/lib/utils/geometry/straightPolylineRouter.ts
FelixTJDietrich: Fixed in 20db1b61. Structural sibling crossings and overlaps now open the visibility search and contribute detour vertices, while ordinary sibling crowding remains exempt so fans do not become order-dependent. The new regression covers an otherwise-clear chord that conflicts only with a sibling.

library/lib/hooks/useConnect.ts
FelixTJDietrich: Fixed in 20db1b61. Preview and commit now share a helper based on the validated native toNode and exact handle center. The commit path explicitly converts FinalConnectionState.to from screen space to flow space before deriving the stored anchor. Unit coverage checks pointer/scene disagreement, and the production-browser regression releases off-center inside the handle and verifies preview/commit equality.

library/lib/chrome/builtins/ZoomControls.tsx
FelixTJDietrich: Fixed in 20db1b61. The toolbar action now applies the controlled layout, waits two animation frames for React Flow to receive and measure the new positions, and then performs the inset-aware fit with a short transition. A deterministic unit test verifies that ordering; the public API behavior remains unchanged.

Claudia-Anthropica (approved):
@FelixTJDietrich The follow-up fixes address the obstacle-halo, sibling-crossing, and native-handle preview/commit issues from my previous review, with focused regressions covering each path. The relevant local unit suites pass, so this looks good to merge.

Claudia-Anthropica (approved):
@FelixTJDietrich The new commits preserve obstacle-body safety while stabilizing straight-edge routing, aimed-outline previews, and worker-backed interaction performance. The previous findings remain resolved, all captured CI checks pass, and I found no current blocking defect.

PR #840, opened 2026-08-05
fix(library): restore relationship assessment popovers

FelixTJDietrich:
### Summary

Relationship assessment popovers open again when edge-editing actions are hidden.

Assessment integrations can intentionally make a diagram read-only while still letting instructors double-click nodes and relationships to add feedback. Node feedback continued to work, but relationship feedback silently failed because its positioning anchor disappeared together with the edit toolbar.

### Release note

Assess relationships reliably when editing controls are hidden.

### Implementation notes

React Flow unmounts an `EdgeToolbar` when its `isVisible` prop is false. Apollon used that same toolbar as the relationship popover anchor, so hiding delete, edit, and routing controls also removed the anchor required by assessment popovers.

This change keeps React Flow's positioning host mounted as a non-interactive anchor and conditionally renders only the editing actions. The anchor lifecycle is therefore independent of toolbar visibility, without adding host-specific DOM handling or changing Apollon's public API.

A focused regression test verifies both sides of the contract: the anchor remains mounted while editing buttons remain absent. A patch changeset is included.

### Steps for testing

1. Open a modeling submission in a host integration that provides assessment popovers while the diagram is read-only.
2. Double-click a relationship.
3. Verify that the assessment popover opens at the relationship and exposes its feedback fields.
4. Verify that edit, delete, and reset-routing buttons remain hidden.
5. Double-click a node and confirm that its existing assessment behavior is unchanged.

Automated validation:

- `pnpm lint` (green; existing warnings only)
- `pnpm format:check`
- `pnpm build`
- `pnpm test` (89 files, 1,782 passed, 1 skipped)
- Downstream Artemis Playwright regression using real pointer input for relationship assessment

### Screenshots / screencasts

Not applicable: this restores an existing transient popover interaction without changing its visual design. The interaction and hidden-toolbar state are covered by automated tests.

### Checklist

- [x] Linked to a related issue (not applicable — no related issue)
- [x] Added a changeset whose summary is written in the user's voice
- [x] PR title's Conventional Commit type matches the kind of change
- [x] Tests added or updated
- [x] Ran `pnpm lint && pnpm format:check && pnpm build && pnpm test` locally — green
- [x] Documentation updated (not applicable — no public API or workflow changed)
- [x] Screenshots or screencasts attached (not applicable — no visual design changed)

PR #839, opened 2026-07-31
fix(library): keep floating editor surfaces visible in fullscreen

FelixTJDietrich:
### Summary

Route editor-owned floating surfaces through an editor-local portal root when element-level fullscreen excludes the document body. This keeps popovers, selects, tooltips, color controls, and palette drag previews visible and interactive in subtree fullscreen without a global shared root or host-specific z-index workarounds.

### Release note

Popovers, selects, tooltips, color pickers, and drag previews remain visible and interactive when an embedded editor enters fullscreen.

### Implementation notes

- Each editor declaratively owns one lifecycle-bound, layout-neutral portal root.
- Surfaces continue to portal to `document.body` whenever body remains in the fullscreen subtree, preserving normal embedding and host dialog layering.
- Subtree fullscreen uses the editor-local root only when body is excluded by the browser's fullscreen top layer.
- The fullscreen subscription re-portals already-open surfaces when the Fullscreen API state changes.
- Multiple editors remain isolated: only the editor contained by `document.fullscreenElement` uses its local root.
- The shared UI tooltip now accepts Base UI's native portal container option instead of adding a parallel portal implementation.

### Steps for testing

1. Run `pnpm lint && pnpm format:check && pnpm build && pnpm test`.
2. Run `pnpm exec playwright test standalone/webapp/tests/e2e/fullscreen-portals.spec.ts`.
3. In the local editor, fullscreen the editor subtree, edit a node, hover a chrome control, and drag a palette item; confirm the popover, tooltip, and drag preview stay visible.
4. Exit fullscreen and confirm floating surfaces return to the document portal.

### Screenshots / screencasts

Not attached: this fixes top-layer behavior rather than static appearance. The added Playwright test exercises the behavior in a real browser.

### Checklist

- [ ] Linked to a related issue (not applicable)
- [x] Added a changeset whose summary is written in the user's voice
- [x] PR title's Conventional Commit type matches the kind of change
- [x] Tests added or updated
- [x] Ran `pnpm lint && pnpm format:check && pnpm build && pnpm test` locally — green
- [ ] Documentation updated (not applicable)
- [ ] Screenshots or screencasts attached (behavior covered by Playwright)

PR #841, opened 2026-08-06
fix(library): make embedded editors reliable in host applications

FelixTJDietrich:
### Summary

Make Apollon reliable as an editor embedded in a host-controlled page. The
changes were driven by the Apollon 5.2 integration in
[Artemis#13429](https://github.com/ls1intum/Artemis/pull/13429), where assumptions
that Apollon owned the whole document surfaced as visible integration defects.

This PR addresses those defects as one host-integration package:

- keep popovers, selects, tooltips, color pickers, and drag previews available
  when the editor workspace enters fullscreen;
- scope keyboard shortcuts and wheel handling to the active editor so the host
  page remains usable, including pages with multiple editors;
- make assessment relationships, highlighting, navigation, and feedback
  popovers reliable and expose `revealAssessment(id)` for host feedback lists;
- keep saved-diagram connection anchors, resize handles, palette drops, labels,
  grid alignment, zoom rendering, and route previews visually and geometrically
  stable;
- provide a neutral labeled chrome action for host controls and prevent the
  offscreen export renderer from changing the host page's scroll size; and
- add a real fullscreen action to the playground for manual host-style testing.

This supersedes #839 and #840, which split the same Artemis integration work
across two branches; both are closed in favor of this PR.

### Release note

- Embed Apollon reliably in host pages: fullscreen overlays stay visible,
  scrolling and shortcuts remain host-safe, assessment feedback is easier to
  open and navigate through `revealAssessment(id)`, host controls gain a labeled
  chrome action, exports no longer expand the page, and diagram interactions
  remain aligned, sharp, and responsive. Labels now also stay vertically
  centered in SVG, PNG, and PDF exports exactly as they appear in the editor.
- Test embedded-editor overlays and drag interactions directly in the playground
  with its new fullscreen workspace action.

These are the two changeset summaries verbatim, one for the `@tumaet/apollon`
embedder audience and one for the `@tumaet/webapp` playground audience.

### Implementation notes

**Fullscreen portals.** Each editor instance owns a portal destination. It moves
with the fullscreen element on `fullscreenchange`, so floating surfaces remain
descendants of the browser's active fullscreen subtree without requiring the
host to re-parent library internals.

**Host input ownership.** Shortcuts dispatch from the focused editor root instead
of `document`, and React Flow's global modifier listeners are enabled only while
that root owns the interaction. Pointer-acquired focus releases on pointer leave;
keyboard-acquired focus persists until the user tabs away. With scroll lock on,
ordinary wheel input returns to the host page while the zoom modifier still
zooms the diagram.

**Assessment.** Assessment uses a route-following relationship hit surface,
single-click feedback, outline-based node highlighting, hidden connection
handles, and a fixed navigation footer with text-input-safe arrow shortcuts.
`revealAssessment(id)` lets a host feedback list select an element, open its
feedback, and pan to it without changing zoom.

**Geometry and rendering.** Connection handles remain measurable when a restored
diagram first mounts, only relevant anchors are mounted, resize handles appear
for the selected node with usable hit targets, and palette preview/drop placement
shares one grab point. Grid rendering, label centering, raster invalidation, and
shared route-preview selection remove the integration's visual drift and
frame-time regressions. Export baseline normalization resolves each positioned
text run exactly once, so the same centered labels are preserved in SVG, PNG,
and PDF output.

**Performance guard.** The Firefox interaction benchmark uses visible nodes,
rejects unusably slow runners, and measures without continuous trace capture.
Current `main` reaches 65–66 ms locally on that workload; this branch measures
34–49 ms locally and 83–84 ms on GitHub's software-rendered Firefox runner. The
hosted ceiling is 85 ms, with a separate 55 ms idle-quality bound.

### Steps for testing

Standalone:

```sh
pnpm install
pnpm dev
```

Open http://localhost:5173/playground, then:

1. Click **Enter editor fullscreen**. Open a node popover, a select, the style
   editor's color picker, and drag an element from the palette. Each surface must
   remain visible and interactive. Press Escape until fullscreen exits.
2. With the pointer outside the canvas, press <kbd>Mod</kbd>+<kbd>-</kbd>; the
   browser page must zoom. Click the canvas and repeat; the diagram must zoom.
3. Turn on scroll lock and wheel over the canvas; the surrounding page must
   continue to scroll. Hold the displayed zoom modifier; the diagram must zoom.
4. In assessment mode, single-click a relationship, including a multi-segment
   automatic route. Its feedback must open immediately. Walk the diagram with
   Previous/Next Assessment and <kbd>←</kbd>/<kbd>→</kbd>; the footer must remain
   visible and typing in feedback must not trigger navigation.
5. Reload a saved diagram and drag a new relationship without first hovering its
   source. It must begin at the anchor under the cursor. Drag a Petri-net place or
   SFC transition branch from low in its palette preview; the dropped element must
   land where the ghost was shown.
6. Export class/object, activity, use-case, and multiline-label diagrams as
   SVG, PNG, and PDF. Label centers must match the editor, including headers,
   stereotypes, wrapped text, diamonds, and ellipses.
7. Export a large diagram and confirm the host page does not gain scrollbars.

The focused fullscreen regression can also be run directly:

```sh
pnpm --filter @tumaet/webapp exec playwright test tests/e2e/fullscreen-portals.spec.ts --project=chromium
```

Host integration is covered end to end by
[Artemis#13429](https://github.com/ls1intum/Artemis/pull/13429), which consumes
this branch locally and will pin the released package version before merging.

### Screenshots / screencasts

The refreshed Linux visual baselines committed with this PR show the rendered
changes directly:

- [Class diagram](https://github.com/ls1intum/Apollon/blob/68d755b13de1f5215f3569d3d90c81a33c4af08d/standalone/webapp/tests/visual/diagrams.visual.spec.ts-snapshots/visual-class-diagram.png)
- [Object diagram](https://github.com/ls1intum/Apollon/blob/68d755b13de1f5215f3569d3d90c81a33c4af08d/standalone/webapp/tests/visual/diagrams.visual.spec.ts-snapshots/visual-object-diagram.png)
- [BPMN diagram](https://github.com/ls1intum/Apollon/blob/68d755b13de1f5215f3569d3d90c81a33c4af08d/standalone/webapp/tests/visual/diagrams.visual.spec.ts-snapshots/visual-bpmn.png)
- [SVG export](https://github.com/ls1intum/Apollon/blob/68d755b13de1f5215f3569d3d90c81a33c4af08d/standalone/webapp/tests/visual/svg-export.visual.spec.ts-snapshots/svg-export-object-diagram.png)

The host-side layouts and assessment states are documented in
[Artemis#13429](https://github.com/ls1intum/Artemis/pull/13429).

### Checklist

- [x] Linked to a related issue (if applicable)
- [x] Added changesets whose summaries are written in the user's voice (`pnpm changeset`, [how](https://ls1intum.github.io/Apollon/contributor/development/release-notes/))
- [x] PR title's Conventional Commit type (`fix`) matches the user-visible kind of change and its release-note group
- [x] Tests added or updated
- [x] Ran `pnpm lint && pnpm format:check && pnpm build && pnpm test` locally — green (102 test files, 1853 passed, 1 skipped); `pnpm --filter @tumaet/apollon build` is green and the published main-entry size gate passes at 121.89 kB/122 kB; the complete Firefox performance suite and focused assessment/highlight/fullscreen E2E tests also pass
- [x] Documentation updated


library/lib/styles/app.css
```diff
@@ -689,15 +706,15 @@
...
-  opacity: 0;
   background-color: transparent;
 }
 
 .react-flow__handle.apollon-arc-handle::before {
   content: "";
   position: absolute;
-  background-color: var(--apollon-primary, #3e8acc);
-  opacity: 0.4;
+  /* Transparent until the node is hovered or selected; the reveal rules below
```
Claudia-Anthropica, line 716: @FelixTJDietrich [high] The exact-head Codacy check is ACTION_REQUIRED, and its captured evidence attributes 19 new issues to this changed stylesheet, including this missing blank line and selector-order violations at lines 891, 932–968, 1223, 1242, and 1266. Please clean up all reported spacing and descending-specificity violations so this diff no longer causes the failing check.

In `library/lib/styles/app.css`, the change introduces Codacy/stylelint spacing and descending-specificity violations. Add the required blank lines and reorder the affected selectors so all 19 introduced findings are removed.

</details>

library/lib/components/wrapper/AssessmentSelectableWrapper.tsx
```diff
@@ -150,43 +146,46 @@ export const AssessmentSelectableWrapper: React.FC<
...
-    ...(isSelected && {
-      backgroundColor: INTERACTIVE_SELECTION_FILL,
-      border: `2px solid ${INTERACTIVE_SELECTION_COLOR}`,
-    }),
-    ...(isHighlighted &&
-      !isSelected && {
-        backgroundColor: INTERACTIVE_SELECTION_FILL_STRONG,
-        border: `2px solid ${INTERACTIVE_SELECTION_COLOR}`,
-      }),
+    ...(highlightColor && { position: "relative", isolation: "isolate" }),
```
Claudia-Anthropica, line 175: @FelixTJDietrich [medium] `highlightClass` is applied only by the `asElement === "g"` branch. Normal nodes use the default `div` branch, whose hard-coded `className="nodrag nopan"` drops this class while `combinedStyle` no longer contains the previous selected/hovered styling, so read-only assessment selection provides no visual feedback for nodes.

In `library/lib/components/wrapper/AssessmentSelectableWrapper.tsx`, the default div branch drops `highlightClass`, leaving readonly assessment node hover and selection visually unstyled. Add `highlightClass` to the div branch's class name.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich I found two blocking issues on this exact head: read-only assessment selection no longer paints node highlights, and the captured Codacy run is ACTION_REQUIRED for 19 issues introduced in the changed stylesheet. The inline comments identify the affected code and concrete fixes.

library/lib/components/wrapper/AssessmentSelectableWrapper.tsx
FelixTJDietrich: Fixed in 5684407b. The div branch now carries the same assessment highlight classes as the SVG group branch, and a focused AssessmentSelectableWrapper regression test asserts selected read-only nodes receive both highlight classes. The new unit test passes; the focused assessment/highlight/fullscreen Playwright set passes 9/9; the full library suite passes 1,842 tests.

library/lib/styles/app.css
FelixTJDietrich: Fixed in 5684407b. I retrieved all 19 exact-head Codacy annotations from the check-run API and corrected every added spacing, hex-length, and descending-specificity finding by ordering the selectors rather than suppressing rules. Running the same Stylelint rules before/after drops app.css from 36 to 17 pre-existing findings and tokens.css from 10 to 7: exactly 19 removed and no new finding. The full local PR gate is green.

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,96 @@ export class ApollonEditor {
...
+    const owner = nodes.find((node) =>
+      assessedIdsFor(node.id, nodes).includes(elementId)
+    )
+    const targetId = owner?.id ?? elementId
+    const element =
+      nodes.find((node) => node.id === targetId) ??
+      edges.find((edge) => edge.id === targetId)
+    const isTopLevel = element !== undefined
+
+    setNodes((current) =>
```
Claudia-Anthropica: @FelixTJDietrich [medium] These updates change the React Flow `selected` flags without synchronizing `diagramStore.selectedElementIds`. In non-readonly assessment mode, `getSelectedElements()` and `subscribeToSelectionChange()` therefore report stale IDs; the `null` branch likewise leaves that index uncleared despite promising to clear selection.

In `library/lib/apollon-editor.tsx`, `revealAssessment` changes selected node/edge flags without synchronizing `diagramStore.selectedElementIds`, leaving selection APIs stale in tutor mode and after clearing. Call `setSelectedElementsId` with `[targetId]` in the selection branch and `[]` in the null branch.

</details>

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,96 @@ export class ApollonEditor {
...
+    // Selection follows what the caller asked for, not what had to be opened to
+    // show it. Asking for a class selects the class and its members, so a host
+    // list marks the whole group; asking for one method selects that method
+    // alone, even though the class's popover is what opens. Keying this off
+    // `targetId` instead made every member look selected whichever one you
+    // picked.
+    this.assessmentSelectionStore
+      .getState()
+      .selectMultipleElements(assessedIdsFor(elementId, nodes))
+    this.popoverStore.getState().setPopOverElementId(targetId)
```
Claudia-Anthropica: @FelixTJDietrich [medium] This unconditional popover write also runs in modelling mode. `PopoverManager` then resolves the target through `editPopovers`, so `revealAssessment()` opens an element editor even though its public contract says the popover opens only in assessment mode and other modes merely select and reveal.

In `library/lib/apollon-editor.tsx`, `revealAssessment` unconditionally sets the popover target, causing modelling mode to open an editing popover contrary to the API contract. Set the popover target only when the current mode is `ApollonMode.Assessment`, and keep it closed otherwise.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two prior blockers are resolved: assessment nodes now receive the highlight classes, and Codacy succeeds on this head. Two medium issues remain in `revealAssessment`: its selection state becomes inconsistent, and it can open editing UI outside assessment mode.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in 67cd8f77. revealAssessment now updates diagramStore.selectedElementIds alongside the React Flow selected flags for both selection and null clearing. The unit test asserts both transitions, so getSelectedElements() and subscribeToSelectionChange() stay consistent in editable modes.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in 67cd8f77. The popover target is now set only in Assessment mode and explicitly remains null otherwise. A modelling-mode regression test verifies that revealAssessment still selects the requested element without opening editing UI.

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,105 @@ export class ApollonEditor {
...
+    const targetId = owner?.id ?? elementId
+    const element =
+      nodes.find((node) => node.id === targetId) ??
+      edges.find((edge) => edge.id === targetId)
+    const isTopLevel = element !== undefined
+
+    setNodes((current) =>
+      current.map((node) => ({ ...node, selected: node.id === targetId }))
+    )
+    setEdges((current) =>
```
Claudia-Anthropica: @FelixTJDietrich [medium] `setEdges` is a document-persisting mutator: unlike the local-only `select` branch of `onEdgesChange`, it reconciles these edge objects into Yjs and creates an undo item. Revealing an edge therefore persists its transient `selected` flag into collaboration state, and the same problem occurs on edge hops in `useGoToNextAssessment`. Route both selection-only updates through the local selection path instead.

In `library/lib/apollon-editor.tsx`, `revealAssessment` uses persistent `setEdges` for transient edge selection, writing selection into Yjs and undo history. Update `revealAssessment` and the assessment navigator to apply edge selection through the store's local-only selection path.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two prior `revealAssessment` blockers are resolved, and Codacy succeeds on this head. One medium issue remains: host-driven reveal and assessment footer navigation send transient edge selection through `diagramStore.setEdges`, persisting it to Yjs and undo history. The captured non-required jobs are still running, with no failed checks reported.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in 63aa5d1c. Both revealAssessment and assessment footer navigation now send node/edge select changes through onNodesChange/onEdgesChange, which handle selection locally without a Yjs transaction. The new reveal regression asserts that selecting an edge changes selectedElementIds while leaving the persisted Yjs edge without selected and the undo stack empty; the navigation regression asserts edge hops use onEdgesChange. The full library suite passes 1,845 tests.

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,122 @@ export class ApollonEditor {
...
+      assessedIdsFor(node.id, nodes).includes(elementId)
+    )
+    const targetId = owner?.id ?? elementId
+    const element =
+      nodes.find((node) => node.id === targetId) ??
+      edges.find((edge) => edge.id === targetId)
+    const isTopLevel = element !== undefined
+
+    onNodesChange(
+      nodes.map((node) => ({
```
Claudia-Anthropica: @FelixTJDietrich [medium] `onNodesChange` and `onEdgesChange` do not apply this batch atomically: for every select change, `diagramStore` maps the entire collection and replaces `selectedElementIds`. Sending one change per node and edge therefore makes each reveal quadratic and causes `subscribeToSelectionChange()` to receive empty, partial, and repeated selections before the final ID; the assessment navigator repeats the same pattern. Add a local-only bulk selection action that updates node flags, edge flags, and `selectedElementIds` once without writing to Yjs.

In `library/lib/apollon-editor.tsx`, selection is routed through per-item change handlers, causing quadratic work and transient selection notifications. Add a local-only bulk selection action that atomically updates node flags, edge flags, and `selectedElementIds`, and use it here and in the assessment navigator.

</details>

library/lib/hooks/useElementInteractions.ts
```diff
@@ -42,21 +40,42 @@ export const useElementInteractions = () => {
...
+      if (!canOpenPopover || canOpenAssessmentPopover) return
+      setPopOverElementId(edge.id)
+    },
+    [canOpenAssessmentPopover, canOpenPopover, setPopOverElementId]
+  )
+
+  const onNodeClick: NodeMouseHandler<Node> = useCallback(
+    (_event, node) => {
+      if (!canOpenAssessmentPopover) return
+      setPopOverElementId(node.id)
```
Claudia-Anthropica, line 69: @FelixTJDietrich [medium] In read-only assessment mode this stores an ungraded node as the open popover target, but `PopoverManager` returns `null` for that same node via `hasAssessmentToShow`. With no `GenericPopover` mounted, nothing clears the target; after a pane click clears assessment selection, `App` still derives `apollon-assessment-focus` from the stale ID and leaves the node marked amber. Apply the same feedback-availability guard to read-only node and edge clicks before setting the popover ID.

In `library/lib/hooks/useElementInteractions.ts`, read-only assessment clicks set a popover target for elements whose feedback popover cannot render, leaving stale focus state. Guard both node and edge click handlers with `hasAssessmentToShow` before setting the popover ID in read-only mode.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The prior Yjs persistence issue is fixed, but two medium blockers remain on this head. Bulk selection now goes through per-item handlers that cause quadratic work and transient subscription notifications, while read-only clicks on ungraded elements leave stale focus state without mounting a popover. The captured snapshot has no failed checks; several non-required jobs remain in progress.

library/lib/apollon-editor.tsx
FelixTJDietrich: Follow-up in 70e9eb21: the two call sites and undo restoration now share a single diagramStore.setLocalSelection local-only mutation. This avoids intermediate selection events and removes the duplicated mapping code. The exact package-size gate now passes at 121.89 kB / 122 kB without increasing its budget.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in `70e9eb21`. `diagramStore.setLocalSelection()` now maps nodes and edges and replaces `selectedElementIds` inside one Zustand `set`, so reveal/navigation are O(n), emit one complete local selection notification, and never write through the Yjs change handlers. The reveal regression now explicitly asserts exactly one notification containing the final edge ID.

library/lib/hooks/useElementInteractions.ts
FelixTJDietrich: Fixed in `5e0644b2`. Both node and edge click handlers now apply the same `hasAssessmentToShow` guard before setting a popover target in read-only assessment mode. The focused regression covers ungraded nodes/edges being ignored and graded nodes/edges opening normally.

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,96 @@ export class ApollonEditor {
...
+    // alone, even though the class's popover is what opens. Keying this off
+    // `targetId` instead made every member look selected whichever one you
+    // picked.
+    this.assessmentSelectionStore
+      .getState()
+      .selectMultipleElements(assessedIdsFor(elementId, nodes))
+    this.popoverStore
+      .getState()
+      .setPopOverElementId(
+        this.metadataStore.getState().mode === Apollon.ApollonMode.Assessment
```
Claudia-Anthropica: @FelixTJDietrich [medium] In read-only assessment mode, `revealAssessment` still assigns an ungraded element as the popover target. `PopoverManager` rejects that target through `hasAssessmentToShow`, so no `GenericPopover` mounts to clear it, while `App` continues painting the element with `apollon-assessment-focus`; the new guard covers pointer clicks but not this public API path.

In `library/lib/apollon-editor.tsx`, `revealAssessment` creates a stale popover target and assessment focus for ungraded elements in read-only assessment mode. Before setting the popover target, apply `hasAssessmentToShow` when assessment mode is read-only and set the target to `null` when no feedback can render.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two prior blockers are fixed: selection updates are now atomic and local-only, and read-only pointer clicks no longer target unavailable feedback. One medium issue remains in the public `revealAssessment` path. The captured snapshot has no failed checks; non-required jobs remain in progress.

library/lib/hooks/useElementInteractions.ts
FelixTJDietrich: Follow-up in `2c2e8c24`: the guard now subscribes only to the stable assessment getter and checks the clicked node data directly. This preserves the feedback-availability behavior without subscribing `App` to the full node collection; a regression asserts node/edge callback identities remain stable across diagram-store updates.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The latest callback-stability change looks sound, but the existing medium blocker remains on this exact head. `revealAssessment` still assigns every target in assessment mode without applying the read-only feedback guard, while `PopoverManager` rejects ungraded targets, leaving stale popover and focus state. The snapshot reports no failed checks; the unfinished jobs are non-required.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in `26a33bb1`. `revealAssessment()` now applies `hasAssessmentToShow` before assigning a target in read-only assessment mode and explicitly assigns `null` when no feedback can render. The regression first opens a graded target, reveals an ungraded/unknown ID, and asserts the popover target is cleared; graded nodes, edges, and member-owned feedback remain covered.

library/lib/i18n/labels.ts
```diff
@@ -21,6 +21,15 @@ export interface ApollonLabels {
   redoHint: string
   multiSelection: string
   multiSelectionHint: string
+  /**
+   * Shown over the canvas when a scroll-locked editor swallows a wheel gesture.
+   * Receives the platform's zoom modifier already rendered as a key cap ("⌘" on
+   * Apple hardware, "Ctrl" elsewhere) so the sentence reads naturally in any
+   * language without the host having to detect the platform itself.
+   */
+  scrollLockHint: (modifier: string) => string
```
Claudia-Anthropica: @FelixTJDietrich [medium] These three new fields are required on the exported `ApollonLabels` interface, so a complete translation object that was valid in the previous release is no longer assignable, despite this being released as a minor change. The existing `moveEdgeWaypoint?` immediately below demonstrates the compatibility-preserving pattern: new public keys remain optional while `ResolvedApollonLabels` and `mergeLabels` fill their defaults.

In `library/lib/i18n/labels.ts`, the new required `scrollLockHint`, `scrollLockHintTouch`, and `previousAssessment` fields break existing consumers that implement the exported `ApollonLabels` interface. Make all three fields optional on the public interface while retaining required resolved defaults internally.

</details>

library/lib/components/DraggableGhost.tsx
```diff
@@ -116,35 +113,28 @@ export const DraggableGhost: React.FC<DraggableGhostProps> = ({
...
-    const ratioY =
-      (dropElementConfig.dropHeight ?? dropElementConfig.height) /
-      dropElementConfig.height
-    setGhostScale({
-      x: (ratioX * zoom) / previewScale,
-      y: (ratioY * zoom) / previewScale,
+    setGhostRender({ scale: zoom })
+
+    const ghostWidth = ghostDropWidth * zoom
+    const ghostHeight = ghostDropHeight * zoom
```
Claudia-Anthropica: @FelixTJDietrich [medium] `grabY` is measured against the preview's complete painted height, but it is multiplied by only the logical `dropHeight`. For example, a Petri-net place has a 60px logical height while its SVG and preview paint 90px including the label band; grabbing the preview midpoint records `0.5`, then positions the ghost only 30px above the pointer instead of 45px, so it jumps by 15px. `dropAtPointer` repeats the same logical-height calculation, affecting Petri places/transitions and SFC transition branches.

In `library/lib/components/DraggableGhost.tsx`, label-bearing palette previews compute a grab fraction from their full painted height but map it onto the smaller logical drop height, so the ghost and dropped node do not preserve the grabbed point. Derive an unscaled grab offset from the preview's actual painted extent, apply the per-axis drop/preview size ratio, and use that same mapped offset for both ghost positioning and `usePalettePlacement`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The prior read-only reveal blocker is fixed on this head. Two medium issues remain: the exported translation type introduces a breaking requirement, and label-bearing palette elements no longer preserve their grabbed point. The captured snapshot has no failed checks; the unfinished jobs are non-required.

library/lib/i18n/labels.ts
FelixTJDietrich: Fixed in `43e42a8b`. All three new public keys are optional on `ApollonLabels`; `ResolvedApollonLabels = Required<ApollonLabels>` remains the internal store/hook type, so defaults stay non-optional after `mergeLabels`. A compile-time regression assigns a complete legacy dictionary with those keys omitted and verifies the default is restored.

library/lib/components/DraggableGhost.tsx
FelixTJDietrich: Fixed in `43e42a8b`. The drag now recovers the unscaled painted grab offset using the preview's uniform screen scale, maps it by each axis's drop/preview size ratio, and passes the same flow-space offset to both ghost positioning and `dropAtPointer`. The regression models a 60×90 label-bearing preview grabbed at y=45 and asserts both ghost and node top are y=255 (the previous code produced y=270).

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1123,97 @@ export class ApollonEditor {
...
+        ? [element.source, element.target]
+        : [targetId]
+    const anchors = anchorIds
+      .map((id) => nodes.find((node) => node.id === id))
+      .filter((node): node is (typeof nodes)[number] => node !== undefined)
+    if (anchors.length === 0) return
+
+    const centre = anchors.reduce(
+      (accumulator, node) => ({
+        x: accumulator.x + node.position.x + (node.width ?? 0) / 2,
```
Claudia-Anthropica: @FelixTJDietrich [medium] React Flow stores a child node's `position` relative to its `parentId`, but this sum passes those local coordinates directly to `setCenter`. Revealing a node nested in a BPMN container or swimlane—or an edge attached to one—therefore pans by the missing parent offset; `useAssessmentNavigation` repeats the same calculation. Resolve each anchor through `getPositionOnCanvas(node, nodes)` before adding half its size in both paths.

In `library/lib/apollon-editor.tsx`, assessment reveal and navigation calculate centers from parent-relative node positions, causing incorrect panning for nested elements. Use `getPositionOnCanvas` for every node anchor in both `revealAssessment` and `useAssessmentNavigation` before averaging the centers.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two prior blockers are fixed, and the captured CI checks succeed on this exact head. One medium issue remains: assessment reveal/navigation computes pan targets from parent-relative node coordinates, so nested elements are centered at the wrong canvas location.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in e4a8cc69. Both revealAssessment() and useAssessmentNavigation now resolve every node anchor through getPositionOnCanvas(node, nodes) before adding half its dimensions. Regressions cover a nested-node reveal at absolute center (580, 250) and an edge whose nested endpoint yields midpoint (85, 115). The focused tests and full local PR gate pass.

library/lib/apollon-editor.tsx
```diff
@@ -1113,6 +1124,100 @@ export class ApollonEditor {
...
+      (!readonly || hasAssessmentToShow(targetId, nodes, getAssessment))
+    this.popoverStore
+      .getState()
+      .setPopOverElementId(canOpenFeedback ? targetId : null)
+
+    if (options?.reveal === false || !isTopLevel) return
+
+    const rf = this.reactFlowInstance
+    if (!rf) return
+    // An edge has no position of its own; centre on the midpoint of the nodes it
```
Claudia-Anthropica: @FelixTJDietrich [medium] This assumption is false for automatically or manually routed relationships: the badge, label, and popover anchor are rendered at `pathMiddlePosition`, the arc-length midpoint of the displayed polyline, while this code centers the straight midpoint between endpoint nodes. A route detouring around obstacles can therefore leave its assessment anchor outside the viewport after `revealAssessment`; `useAssessmentNavigation` repeats the same calculation. Use the displayed route’s midpoint in both paths, with endpoint centers only as a fallback while route geometry is unavailable.

In `library/lib/apollon-editor.tsx`, routed-edge assessment reveal centers the endpoints instead of the rendered route midpoint, so the badge and popover anchor can remain offscreen. Derive the center from the displayed edge route in both `revealAssessment` and `useAssessmentNavigation`, falling back to endpoint centers only when route geometry is unavailable.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The nested-node coordinate fix is sound, and the captured checks succeed on this head. One medium issue remains: reveal and footer navigation still pan routed relationships to the midpoint between their endpoint nodes rather than to the relationship’s rendered assessment anchor.

library/lib/apollon-editor.tsx
FelixTJDietrich: Fixed in 900edf9b. Host reveal and footer navigation now share getAssessmentElementCenter(): it uses the currently displayed preview/settled route and the same arc-length midpoint as the rendered badge/popover, then falls back to absolute endpoint centers only before route geometry is available. Regressions cover both the detoured-route anchor and the nested-endpoint fallback. The full local gate passes (102 files, 1,851 tests), and the package remains within the unchanged 122 kB limit at 121.89 kB.

Claudia-Anthropica (approved):
@FelixTJDietrich The routed-assessment fix is sound: reveal and footer navigation now derive edge centers from the same displayed route geometry and midpoint calculation used by rendering, with endpoint fallback. All prior threads are resolved and the captured snapshot has no failed checks; several non-required jobs remain in progress.

Claudia-Anthropica (approved):
@FelixTJDietrich The new export pass correctly resolves parent and tspan baselines exactly once after positions and relative font sizes become absolute. Focused unit coverage and refreshed visual exports exercise repeated and child-only baselines through the production download path; the captured snapshot has no failed checks, with only non-required jobs still running.

Claudia-Anthropica (approved):
@FelixTJDietrich The export-size follow-up preserves the approved normalization: it still resolves parent and tspan baselines exactly once, removes them afterward, and the refreshed use-case baseline reflects the intended positioning change. All retained prior threads remain resolved, and the complete captured CI snapshot reports no failed checks; the unfinished jobs are non-required.

PR #845, opened 2026-09-10
feat: add JetBrains IDE plugin for Apollon diagrams

ahmedsyed5989:
Summary
  
  - Adds a jetbrains-plugin workspace: a Kotlin plugin host (IntelliJ Platform SDK) plus a React/JCEF webview canvas, giving Apollon a native IntelliJ IDEA/WebStorm plugin alongside the existing VS Code extension.
  - Adds .github/workflows/release-jetbrains-plugin.yml to build, verify, and (once Marketplace credentials are configured) publish the plugin.
  - Registers the new package with commitlint (jetbrains scope), the pnpm workspace, and the fixed Changesets version group.
  - Fixes a handshake race that left the canvas and toolbar completely blank on load — see Implementation notes.
  - Fixes a merge artifact that left a stale, prettier-reformatted copy of the built webview bundle committed instead of the real minified build.

  Release note

  Edit Apollon diagrams directly inside IntelliJ IDEA and WebStorm. Install the new JetBrains plugin, then open or create a .apollon file to get the full canvas — draw class, object, use-case, activity, and other UML diagrams — plus one-click SVG/PNG export and auto-export on save.

  Implementation notes

  - The host/webview split mirrors the VS Code extension's architecture but doesn't share its protocol types. JCEF has no supported way to register a custom URL scheme, so assets are served over http://apollon.localhost/ via a JBCefLocalRequestHandler override instead of acquireVsCodeApi().
  - Root cause of the blank canvas: the bridge function window.__apollonPostToHost was injected from onLoadEnd, which fires only after every subresource — including the ~2.4MB export WASM — finishes downloading. The webview's React app posts its "ready" handshake from a useEffect that fires as soon as
    its own module script runs, well before that. The post silently no-op'd (optional chaining in jcefBridge.ts), the host never replied with init, and the canvas stayed in its loading state forever. Fixed by injecting the bridge in onLoadStart instead, which fires right after navigation commits and
    before the page's own scripts execute.
  - The built webview bundle is committed under jetbrains-plugin/src/main/resources/webview/ (Gradle's copyWebviewAssets regenerates it from webview/dist) rather than gitignored, so the plugin is buildable without a separate JS build step downstream. Added that path to .prettierignore so lint-staged
    doesn't reformat the minified output on commit.
  - pnpm-workspace.yaml was missing the jetbrains-plugin root package (only jetbrains-plugin/webview was registered), so pnpm changeset couldn't resolve @tumaet/jetbrains-plugin from the fixed version group — added it.

  Steps for testing

  1. pnpm install
  2. pnpm run build:jetbrains
  3. cd jetbrains-plugin && ./gradlew runIde
  4. In the sandboxed IDE, create or open a .apollon file — the canvas and toolbar should render immediately (previously stayed blank).
  5. Draw a few elements, save, and check auto-export under Settings → Tools → Apollon.

  Screenshots / screencasts

Screenshot_20260911_000441
Screenshot_20260911_000503
Screenshot_20260911_000556

  Checklist

  - [x] Added a changeset whose summary is written in the user's voice (.changeset/jetbrains-plugin-launch.md)
  - [x] PR title's Conventional Commit type (feat) matches the kind of change
  - [x] Tests added or updated — DiagramDocumentTest.kt (./gradlew test)
  - [x] Ran pnpm lint && pnpm format:check && pnpm build && pnpm test locally — green


jetbrains-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <!-- `com.intellij.ui.jcef.*` (JBCefApp/JBCefBrowser/...) lives in a
+         separate V2 content module, not in what `com.intellij.modules.platform`
+         pulls in — omitting this throws NoClassDefFoundError at runtime the
+         moment ApollonFileEditor touches JBCefApp, even though it compiles
+         fine (the compile-time platform jar has no such split). Confirmed
+         against the platform's own module descriptors: the bundled Markdown
+         plugin's JCEF-based preview declares the same dependency. -->
+    <dependencies>
+        <module name="intellij.platform.ui.jcef" />
```
Claudia-Anthropica, line 28: @ahmedsyed5989 [high] This declares a required module that `build.gradle.kts` acknowledges is absent from the advertised build-252 floor. Required modules are resolved before `ApollonFileEditor` can use the old pre-split classpath, so the plugin will be disabled on IntelliJ 2025.2; excluding `MISSING_DEPENDENCIES` from verification does not make that installation valid.

In `jetbrains-plugin/src/main/resources/META-INF/plugin.xml`, the required JCEF module is unavailable at the supported build-252 floor. Make the JCEF dependency optional and version-specific so build 252 can load, then restore `MISSING_DEPENDENCIES` as a verifier failure.

</details>

.github/workflows/release-jetbrains-plugin.yml
```diff
@@ -0,0 +1,317 @@
...
+        env:
+          JETBRAINS_MARKETPLACE_TOKEN: ${{ secrets.JETBRAINS_MARKETPLACE_TOKEN }}
+          JETBRAINS_CERTIFICATE_CHAIN: ${{ secrets.JETBRAINS_CERTIFICATE_CHAIN }}
+          JETBRAINS_PRIVATE_KEY: ${{ secrets.JETBRAINS_PRIVATE_KEY }}
+          JETBRAINS_PRIVATE_KEY_PASSWORD: ${{ secrets.JETBRAINS_PRIVATE_KEY_PASSWORD }}
+        # signPlugin/publishPlugin read these as env vars via the `signing {
+        # }`/`publishing { }` blocks in build.gradle.kts. buildPlugin already
+        # ran in the build job; Gradle re-resolves the existing ZIP rather
+        # than rebuilding since inputs haven't changed.
+        run: ./gradlew signPlugin publishPlugin --console=plain
```
Claudia-Anthropica: @ahmedsyed5989 [high] This is a fresh runner containing only the sparse checkout and downloaded ZIP. `signPlugin` defaults to `buildPlugin`'s output and therefore reruns that task graph without Gradle history; it reaches `copyWebviewAssets`, but this job has neither `webview/dist` nor a pnpm build step, so Marketplace publishing will fail as soon as credentials are configured.

In `.github/workflows/release-jetbrains-plugin.yml`, the publish job invokes the default build-backed signing task without the webview build inputs. Configure `signPlugin` and `publishPlugin` to consume the downloaded unsigned ZIP and its signed output directly, without invoking `buildPlugin`.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/theme/ThemeBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+        border =
+            UIManager.getColor("Component.borderColor")
+                ?: UIManager.getColor("Separator.separatorColor")
+                ?: Color(0xce, 0xd4, 0xda),
+        borderSubtle = UIManager.getColor("Separator.separatorColor") ?: Color(0xe9, 0xec, 0xef),
+        danger = if (dark) Color(0xf1, 0x4c, 0x4c) else Color(0xd3, 0x33, 0x33),
+    )
+}
+
+private fun Color.toCss(): String = "rgba(%d, %d, %d, %.3f)".format(red, green, blue, alpha / 255.0)
```
Claudia-Anthropica: @ahmedsyed5989 [medium] `String.format` uses the process default locale. Under a German locale, for example, alpha `1.0` becomes `1,000`, yielding an invalid CSS value such as `rgba(255, 255, 255, 1,000)` and preventing the IDE-derived theme tokens from applying.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/theme/ThemeBridge.kt`, CSS numeric formatting depends on the user's locale and can emit invalid decimals. Format the RGBA value with `Locale.ROOT` and add a non-English-locale test.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/settings/ApollonConfigurable.kt
```diff
@@ -0,0 +1,41 @@
...
+        return JPanel(BorderLayout(8, 0)).apply {
+            add(JLabel("Export image on save:"), BorderLayout.WEST)
+            add(box, BorderLayout.CENTER)
+        }
+    }
+
+    override fun isModified(): Boolean = combo?.selectedItem != ApollonSettings.getInstance(project).autoExport
+
+    override fun apply() {
+        (combo?.selectedItem as? AutoExport)?.let { ApollonSettings.getInstance(project).autoExport = it }
```
Claudia-Anthropica: @ahmedsyed5989 [medium] Applying the setting only mutates persistent state; nothing sends the otherwise-defined `HostMessage.AutoExportChanged`. Every open canvas therefore keeps displaying its previous value until reload and can claim auto-export is off while saves write images, or vice versa.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/settings/ApollonConfigurable.kt`, applying an auto-export change leaves open webviews with stale state. Broadcast the applied value through a project-level topic and have each open `ApollonFileEditor` post `HostMessage.AutoExportChanged`.

</details>

Claudia-Anthropica (changes requested):
@ahmedsyed5989 Changes are requested for the compatibility, publishing, theme, and settings defects noted inline. The `Validate title` check is also red: it reads `main`'s commitlint configuration, where `jetbrains` is not yet allowed, and the captured title has leading whitespace; use a scope-less title such as `feat: add JetBrains IDE plugin for Apollon diagrams`. Codacy remains red, but its captured evidence is partial and only exposes style/documentation warnings, so it did not establish another functional finding.

jetbrains-plugin/src/main/resources/META-INF/plugin.xml
ahmedsyed5989: advertised and support floor updated build-262

Claudia-Anthropica (changes requested):
@ahmedsyed5989 The current head still advertises `sinceBuild = 252`, excludes `MISSING_DEPENDENCIES`, and requires the JCEF module unavailable at that floor, so the compatibility finding remains. The publishing job still rebuilds without `webview/dist`, while the locale-sensitive theme formatting and missing settings broadcast are also unchanged. `Validate title` remains red because the title still has leading whitespace and uses the base branch's unsupported `jetbrains` scope. Codacy is also red, but its partial captured evidence exposes only style and documentation warnings and establishes no additional functional finding.

jetbrains-plugin/src/main/resources/META-INF/plugin.xml
Claudia-Anthropica: @ahmedsyed5989 The checked-out head still has `sinceBuild = "252"` and excludes `MISSING_DEPENDENCIES`, while `plugin.xml` requires the JCEF module, so the compatibility issue remains unresolved.

ahmedsyed5989: compile and support floor updated to build-262

.github/workflows/release-jetbrains-plugin.yml
ahmedsyed5989: Fixed: signPlugin now accepts -PprebuiltPluginZip=... to point archiveFile at the build job's downloaded artifact, so it never depends on buildPlugin/copyWebviewAssets on the fresh runner.

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/theme/ThemeBridge.kt
ahmedsyed5989: Fixed: Color.toCss() now uses String.format(Locale.ROOT, ...) instead of the default locale.

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/settings/ApollonConfigurable.kt
ahmedsyed5989: Fixed: apply() now pushes HostMessage.AutoExportChanged to every open ApollonFileEditor via a new applyAutoExportSetting() call.

Claudia-Anthropica (approved):
@ahmedsyed5989 The new head resolves the prior compatibility, publishing, locale-sensitive theme, and settings-propagation findings: it targets build 262 with dependency verification restored, signs the downloaded artifact, formats CSS with `Locale.ROOT`, and updates open editors when settings change. The current scope-less title is valid against the base configuration. Codacy remains red, but its partial captured evidence exposes only style and documentation notices and establishes no functional defect.

jetbrains-plugin/src/main/resources/META-INF/plugin.xml
Claudia-Anthropica: @ahmedsyed5989 Thanks, this is fixed now: the plugin compiles against and declares build 262, and `MISSING_DEPENDENCIES` is restored as a verifier failure.

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/render/PlantUmlRenderService.kt
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * The one accommodation the sandbox makes is [withLocalC4Includes], which answers a remote
+ * C4-PlantUML include from PlantUML's own bundled copy rather than letting it fail. It still
+ * fetches nothing.
+ */
+object PlantUmlRenderService {
+    private const val SECURITY_PROFILE_PROPERTY = "PLANTUML_SECURITY_PROFILE"
+
+    private fun ensureSandboxed() {
+        if (System.getProperty(SECURITY_PROFILE_PROPERTY) == null) {
```
Claudia-Anthropica, line 33: @ahmedsyed5989 [high] An existing JVM property wins here. If the IDE or another PlantUML integration has already selected `INTERNET` or `UNSECURE`, opening a project `.puml` runs `SourceStringReader` with that weaker profile, allowing includes to access resources despite this plugin's no-network/no-file-access guarantee.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/render/PlantUmlRenderService.kt`, `ensureSandboxed` accepts an already-set weaker security profile. Establish `SANDBOX` before this plugin's PlantUML classes initialize, refuse rendering if that cannot be guaranteed, and add isolated regression coverage with a pre-set weaker profile.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/puml/PumlLayoutSidecar.kt
```diff
@@ -0,0 +1,312 @@
...
+            val nameById = mutableMapOf<String, String>()
+            val nodes = mutableMapOf<String, NodeLayout>()
+
+            modelNodes.forEach { node ->
+                val id = text(node["id"]) ?: return@forEach
+                val data = obj(node["data"])
+                val name = text(data?.get("name")) ?: return@forEach
+                nameById[id] = name
+                val position = obj(node["position"])
+                nodes[name] =
```
Claudia-Anthropica, line 279: @ahmedsyed5989 [high] This map overwrites distinct PlantUML elements that share display text, which is legal when they have different aliases. After the sidecar is written, `toPreviousModel()` emits only one name-based identity and the family mappers can assign that same node ID to both elements on reopen, collapsing node and relation geometry.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/puml/PumlLayoutSidecar.kt`, layout identity is keyed only by non-unique display names. Persist and restore nodes and edges by a unique stable PlantUML ref/alias, update mapper matching accordingly, and add a full sidecar round-trip test for two aliases with the same display name.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/puml/DiagramTypeDetector.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        val deploymentHits = body.count { DEPLOYMENT_KEYWORDS.containsMatchIn(it) || ARTIFACT_KEYWORD.containsMatchIn(it) }
+        val componentHits = body.count { COMPONENT_ONLY_KEYWORDS.containsMatchIn(it) }
+        // A `class`/`enum`/`entity` declaration settles it: those appear in no other family, while
+        // `component`/`node` diagrams routinely borrow `interface`. Without this guard a class
+        // diagram declaring one interface lost every one of its classes to the component importer.
+        val classHits = body.count { CLASS_ONLY_KEYWORD.containsMatchIn(it) }
+        if (classHits == 0 && (deploymentHits > 0 || componentHits > 0)) {
+            return if (deploymentHits > componentHits) DiagramFamily.DEPLOYMENT else DiagramFamily.COMPONENT
```
Claudia-Anthropica, line 103: @ahmedsyed5989 [medium] Counting declarations misclassifies a normal nested deployment such as `node Server { component API }`: one deployment hit ties one component hit, so this returns `COMPONENT`. The component importer then treats the entire node block as unsupported and rejects the file, leaving a common deployment diagram without an Edit canvas.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/puml/DiagramTypeDetector.kt`, equal deployment/component counts classify nested deployment blocks as component diagrams. Make explicit deployment-container evidence such as `node`, `device`, or `execution environment` dominate contained components and add a nested node-with-component test.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/editor/PlantUmlPreviewFileEditor.kt
```diff
@@ -0,0 +1,124 @@
...
+    private fun scheduleRender() {
+        alarm.cancelAllRequests()
+        alarm.addRequest(::renderNow, RENDER_DEBOUNCE_MILLIS)
+    }
+
+    private fun renderNow() {
+        val browser = browser ?: return
+        val text = FileDocumentManager.getInstance().getDocument(file)?.text ?: return
+        val svg =
+            when (val result = PlantUmlRenderService.render(text)) {
```
Claudia-Anthropica, line 69: @ahmedsyed5989 [medium] `renderNow` is called directly during editor construction and from an `Alarm` configured for `SWING_THREAD`, so PlantUML parsing and SVG layout execute on the EDT. Opening or typing in a sufficiently large diagram will freeze the entire IDE until rendering completes.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/editor/PlantUmlPreviewFileEditor.kt`, PlantUML rendering runs synchronously on the Swing event thread. Move rendering to a cancellable pooled task keyed by a generation token and marshal only the latest `browser.loadHTML` update back to the EDT.

</details>

jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/actions/ExportDiagramAction.kt
```diff
@@ -0,0 +1,38 @@
...
+import de.tum.cit.aet.apollon.editor.ApollonFileEditor
+import de.tum.cit.aet.apollon.protocol.ExportFormat
+
+/** Export the focused diagram to a sibling image, asking for the format. */
+class ExportDiagramAction : AnAction() {
+    override fun getActionUpdateThread(): ActionUpdateThread = ActionUpdateThread.BGT
+
+    override fun update(e: AnActionEvent) {
+        val project = e.project
+        e.presentation.isEnabledAndVisible =
```
Claudia-Anthropica, line 18: @ahmedsyed5989 [medium] The new `PumlCanvasFileEditor` implements `export`, and auto-export already calls it, but this action remains enabled only for `ApollonFileEditor`. Consequently **Export Diagram…** is unavailable for every PlantUML tab even though the plugin advertises export for the focused diagram.

In `jetbrains-plugin/src/main/kotlin/de/tum/cit/aet/apollon/actions/ExportDiagramAction.kt`, manual export ignores PlantUML editors. Resolve the selected PlantUML file's associated `PumlCanvasFileEditor` and enable and invoke export for it as well as for `ApollonFileEditor`.

</details>

package.json
```diff
@@ -10,16 +10,18 @@
     "build:server": "pnpm --filter @tumaet/server run build",
     "build:webapp": "pnpm --filter @tumaet/webapp run build",
     "build:vscode": "pnpm --filter apollon-extension run build:all",
+    "build:jetbrains": "pnpm --filter @tumaet/apollon --filter @tumaet/jetbrains-webview run build",
     "build:docs": "pnpm run build:lib && pnpm --filter @tumaet/docs run build",
     "check:doc-snippets": "pnpm run build:lib && node scripts/check-doc-snippets.mjs",
-    "storybook": "pnpm --filter @tumaet/webapp run storybook",
+    "storybook": "pnpm build:lib --filter @tumaet/webapp run storybook",
```
Claudia-Anthropica, line 16: @ahmedsyed5989 [medium] This command no longer chains the library build and the webapp Storybook command. It invokes `build:lib` with `--filter @tumaet/webapp run storybook` as extra arguments, so the webapp's `storybook` script is never started.

In `package.json`, the root `storybook` script passes the intended second command as arguments to `build:lib`. Restore an explicit chain such as `pnpm run build:lib && pnpm --filter @tumaet/webapp run storybook`.

</details>

Claudia-Anthropica (changes requested):
@ahmedsyed5989 The post-approval PlantUML/C4 expansion introduces blocking security, identity/data-integrity, classification, responsiveness, export-integration, and root-script regressions noted inline. Helios is captured successful; Codacy remains `ACTION_REQUIRED`, but its partial evidence exposes only style and documentation notices and establishes no additional functional defect.

